Issue #39, opened 2018-09-25
Create endpoint for BBL <> Landlord name and business address


romeboards:
This is for https://github.com/JustFixNYC/tenants2/issues/169 (cc @toolness)

It should be a get request with a BBL query param and return json something like: 

```
{
    name: "BOBBY DENVER",
    address: "12 SPENCER STREET 4 11205"
}
```

We can split name / address info in various formats if we want as well...

PR #40, opened 2018-09-26
adding landlord contact table and endpoint

romeboards:
This addresses #39 by adding a `landlord_contact` table to Postgres and supplying an accompanying endpoint at `/api/landlord`. It requires a BBL in the querystring as such:

`https://whoownswhat.justfix.nyc/api/landlord?bbl=3012380016`

should return:

```
{
    "result": [
        {
            "housenumber": "654",
            "streetname": "PARK PLACE",
            "zip": "11216",
            "boro": "BROOKLYN",
            "registrationid": 352819,
            "bbl": "3012380016",
            "bin": 3031404,
            "businessaddr": "12 SPENCER STREET 4 11205",
            "ownername": "MOSES GUTMAN"
        }
    ]
}
```

Some notes:

Fields aren't guaranteed, obviously. See this query:

image

So we usually have the name but maybe not the business address. We know from WoW development that we're taking a narrow view of the address (i.e. only looking at the `CorporateOwner` field, so there's definitely room for improvement at increasing coverage for this. But definitely a cool starting point!

### DIFF
--- a/server/controllers/landlord.js
+++ b/server/controllers/landlord.js
@@ -1,0 +1,16 @@
+const db = require('../services/db'),
+      geo = require('../services/geoclient'),
+      csv = require('csv-express'),
+      rollbar = require('rollbar');
+      Promise = require('bluebird');
+
+module.exports = {
+  query: (req, res) => {
+    db.queryLandlord(req.query.bbl)
+      .then(result => res.status(200).send({ result: result }) )
+      .catch(err => {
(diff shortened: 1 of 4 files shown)


romeboards:
cc @toolness @sraby 

Just need to redeploy after merging...

romeboards:
@sraby we'll need to add this as a step in the build script, just FYI

PR #48, opened 2018-10-16
Terms of Use and Privacy Policy pages added

sraby:
- added in the Terms of Use and Privacy Policy pages, liked to via a footer panel on most pages
- fixed bug where navigating to the pages didn't load from the top of the page


sraby:
moving new updates to another branch

PR #49, opened 2018-10-16
Adds handling for browsers where WebGL is disabled or unavailable

romeboards:
This was a tricky one. Apparently there is little or no accounting for this - I looked through [react-mapbox-gl](https://github.com/alex3165/react-mapbox-gl/blob/master/docs/API.md) and tried some of the event handlers (`onWebGlContextLost`, `onError`?) but they didn't seem to affect anything.

So in the meantime I threw in some code that detects if the browser has WebGL or not (see `MapHelpers`), and throws up a terrible error message and doesn't allow the user to use the map at all:

image

### How to reproduce
You can disable WebGL/hardware acceleration on Chrome in Settings > Advanced Settings > Use hardware acceleration when available:

image

@toolness @sraby interested to hear your thoughts! Specifically it would be good to know if we can find a way to degrade this in a better fashion so a version of the map could still maybe be displayed...

### DIFF
--- a/client/src/components/PropertiesMap.js
+++ b/client/src/components/PropertiesMap.js
@@ -53,7 +53,10 @@ export default class PropertiesMap extends Component {
     super(props);
 
     this.state = {
-      mapLoading: true,
+      // maybe theres a better way for this, but the initial loading boolean
+      // could just be driven by if WebGL is an option or not
+      mapLoading: MapHelpers.hasWebGLContext(),
+      hasWebGLContext: MapHelpers.hasWebGLContext(),
       mapRef: null,
       mobileLegendSlide: false,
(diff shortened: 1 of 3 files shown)


toolness:
Oh interesting... what happens if you use `canvas.getContext('3d')` or something?  I thought that was the canonical way to detect whether 3D was supported but I dunno.

romeboards:
@toolness yeah I found the logic here: https://github.com/mapbox/mapbox-gl-js/issues/3437#issuecomment-380177701

sraby:
If it makes you feel any better, Zola
<https://zola.planning.nyc.gov/?commercial-overlays=false&zoning-districts=false>'s
map also crashes when hardware acceleration is disabled...

Issue #76, opened 2018-11-20
Share URLs need to be encoded in order to handle spaces


romeboards:
![image](https://user-images.githubusercontent.com/627751/48742940-05d0fb00-ec2f-11e8-9aa8-763d648e2213.png)

romeboards:
Addressed by #77

PR #77, opened 2018-11-21
url support for searching via full bbl

sraby:
(mainly out of convenience for myself) I added a route with one single url parameter for a full non-delimited bbl, and a small IF statement to BBLPage to slice out each boro, block, and lot from that input. 

So, for example, the url `.../bbl/2031840066` should work sans backslashes!


romeboards:
Expanded this to include #78 

Addresses #76

PR #82, opened 2018-12-07
Add some support for deploying to the cloud via Docker.

toolness:
Here u go.


toolness:
This isn't super necessary right now and it's clearly way out of date so I'm just going to close it for now, we can always bring it back later if needed.

PR #86, opened 2019-01-24
Bug fix: map now pans when new address selected from list view

sraby:
I refactored PropertiesMap component so that the map pans anytime the detail address prop changes. I suspect this isn't the best way to do this? Open to discussion @romeboards @toolness :) 

Still to do: 
- [x] Test whether we need `this.props.onOpenDetail(addr);` to be called in onClick event handler for point clicks in Property Map component


romeboards:
@toolness would love to hear your thoughts on the best way for @sraby to approach this in good react style when you get a chance...

PR #89, opened 2019-02-05
Fix vertical scrolling on Firefox

### LINKED ISSUE #87: Scrollbar missing in Firefox
Hi JustFixNYC! I thought this site was awesome, so I shared with some friends and one noticed that she couldn't scroll in the right-side panel for more building info. Turns out the scrollbar doesn't appear in Firefox browsers.

toolness:
Fixes #87.

~~I haven't tried this out on other browsers to make sure it doesn't break compatibility with them, though.~~ See below, I think this is good to go!


toolness:
We can also ignore the CircleCI error--it's because #88, which adds the CircleCI configuration, hasn't been merged yet, so CircleCI is just really confused right now.

toolness:
Okay, I've tried this out on Edge, Firefox, and Chrome on Windows, as well as Safari on iOS, and it works fine on all of them. I think this is good to go!

toolness:
Oh I forgot about IE11!

Just tested it on IE11 and it does indeed still work fine. 🎉

romeboards:
Took a quick look and seems to check out on my end as well! Thank you!

@sraby @toolness @jokudasai Can we also make sure to test on mobile devices?  **Longer term:** What do we think is the best strategy for UI/UX related cross-device/browser testing?

toolness:
Yep yep, I tried it on both iOS and iPhone safari and it looks good!  But as far as cross-browser testing I think it makes sense to make sure we do it manually before merging a PR, like we're doing here... @sraby and I just need to hold each other accountable for it. But also it might help to have an additional QA pass before deploying to production to make sure nothing slipped through the cracks, I'm not sure.

toolness:
Ok @sraby I think this is good to merge if you're happy with it!

PR #92, opened 2019-02-05
Add 'exporttestdata' and 'loadtestdata' commands to dbtool.

### LINKED ISSUE #91: Add fixture data
Right now, in order to use WoW, a developer needs to build the production database, which takes a really long time, and is also infeasible for test suites.  It would be nice if we had a very small set of data, e.g. the portfolio of a single landlord, that we could ship with the repository, so that developers could easily install it and test suites could quickly use it.

toolness:
This adds a `dbtool.py exporttestdata` command that outputs SQL to insert the portfolio for ~~E&M Associates~~ All Year Management into the WoW database.

It also adds a `dbtool.py loadtestdata` command that loads the data from a file in a repo, which fixes #91.

## Notes

* ~~There were some weird bugs I ran into when trying to get my local server running, which I fixed in 7165bb1f6d2fe3f3dd8b4ff2c262b30b0bcf8195 and 9c5b7d262d033b30d1a7362d13129634c5929316. I might file those as separate PRs though.~~ Filed as #94.

* I wanted to add a smoke test for `dbtool.py exporttestdata` but I couldn't manage to get `pg_dump` (which the command needs) on CircleCI's path using [CircleCI's own instructions](https://circleci.com/docs/2.0/databases/#using-binaries), so I eventually gave up.

## To do

- [x] Update the README to tell devs to use `loadtestdata`.


dbtool.py
```diff
@@ -193,6 +193,48 @@ def dbshell(db: DbContext):
...
+    with db.connection() as conn:
+        cur = conn.cursor()
+        cur.execute(create_sql)
+
+    try:
+        env = os.environ.copy()
+        env['PGPASSWORD'] = db.password
+        output = subprocess.check_output([
+            'pg_dump',
+            '-h', db.host, '-p', str(db.port), '-U', db.user, '-d', db.database,
```
toolness: This line and the environment are duplicated from the `dbshell` command so it'd be nice to keep things DRY here.

toolness: Fixed this in b55e09b662113bb5857a14b053efa9ac496289b7!

PR #95, opened 2019-02-06
Fix a bunch of linter warnings and add linting to CircleCI

### LINKED ISSUE #70: Fix linter warnings
Running `yarn start` spews so many linter warnings that it's hard to see the log output that _isn't_ a linter warning.  We should clean these up, and then set up a CircleCI job to ensure that future PRs don't introduce new warnings.

toolness:
Fixes #70.

## Notes

* A large number of linter warnings were for [react/jsx-no-target-blank](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md).  However, many of the links with `target="_blank"` are to locations that we trust (like the JustFix website), and to which we _want_  referrer information passed to. For these particular ones, I opted to _remove_ the `target="_blank"` instead of adding `rel="noopener noreferrer"`.  It seems to make sense since we're sending them to one of our own sites.

    I actually considered just disabling this rule altogether but it seems that would require ejecting the app, which I didn't want to do.

* Other linter warnings were accessibility-related. I've disabled those for now, as I'm pretty sure this whole app is not very accessible right now--instead, we can start an issue to do an accessibility audit at a later date, at which point we can fix and re-enable the a11y warnings. I've filed #98 to address this later.

## To do

- [x] Remove all the linter warnings.

- [x] Run the linter as part of the test suite so we don't have regressions.


toolness:
Ok @sraby, I think this is good to go but we should definitely "kick the tires" on the app before merging--hopefully none of the linter fixes (I'm thinking especially of the conversions from `==` to `===`) broke anything!

Issue #53, opened 2018-10-22
Switch out Geoclient for Geosearch


sraby:

toolness:
Ok, I am exploring some ways to make it easier to reuse the relevant logic from the tenants app in https://github.com/JustFixNYC/tenants2/pull/442.

toolness:
Update: I've merged the aforementioned PR and I think we can, at the very least, use the new `geo-autocomplete-base.tsx` module I made in WoW.  There's still some UI/view logic that is probably easiest to write in WoW itself.  We'll have to figure out how we want to process the TypeScript code too, whether it means adding optional TSX support to WoW or pre-compiling the TSX into JS (and perhaps publishing as an npm module) that WoW uses.

PR #99, opened 2019-02-07
Upgrade react-scripts to latest version (2.1.3)

toolness:
~~**Note: This is a PR against #95, not `master`! I will re-target it to `master` once #95 is merged.**~~

This is an attempt to upgrade react-scripts to the latest version--partially just to stay up-to-date, but also because 2.1.0 and higher [support Typescript out of the box](https://facebook.github.io/create-react-app/docs/adding-typescript), which could be useful in sharing code between this app and the tenants app, specifically for #53.

## To do

- [x] Upgrade from 1.0.10 to 2.0.3.
- [x] Upgrade from 2.0.3 to the latest 2.x release.
- [x] Fix all new linter warnings.
- [x] Ensure that at least Node 8 is used in production and on @sraby's dev machine, as [2.x drops support for node 6](https://github.com/facebook/create-react-app/blob/1deb811c5d91ac6d337d2fd4e4f36c7d8c352ebd/CHANGELOG.md#node-6-is-no-longer-supported). **Update: see #100 for more details.**
- [x] Add polyfills for IE11 at the very least, since [IE polyfills are no longer included by default in 2.x](https://github.com/facebook/create-react-app/blob/1deb811c5d91ac6d337d2fd4e4f36c7d8c352ebd/CHANGELOG.md#polyfills-for-ie-9-ie-10-and-ie-11-are-no-longer-included-by-default-but-you-can-opt-in). Then verify that the app still works on IE11.
- [x] Verify that the app doesn't use `require.ensure()` (also a breaking change in 2).
- [x] Ensure dynamic `import()` follows new semantics in 2 (not an issue, we don't use dynamic import)
- [x] Ensure testing suite still works (Jest's default environment was changed to `jsdom` in 2 but we don't have any tests in it yet).
- [x] Ensure the proxy configuration still works (another breaking change in 2).
- [x] Ensure we don't use `PropTypes` in weird ways (another breaking change in 2).


sraby:
@toolness I'm currently using node v8.9.1 on my dev machine :)

toolness:
@sraby Ok I think this is good to go! I kicked the tires on IE11, Edge, Chrome, Firefox, and iPhone Safari.

toolness:
Er, actually hold on a sec @sraby... @romeboards is investigating whether the current WoW supports IE9 and IE10. If it does, then I will have to add the react polyfill thing for it (and we will have to test those browsers going forward, which is hard even for me on Windows b/c I only have IE11 here, so I guess I will have to use BrowserStack).

toolness:
Ok, Dan says we're only supporting IE11 so I think this is indeed good to go.

PR #102, opened 2019-02-11
Move WoW db config out to a YAML file.

toolness:
We need to be able to load/update the WoW database from our containerized DB updater at https://github.com/JustFixNYC/nycdb-k8s-loader/pull/22.  One simple way to do this is by putting all our database configuration into a YAML file, similar to what NYCDB is already doing.

This PR moves the names of all the NYCDB dataset dependencies and the SQL filenames into a new `who-owns-what.yml` file.  It also makes the following changes:

* We no longer load the `pluto_17v1` dataset; it seems this was entirely superseded by `pluto_18v1`, although I should check with @sraby to make sure this is OK.

* The `--update` option for `dbtool.py builddb` now wipes _all_ nycdb data instead of just a subset of tables. This simplified the structure of the YAML and seems OK since nycdb-k8s-loader will now be responsible for updating WoW in a more rigorous manner. If we really want to add this feature back in, we might want to add a separate flag that allows the user to manually specify which datasets to refresh.

* It automatically figures out what the tables in a dataset are by consulting the `nycdb` module.

### DIFF
--- a/dbtool.py
+++ b/dbtool.py
@@ -3,6 +3,9 @@ import sys
 import subprocess
 import argparse
 import time
+import yaml
+import nycdb.dataset
+from nycdb.utility import list_wrap
 from urllib.parse import urlparse
 from typing import NamedTuple, Any, Tuple, Optional, Dict, List
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


sraby:
Checked: we only use pluto_18v1 so removing those dependent datasets is great!

Issue #93, opened 2019-02-06
Navigating back from search results page doesn't go to search page (homepage)


toolness:
From the homepage, if you do a search for an address and then press the browser's "back" button from the search results, instead of going back to the homepage, it goes back to whatever page you were on _before_ the homepage.

toolness:
Fixed by #106!

PR #106, opened 2019-02-13
Resolves issue #93

jessehall3:
I found an explanation in this SO post:
https://stackoverflow.com/questions/48731207/react-router-dom-and-redirect-not-being-added-to-history

That post references this part in the router docs:
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md#push-bool

When redirecting to the search results, nothing was being added to the history, as if the user had never left the initial home page. When clicking the back button, the browser would have to go back to a different page unless the user had clicked any other links on the WOW page.

### DIFF
--- a/client/src/containers/HomePage.js
+++ b/client/src/containers/HomePage.js
@@ -77,7 +77,7 @@ class HomePage extends Component {
       if(!this.state.results.addrs || !this.state.results.addrs.length) {
         window.gtag('event', 'search-notfound');
         return (
-          <Redirect to={{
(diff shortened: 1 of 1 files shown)


toolness:
Whoa, awesome!  Thanks @jessehall3, this works wonderfully (and thanks for documenting the rationale behind your changes too).

Since @sraby is out this week, I'm going to go ahead and merge this PR.  I think the change will be deployed in about a week or so, once @sraby gets back...

Thanks again for your contribution!

PR #107, opened 2019-02-19
Add TypeScript support to client, convert browser.js to ts

toolness:
This adds TypeScript support to the client, a prerequisite for sharing code from https://github.com/JustFixNYC/tenants2/pull/442 to make progress on #53, by following the [Adding TypeScript](https://facebook.github.io/create-react-app/docs/adding-typescript) instructions for CRA.

It also almost-trivially converts `browser.js` to `browser.ts` to show that the TS support works.

## Notes

* I've manually verified that inserting a type error to `browser.ts` causes `yarn build` (which is run on CircleCI) to fail, so we will be alerted whenever type safety is violated.

* I've manually verified that everything still works on IE11.  I've also added some ES6-specific syntax to `browser.ts`, like arrow functions and default arguments, and verified that everything still works on IE11 (i.e., that the TS is properly transpiled to ES5).

* `tsconfig.json` and `react-app-env.d.ts` were added by the dev server once it noticed the project had TS files.

### DIFF
--- a/client/src/util/browser.ts
+++ b/client/src/util/browser.ts
@@ -20,7 +20,7 @@
 
 export default {
 
-  isMobile() {
+  isMobile(): boolean {
     let width = 0;
     if(typeof window != 'undefined') {
       width =  window.innerWidth && document.documentElement.clientWidth ?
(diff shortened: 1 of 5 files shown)


client/src/util/browser.ts
```diff
@@ -32,13 +32,14 @@ export default {
...
-  isIE() {
+  isIE(): boolean {
     if(typeof window != 'undefined') {
       return /MSIE \d|Trident.*rv:/.test(window.navigator.userAgent);
     }
+    return false;
   },
 
-  getParameterByName(name, url) {
+  getParameterByName(name: string, url: string): string|null {
```
toolness: As far as I can tell, neither this function nor `isIE()` are called from anywhere in the code, so we might just want to get rid of them?

Note also that `getParameterByName` is also somewhat obviated by the existence of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams); while it doesn't appear to exist on IE11, I suspect there's a polyfill available.

PR #118, opened 2019-04-09
Add SQL test for fuzzy string matching.

toolness:
So this adds a test to make sure that `get_assoc_addrs_from_bbl` links buildings whose HPD registration contacts have similar names, but it also reveals some funky things.  Most notably, the portfolio generation is very specific to a BBL and `get_assoc_addrs_from_bbl` isn't transitive (I had to remove the bidirectionality test because of this).  This is (at the very least) because the name matching is done based on the name(s) of the HPD contacts of the building that starts the search.  See the test code for more details.

### DIFF
--- a/tests/test_sql.py
+++ b/tests/test_sql.py
@@ -177,13 +177,13 @@ class TestSQL:
 
         self.get_assoc_addrs_from_bbl(FUNKY_BBL, expected_bbls=[
             # Ideally this should include SPUNKY, but because the head officer of FUNKY is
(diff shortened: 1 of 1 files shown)


toolness:
[Merge fuzzy string matching test](https://trello.com/c/dfpwae0O/230-merge-fuzzy-string-matching-test)

sraby:
thank you @toolness!! Why no "BOOP_BBL" ??

PR #119, opened 2019-05-06
Use NYCDB w/ marshal evictions 2018

toolness:
This modifies WoW to use the version of nycdb with 2018 marshal evictions from https://github.com/aepyornis/nyc-db/pull/96, 0.1.22.

It does _not_ actually modify WoW to show 2018 marshal evictions; that will need to be done in a separate PR.


sraby:
@toolness now that the PR was merged into NYCdb do we need this anymore?

toolness:
Absolutely, this is the PR that modifies WoW to use the latest nycdb... without it, it will still use the old nycdb that doesn't include the 2018 marshal evictions!

PR #141, opened 2019-06-19
refactoring animation change process for Chart.js visualization

sraby:


sraby:
@toolness, I wrote you a note in IndicatorsViz.js about the issue I've been having trying to dynamically change the animation speed and rerendering prop for the Chart.js visualization. Still can't figure out a way to get what we want without using `componentWillReceiveProps`. No huge rush, but if you want to take a look I'd be super appreciative :)

PR #164, opened 2019-07-30
Simplified urls for not found pages

abromanos:


sraby:
thank you @abromanos for this!!!

PR #170, opened 2019-08-06
Convert AddressToolbar and Modal components to TypeScript

toolness:
This converts the `<AddressToolbar>` and `<Modal>` components to TypeScript.

The main benefit is that props for the components are now semi-documented--I say "semi" because I haven't added any comments documenting what each property is for, because I'm not very familiar with the code, but at least the name and type of the props are documented now.  Although even the types might be slightly inaccurate--since the consumers of these components haven't been converted yet, I'm just making educated guesses about the types, but we can revise them going forward.

Anyways, it's a start at least.  I think that as we make the types more explicit, it will become easier for folks to understand the code, and it should also uncover potential type errors, like our buddy `Cannot read property 'bbl' of undefined`.

### DIFF
--- a/client/src/components/AddressToolbar.tsx
+++ b/client/src/components/AddressToolbar.tsx
@@ -1,11 +1,25 @@
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
-import Modal from 'components/Modal';
+import Modal from './Modal';
 
 import 'styles/AddressToolbar.css';
 
-export default class AddressToolbar extends Component {
-  constructor(props) {
+export type AddressToolbarProps = {
+  onExportClick: () => void;
+  numOfAssocAddrs: number;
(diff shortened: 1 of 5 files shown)


client/src/components/AddressToolbar.tsx
```diff
@@ -1,11 +1,25 @@
...
 
-export default class AddressToolbar extends Component {
-  constructor(props) {
+export type AddressToolbarProps = {
+  onExportClick: () => void;
+  numOfAssocAddrs: number;
+  userAddr: {
+    housenumber: string;
+    streetname: string;
+    boro: string;
```
toolness: I suspect we will eventually separate this data type into a different place, because I've seen it elsewhere, but that can wait for a future PR where we actually reuse the pattern in a different component.

PR #173, opened 2019-08-12
Update Indicators.js

LC15:
added "name" attribute to radio buttons

name="Dataset" (for the 3 buttons labeled: HPD Complaints, HPD Violations, and Building Permit Applications)
name="Time" (for the 3 buttons labeled: Month, Quearter, and Year)


client/src/components/Indicators.js
```diff
@@ -288,23 +288,26 @@ export default class Indicators extends Component {
                     <em className="Indicators__linksTitle">Select a Dataset:</em> <br/>
                     <li className="menu-item">
                         <label className={"form-radio" + (this.state.activeVis === "complaints" ? " active" : "")} onClick={() => {window.gtag('event', 'complaints-timeline-tab');}}>
-                          <input type="radio" 
+                          <input type="radio"
+                          	name="Dataset"
```
toolness: This seems to be indented more than the other lines--it would be nice if it was indented the same amount as the lines below it.

PR #175, opened 2019-08-13
redirected "Take Action" CTAs to DDO page

sraby:


client/src/components/DetailView.js
```diff
@@ -58,13 +58,15 @@ export default class DetailView extends Component {
 
   render() {
 
-    let boro, block, lot, ownernames, userOwnernames;
+    let boro, block, lot, ownernames, userOwnernames, takeActionURL;
     if(this.props.addr) {
       ({ boro, block, lot } = Helpers.splitBBL(this.props.addr.bbl));
+      takeActionURL = 'https://app.justfix.nyc/ddo?address=' + encodeURIComponent((this.props.addr.housenumber + ' ' + this.props.addr.streetname).trim()) + '&borough=' + encodeURIComponent(this.props.addr.boro) + '/?utm_source=whoownswhat_take_action';
```
toolness: So the borough is supposed to be one of the borough constants defined here:

https://github.com/JustFixNYC/tenants2/blob/master/common-data/borough-choices.json

That basically means that you will want to uppercase the borough name and convert spaces to underscores...

client/src/containers/NychaPage.js
```diff
@@ -83,6 +83,8 @@ export default class NychaPage extends Component {
         } : 
         null );
 
+    const takeActionURL = (usersInputAddress && ('https://app.justfix.nyc/ddo?address=' + encodeURIComponent((usersInputAddress.housenumber + ' ' + usersInputAddress.streetname).trim()) + '&borough=' + encodeURIComponent(usersInputAddress.boro) + '/?utm_source=whoownswhat_take_action_nycha'));
```
toolness: I think you will want the part before the `utm_source` to be a single ampersand (`&`) rather than `/?`.

client/src/components/DetailView.js
toolness: Err also see the comment I left here: https://github.com/JustFixNYC/who-owns-what/pull/175/files#r314397915

Shouldn't both of these almost-identical lines of code in `DetailView.js` and `NychaPage.js` really be factored out into a separate function, e.g. `createTakeActionURL()`?

client/src/containers/NychaPage.js
sraby: It looks like we actually need neither: https://en.wikipedia.org/wiki/UTM_parameters. Or am I missing something?

toolness: Hmm, that's not a valid querystring, though... At least from what I can tell, it looks like you will be generating a URL like `https://foo.com/?address=boop&borough=BROOKLYN/?utm_source=whoownswhat_take_action_nycha`, which means that instead of `address`, `borough`, and `utm_source` arguments, a querystring parser will only see `address` and `borough` (the latter will have a value of `BROOKLYN/?utm_source=whoownswhat_take_action_nycha`).  Maybe I am mis-reading your code though...

toolness: e.g., if you run `node` and give it the following:

```js
require('querystring').parse('address=boop&borough=BROOKLYN/?utm_source=whoownswhat_take_action_nycha')
```

you will get:

```js
{
  address: 'boop',
  borough: 'BROOKLYN/?utm_source=whoownswhat_take_action_nycha'
}
```

sraby: ohhh I see what you mean. Mentally, I've been treating the `address` and `borough` arguments as part of the "URL" and the `utm_source` as an additional argument, when they all should be treated equally as additional arguments.

sraby:
Okay @toolness, check out my refactoring. We also got some news from our new marketing people that we should be splitting up the `utm` arguments more granularly, so I factored that in as well. Also, when an address is not properly formatted for DDO (which technically shouldn't happen, if I'm not mistaken), I'm logging a rollbar error and directing the link to the app.justfix landing page. Let me know what you think!

toolness:
NOICE this looks great!  Thanks sir!

PR #189, opened 2019-12-02
Return 404 for nonexistent static files

### LINKED ISSUE #169: Old JS bundle URLs yield WoW HTML pages
## The problem

We're getting Rollbar errors that look like this:

```
SyntaxError: Unexpected token '<'
File https://whoownswhat.justfix.nyc/static/js/1.4d745252.chunk.js line 1 col 0 in [anonymous]
```

Visiting the link mentioned in the above traceback actually yields an empty WoW page, though, _not_ a JS file:

> image

Every push of WoW to production yields a JS bundle with a different name--this is done for the sake of cache busting.  At the time of this writing, the current main JS bundle is called `/static/js/1.ffb84278.chunk.js`.  Visiting the URL corresponding to that bundle yields, as expected, an actual JS file, rather than an empty WoW page.  But as soon as we push a new version of WoW to production, it will become an empty WoW page.

## The cause

Because we self-host the JS bundle from our app server, this means that when we push new versions of the code, the _old_ bundles no longer exist.  We should really be 404'ing anything under `/static/` that doesn't map to a real file, so our server is falling back to rendering our react app.

This means that when we last pushed to production, there's a chance that people who had the _old_ version of WoW's HTML, which had a `<script>` tag referring to `1.4d745252.chunk.js`, actually got an HTML page instead of JavaScript, which then triggered the error we saw.

There's still a question of how anyone had the old version of WoW's HTML, though, since it's not like this JS is being requested long after the HTML page is originally loaded (afaik we're not doing code splitting in WoW).  But I think it could still occur, perhaps, in the small window of time that exists between a browser retrieving the initial HTML content from WoW and when it went out to fetch the JS referenced by it.

In other words, this is one potential sequence of events:

1. User's browser requests a HTML page from WoW and gets the initial HTML.
2. We push a new version of WoW to production.
3. The user's browser sees the (now outdated) script tag in the HTML it got in step 1 and attempts to retrieve it, but gets an HTML page instead of a JS file. Its JS parser barfs and Rollbar logs an error.

## Solutions

There's a number of solutions to this problem.  Here's two:

1. **Push WoW's static files to a separate static file server or CDN.** This is what the tenants app currently does.  The advantage of it is that the existing files on the external service don't get obliterated every time we push; so `1.ffb84278.chunk.js` will still exist once we push a new version to production.  I think this is the most preferable solution, because it allows us to aggressively cache the static files, which means we can scale more easily.  However, this will require a bit of work because we need to integrate it into our deployment process.

2. **Don't include a content hash in the JS bundle filenames.**  This is more of a quick fix and I'm not even sure if it's possible to tell create-react-app to not include the hash in the filename, based on their [Creating a Production Build](https://facebook.github.io/create-react-app/docs/production-build) docs. However, this could still easily result in weird problems down the line--unpredictable things could happen if e.g. the old version of HTML loaded the _new_ version of JS.

So really, I think option 1 is our only solution, maybe.

toolness:
This should fix #169 by making the server return 404 for any paths that miss express' `static` middleware and look like static files.

### DIFF
--- a/server/express.js
+++ b/server/express.js
@@ -37,6 +37,14 @@ if (process.env.NODE_ENV === 'production') {
 
 app.use('/api', routes);
 
+const return404 = function(req, res) {
+  res.status(404).end();
+};
+
+// These will help prevent https://github.com/JustFixNYC/who-owns-what/issues/169.
+app.get(/^\/static\/.*/, return404);
+app.get(/^\/[0-9A-Za-z_.\-]+\.(js|json|ico|html)$/, return404);
+
 app.get('*', function(req, res) {
   res.sendFile(path.resolve(__dirname + '/../client/build/index.html'));
 });


sraby:
This looks awesome, Atul! thank you!

PR #190, opened 2019-12-02
Upgrade to React 16 and other dependencies.

### LINKED ISSUE #101: " > react-mapbox-gl@3.8.0" has incorrect peer dependency "mapbox-gl@0.44.0"
This keep getting logged when installing client dependencies, we should figure out what's going on and fix it.

### LINKED ISSUE #187: React typings are for the wrong version of React
I was working on #186 and started using React hooks, which TypeScript was fine with, only to discover runtime errors because the hook functions I was using didn't actually exist!  This is because our `package.json` is using typings for React 16, while we're actually using React 15.

### LINKED ISSUE #188: Upgrade to React 16
Agh, this is partly because React 15 will eventually be (or possibly already is) unsupported, but also because [some React libraries _claim_ to support React 15 but actually don't](https://github.com/lingui/js-lingui/issues/592), and figuring that out can take a really long time!

toolness:
This is an attempt to fix #188, which gets us to React 16.8--the same version of React as the tenant platform, and the one right before all the incredibly annoying lifecycle deprecation warnings for third-party packages we don't control.

It also fixes #101 and fixes #187.

## Notes

* Right now if we upgrade react-google-maps to the earliest version of the package to support React 16, the property detail page is exploding with `Did you render <StreetViewPanorama> as a child of <GoogleMap> with withGoogleMap() HOC?`. It seems like like package went through some big API changes.

  For now I am just keeping our old version, which _claims_ it doesn't support React 16 but actually seems to work fine with it, at least insofar as our use case is concerned. The only annoying thing about this is that we get an "incorrect peer dependency" warning from yarn.  That said, it's now the _only_ such warning from yarn, so maybe it's still a net win, or at least a break-even? Lol.

* This upgrades us to `set-value` 2.0.1, resolving our vulnerability to [CVE-2019-10747](https://github.com/advisories/GHSA-4g88-fppr-53pp).

## To do

- [x] Make sure we remove the lingui workaround mentioned in https://github.com/JustFixNYC/who-owns-what/issues/188#issuecomment-557522108.
- [x] Kick the tires on this _a lot_, since we are updating lots of dependencies.
- [ ] If we end up merging this w/ the google maps workaround (bullet point 1 in notes) then file an issue to resolve it later, w/ notes on why the warning is there now.


toolness:
Hey @sraby are we thinking of dropping the panorama street view thing in WoW because of the whole "not enough credits" thing?  I'm running into an issue w/ upgrading to React 16 because the version of [react-google-maps](https://github.com/tomchentw/react-google-maps) that supports React 16 changed its API a bunch and I am not sure if it is worth the effort upgrading if we are going to drop it soon anyways.

toolness:
@sraby err I found a potential workaround for the react-google-maps thing, see the first bullet point under "notes" in the PR description.

toolness:
[ch1143]

sraby:
Hey Mr. Atul! Just updated your PR branch with master and kicked the tires a lot. Looks great, except for two specific snags that I found: 

1. The react-modal package seems to print this warning in the console: 

Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`.

1. "Export Data" oddly downloads an empty CSV that just says `{"error":"403 - \"Authentication parameters missing\""}`

Other than that, things looks good.

sraby:
Hmmm... so after further inspection, it looks like the "Export Data" issue is actually some issue that's only occurring in development and doesn't have anything to do with this specific PR. See issue #194

toolness:
Roight, so I've fixed the modal issue in 39b15e9afebab1a73443cdc29f04ab91101248ad, is all good now?  I will try investigating the funky download issue but it sounds like that was not introduced in this PR.

toolness:
Umm so I just tried exporting my data and it seems to work fine on my local system. Huh.

toolness:
Oh _weird_, I think I figured it out--will post about it in #194 though.  Anyhow this PR should be good to merge!

sraby:
Looks great! Just kicked _alll_ of the tires and everything looks ok. Merging this now.

PR #201, opened 2019-12-04
Only create StreetView when we know it's visible.

toolness:
This fixes the issue mentioned in https://github.com/JustFixNYC/who-owns-what/pull/200#issuecomment-561656193 whereby we actually fully render _two_ street view widgets per detail page--one for the mobile view and one for the desktop view.

This uses an [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to only instantiate the widget if we know it's going to be visible.

However, Intersection Observer is a newer browser API that's not available on all browsers, such as IE11.  If it's not available, we simply fall back to the old behavior of just rendering the street view widget.  Since a small minority of browsers won't have the API, I think this will still be a net win.

### DIFF
--- a/client/src/components/DetailView.js
+++ b/client/src/components/DetailView.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { CSSTransition } from 'react-transition-group';
 import { LocaleLink as Link } from '../i18n';
 import { StreetView } from './StreetView';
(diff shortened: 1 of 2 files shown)


sraby:
Net win for sure!! thanks so much, can't wait to tell Georges and Johanna

PR #202, opened 2019-12-06
Convert more code to typescript, and fix a few minor bugs

toolness:
This converts more of our code from JS to TS and also fixes a few minor bugs.

## Minor bugs fixed

* The `<LegalFooter>` component wasn't using the `<LocaleNavLink>` component, and as such, links to the privacy policy and terms of use weren't being properly prefixed with the current locale.
* The API client's `postNewSubscriber` function was sending a mysterious `list` parameter to the server that the server wasn't actually doing anything with (and which was always set to `undefined` anyways), so this parameter has been removed.
* The `helpers.jsonEqual()` function was comparing the JSONified representation of objects, but this doesn't work if objects are equal but have their keys ordered differently (e.g. `{a: 1, b: 2}` and `{b: 2, a: 1}`)).  I fixed this by wrapping `assert.deepEqual()`.
* Some pages use an obsolete `<nobr>` tag that TypeScript doesn't like; I've replaced this with a new `<Nobr>` component that trivially implements the [MDN workaround](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr).

## Suggestions

Since this is the first time I've looked at some of the code, I have a few suggestions.

* The `helpers.maxArray()` function seems to have some funky behaviors, in that it returns 0 if the array it's given is empty, and if it's given an array of entirely negative numbers, it will return 0.  Also, there's a built-in function called [`Math.max()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max), can we just use that?
* The `helpers.uniq()` function appears to also clone the objects it's given, and assumes that they're JSON-serializable... we might want to rename the function to make these assumptions explicit, to something like e.g. `uniqCopy()`.
* In the `helpers` and `mapping` modules, rather than exporting a big default object with a bunch of functions as keys, it might be better to export each function individually: aside from making it easier to import each individual function in an editor like VS Code, it can also help with tree-shaking.  The existing behavior will still work via e.g. `import * as helpers from "helpers"` too.
* The `<BBLPage>` makes API requests on component mount and update, and calls `setState()` when the API returns data, but it doesn't check to make sure that the component is still mounted when it does this.  This could result in React logging a warning.  We might also want to make sure that we _don't_ set the state if the API request is stale (e.g., if a subsequent state/props change kicks off another API request, which renders the previous API request irrelevant).

## Other notes

* This adds `rollbar` to the project dependencies, but only for its typings, so that TS files calling `window.Rollbar` methods are properly type-checked.


client/src/components/ScrollToTop.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { withRouter, RouteComponentProps } from "react-router";
+
+const APP_BODY_SELECTOR = '.App__body';
+
+class ScrollToTop extends Component<RouteComponentProps> {
+  componentDidUpdate(prevProps: RouteComponentProps) {
+    if (this.props.location !== prevProps.location) {
+      const appBody = document.querySelector(APP_BODY_SELECTOR);
+      if (!appBody) {
+        throw new Error(`"${APP_BODY_SELECTOR}" does not exist!`);
```
toolness: Here we're just throwing a more informative error than the `Cannot set property 'scrollTop' of null` error that would've been thrown if the `querySelector` call failed (while also appeasing TypeScript).

client/src/components/APIClient.ts
```diff
@@ -76,25 +85,32 @@ async function verifyIsJson(response) {
...
-  const error = new Error(`HTTP Error ${response.statusText}`);
-  error.status = response.statusText;
-  error.response = response;
+  const error = new HTTPError(response);
   console.log(error); // eslint-disable-line no-console
   throw error;
 }
 
-function parseJSON(response) {
+class HTTPError extends Error {
```
toolness: Rather than creating a regular `Error` and then adding arbitrary properties to it, which TypeScript doesn't like, we're just going to create our own `Error` subclass instead.  This is a best practice anyways, since it allows exception handlers to check if the caught exception is an `instanceof HTTPError` and then do custom logic based on it.

client/src/components/APIClient.ts
```diff
@@ -1,6 +1,18 @@
 /* eslint-disable no-undef */
 
-function searchAddress(q) {
+type WithBoroBlockLot = {
+  boro: string,
+  block: string,
+  lot: string,
+};
+
+type Q = {
```
toolness: I have no idea what the `q` parameters semantically meant (maybe "query"? 🤷‍♂ ) so I just called the type for it `Q`.

client/src/components/PropertiesList.tsx
```diff
@@ -60,16 +77,19 @@ const PropertiesList = (props) => {
                 },
                 {
                   Header: 'Zipcode',
-                  accessor: 'zip',
+                  accessor: d => d.zip,
+                  id: 'zip',
```
toolness: react-table doesn't have great third-party typings--the `accessor` can be just a "string" or a function that takes an `Addr`, and as there's no type-checking on the string, I opted to make things a bit more verbose by providing a function instead.  This triggered an error from react-table claiming I had to also provide a string `id` if `accessor` was a function, so I did that, even though I don't really understand what the `id` is for.

client/src/components/ScrollToTop.tsx
sraby: noice!

client/src/components/PropertiesList.tsx
sraby: Hmmm, yeah not sure what the id field is either. I'd assume it's used in the html structure somewhere but I don't see it 🤷‍♂

PR #203, opened 2019-12-10
[WIP] Pull content of long-form pages from Contentful

toolness:
Some of our pages, like "About" and "How it works", contain long-form content that will be easiest to localize in Contentful, rather than Crowdin or another tool that isn't rich-text aware.

This PR is an attempt to have us pull localized content for such pages from Contentful.  However, unlike our Gatsby-based sites, I'd like it to be _optional_ for developers to require access to Contentful in order to develop the site, rather than a requirement, particularly because the amount of long-form content is very minimal, well-defined, and unlikely to be updated frequently.

## Notes

* This PR borrows liberally from https://github.com/JustFixNYC/justfix-website/pull/63.

## To do

- [x] Save the contentful entries to JSON and render them with something like [@contentful/rich-text-react-renderer](https://github.com/contentful/rich-text/tree/master/packages/rich-text-react-renderer).
- [x] Show a different page depending on the current locale.

### DIFF
--- a/client/.env.local.sample
+++ b/client/.env.local.sample
@@ -39,3 +39,13 @@ REACT_APP_ROLLBAR_ACCESS_TOKEN=
 # than English to users.
 
 REACT_APP_ENABLE_PUBLIC_FACING_I18N=
+
+# =========================
+# CONTENTFUL INTEGRATION (OPTIONAL)
+# =========================
+#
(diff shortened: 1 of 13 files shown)


toolness:
Ok I am going to close this and instead build this stuff atop #205.

PR #215, opened 2020-01-16
Internationalize the rest of WoW

toolness:
## To do

- [x] Remove `assets/img/howitworks.jpg`, since this is now hosted by Contentful.
- [x] Get rid of `<Trans>` tags around proper names of real estate companies
- [x] Internationalize `components` directory files:
  - [x] `Subscribe.tsx`
  - [x] `SocialShare.tsx`
  - [x] `PropertiesMap.js`
  - [x] `PropertiesList.tsx`
  - [x] `OwnersTable.tsx`
  - [x] `LegalFooter.tsx`
  - [x] `Indicators.js`
  - [x] `EngagementPanel.tsx`
  - [x] `DetailView.js`
  - [x] `AddressToolbar.tsx`
  - [x] `AddressSearch.tsx`
- [x] Internationalize `containers/NychaPage.js`.
- [ ] Figure out what to do about `containers/PrivacyPolicyPage.tsx` and `containers/TermsOfUsePage.tsx`.
- [ ] We should figure out what we want to do with date formatting.  Right now it seems like we want to always format it in en-US (MM/DD/YY) but arguments could be made to just use ISO syntax (YYYY-MM-DD) since it's unambiguous and kind of jives with the current Windows-95-hacker aesthetic of the site. Regardless, we can boot to a separate issue.
- [x] Figure out if we want to internationalize Boro-Block-Lot.
- [x] Re-run `yarn lingui extract` and make sure the extracted messages make sense, and that there's no near-duplicates (e.g. `Portfolios` and `Portfolios:`).  **Update:** see https://github.com/JustFixNYC/who-owns-what/pull/215#issuecomment-576893622.
- [x] Oof, about halfway though doing this I realized that `<LocaleNavLink>` only localizes the _path_ of the URL, not the actual link text.  Will want to check the actual text content to make sure it's wrapped in `<Trans>`.


client/src/containers/HomePage.tsx
```diff
@@ -117,56 +118,56 @@ class HomePage extends Component<HomePageProps, State> {
...
-            <h5 className="text-center">... or view some sample portfolios:</h5>
+            <h5 className="text-center"><Trans>... or view some sample portfolios:</Trans></h5>
             <div className="container">
               <div className="columns">
                 <div className="column col-4 col-sm-12">
                   <div className="HomePage__sample">
                     <h6>
                       <Link to={this.state.sampleURLs[0]} onClick={() => {window.gtag('event', 'example-portfolio-1-homepage');}} >
-                        Kushner Companies / Westminster Management
+                        <Trans>Kushner Companies / Westminster Management</Trans>
```
toolness: @sraby should these company names be localized, or is it OK for them to just be in English?

client/src/containers/HomePage.tsx
```diff
@@ -117,56 +118,56 @@ class HomePage extends Component<HomePageProps, State> {
...
                 <div className="column col-4 col-sm-12">
                   <div className="HomePage__sample">
                     <h6>
                       <Link to={this.state.sampleURLs[1]}  onClick={() => {window.gtag('event', 'example-portfolio-2-homepage');}} >
-                        A&E Real Estate
+                        <Trans>A&amp;E Real Estate</Trans>
                       </Link>
                     </h6>
                     <Link to={this.state.sampleURLs[1]} onClick={() => {window.gtag('event', 'example-portfolio-2-homepage');}} >
-                      <img className="emassoc img-responsive" src={aeLogo} alt="E&M Associates" />
```
toolness: Looks like this was a typo?  Was this originally an E&M Associates portfolio that got changes to A&E?

client/src/containers/HomePage.tsx
sraby: definitely ok for them to just be English! Shouldn't all proper names stay in English?

client/src/containers/HomePage.tsx
sraby: oh yes, thanks for catching that! I changed that to A&E right before break

client/src/containers/HomePage.tsx
toolness: Yeah, that's what I thought... I wasn't sure if it would be more appropriate to e.g. replace "management" with the localized translation of that word or something, but if people only ever see it written "a&e management" elsewhere then it makes sense to keep it the same for us too!

client/src/components/IndicatorsDatasets.tsx
```diff
@@ -0,0 +1,123 @@
...
+   * The localized name for a particular "quantity" of the dataset, e.g.
+   * "15 HPD Complaints issued since 2014".
+   */
+  quantity: (i18n: I18n, value: number) => string,
+
+  /**
+   * A localized explanation for what the dataset means, and where to find more information.
+   */
+  explanation: (i18n: I18n) => JSX.Element,
+}
```
toolness: @sraby I factored out a new module, `IndicatorsDatasets`, which abstracts out the concept of a "dataset" to make the indicators visualization code a bit more DRY and less branchy.  Right now it mostly only deals with localization-related data/functionality for datasets but it could be expanded to provide more in the future.

toolness:
Ok @sraby, there was a question of what to do about some text that appears in multiple places, but slightly differently: the biggest version of this was where we had a phrase followed by a colon in one place, but just the phrase on its own in another place, yet both places had a very similar context.  An example of this is the string "Share this page with your neighbors", which sometimes appears with a colon after it, such as on the NYCHA page:

> image

and sometimes doesn't, such as on the property details page:

> image

Given that the consensus on internationalizing strings appears to be to [leave punctuation as part of the localizable string wherever possible](https://softwareengineering.stackexchange.com/a/234858), as opposed to e.g. writing code that displays the localized string followed by a colon, I opted to actually **remove colons from the UI entirely** where it made sense, so as not to have two extremely similar strings to localize (one with the colon and one without).  So in the above case I actually just removed the trailing colon from the NYCHA page, so there's no question of whether to include the colon or not, and rather the meaning can be derived from the semantic markup/visual styling. Hopefully this is an OK path to take.

toolness:
Ok @sraby, with the major outstanding question of what to do about the privacy policy and TOU pages, I think this is ready for your review!

client/src/components/IndicatorsDatasets.tsx
sraby: Ok cool! I should've warned you about a WIP branch I have called [timeline-page-v2-updates](https://github.com/JustFixNYC/who-owns-what/tree/timeline-page-v2-updates) that also abstracted out the concept of a "dataset" wherever possible in the Indicators code. Im sure when the time comes we can figure out the most strategic way to merge

sraby:
Amazing work, @toolness! Thank you so so so much. Will review shortly :)

sraby:
@toolness I merged this and also preserved the two remaining "To do" checklist items ass issues #217 and #218.

PR #228, opened 2020-03-12
Add docs about updating data

toolness:
#209 was a bit challenging because the process of updating WoW's data wasn't very well-documented!  This attempts to add a bit of documentation about the whole shebang so it's not as hard next time around.


sraby:
This looks great to me @toolness! Very clear and helpful for future data updates.

PR #236, opened 2020-05-06
Avoid division by zero when getting summary

### LINKED ISSUE #231: Some buildings have 0 residential units, according to PLUTO
Example: whoownswhat.justfix.nyc/en/address/MANHATTAN/343/WEST%2086TH%20STREET 

WOW seems to handle these cases pretty well, except for when the portfolio consists only of 0 residential unit buildings, in which case our aggregate function for the "Summary Tab" seems to crash

sylvansson:
PLUTO indicates that certain buildings have 0 residential units. This causes two different divisions by zero when fetching a portfolio's summary. This PR updates the `get_agg_info_from_bbl` function to avoid these divisions by zero and return NULL instead.

This is my first contribution to this project, so let me know if I missed anything.

Close #231


sylvansson:
I see that the frontend doesn't handle NULLs (it displays nothing) for the number of open violations per unit. I'll fix that right away.

sylvansson:
Here we go. The frontend now displays a different message when the number of open violations per unit is not available:

image

sraby:
@sylvansson thanks so much for this bug fix! I will give this a good review early next week 👍

sylvansson:
Great, thanks @sraby. Just fixed the conflicts with your latest changes.

client/src/components/PropertiesSummary.js
```diff
@@ -8,15 +8,11 @@ import APIClient from "components/APIClient";
 import "styles/PropertiesSummary.css";
 import { EvictionsSummary } from "./EvictionsSummary";
 import { RentstabSummary } from "./RentstabSummary";
+import { ViolationsSummary } from "./ViolationsSummary";
```
sraby, line 11: Nice! Thank you for refactoring this piece of the Summary tab into its own component— it's a lot easier to follow along that way.

client/src/components/ViolationsSummary.tsx
```diff
@@ -0,0 +1,58 @@
...
+}> = props => {
+  const { openviolationsperresunit, totalviolations } = props;
+
+  // openviolationsperresunit is not defined when the portfolio has 0 residentials
+  // units in PLUTO, so we can't calculate a proportion.
+  const openViolations = _.isNil(openviolationsperresunit) ? (
+    <Trans>
+      The average number of open HPD violations per residential unit is not available for this
+      portfolio.
+    </Trans>
```
sraby, line 22: Hmm, I wonder if instead of saying that this statistic is "not available" here, we could just remove this line altogether. In my experience, having HPD-registered buildings with 0 residential units either means there is an error in PLUTO's tally, or there are actually no residential units and the building had to register for some other reason (or was registered at a time when there used to be residential units). Therefore, it may be simpler just to remove this line for portfolios of 0 units, as finding the average by unit just wouldn't make logical sense. Happy to hear your thoughts on this though!

sylvansson: That seems fair. I'll remove it.

sylvansson:
I did some QA and everything seems to work as intended. Let me know if you need me to do anything else for this issue.

sraby:
Thank again for this @sylvansson!

sylvansson:
My pleasure @sraby!

PR #237, opened 2020-05-13
Add urls and focus styles to address page tabs

sraby:
This PR improved the accessibility and usability of Who Owns What by assigning a specific url path to each of the four tabs we include on the address page. Now users can tab through links to navigate between these pages as well as share each individual URL allowing another user to have a direct link to that page. 

Specifically, this PR:
- [x] Adds unique URL paths to each address page tab
- [x] Verifies that direct links (and direct links via bbl) still work
- [x] Converts all previous tab changes to internal links
- [x] Adds focus class for tabs

### DIFF
--- a/client/src/containers/AddressPage.js
+++ b/client/src/containers/AddressPage.js
@@ -181,6 +181,7 @@ export default class AddressPage extends Component {
                   >
                     <Link
                       to={this.generateBaseUrl()}
+                      tabIndex={this.props.currentTab === 0 ? -1 : 0}
                       onClick={() => {
                         if (
                           Browser.isMobile() &&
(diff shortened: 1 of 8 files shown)


client/src/containers/AddressPage.js
```diff
@@ -139,84 +159,154 @@ export default class AddressPage extends Component {
...
-                  <li className={`tab-item ${this.state.currentTab === 0 ? "active" : ""}`}>
-                    <a // eslint-disable-line jsx-a11y/anchor-is-valid
+                  <li
+                    className={`tab-item ${
+                      this.props.currentTab === 0 ? "active" : ""
+                    }`}
+                  >
+                    <Link
+                      to={this.generateBaseUrl()}
+                      tabIndex={this.props.currentTab === 0 ? -1 : 0}
```
sraby, line 184: I added this `tabIndex` prop here so that folks tabbing through the tabs (hah!) won't be prompted to click on the tab they are currently viewing. 

I also acknowledge that this isn't the most DRY solution in checking our `currentTab` prop, but since there are so many more severly non-DRY implementations throughout this component, I thought we should save refactoring for when we do a full sweep of the codebase.

sraby:
Agh, so sorry @toolness! That was not thoughtful of me. Will definitely not run any reformatting on future PRs that need your review.

PR #247, opened 2020-05-20
Fixed general spacing issues created by new Spanish translations

sraby:
This PR makes various styling tweaks to WOW to accommodate our new Spanish translations, which are generally much longer strings of text than their English counterparts.


client/src/components/BuildingStatsTable.js
```diff
@@ -0,0 +1,148 @@
...
+  const YearBuilt = () => (
+    <div
+      title={i18n._(
+        t`The year that this building was originally constructed, according to the Dept. of City Planning.`
+      )}
+    >
+      <Trans render="label">Year Built</Trans>
+      {props.addr.yearbuilt !== 0 ? props.addr.yearbuilt : "N/A"}
+    </div>
+  );
```
sraby, line 46: @toolness something about defining these components as constants within the larger component felt like I was doing something very very wrong! At the same time, this felt like the DRYest solutions as it meant I didn't need to pass down properties to the smaller components at al. Let me know if this is kosher!

sraby:
@toolness thanks again for your review on this! Contexts worked out a lot more smoothly than the original implementation :)

PR #257, opened 2020-05-26
Add page titles to each core page on Who Owns What

sraby:
This PR adds page titles to each page of Who Owns What using React Helmet. More accessibility concerns, such as adding [Aria Live announcements](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) will be addressed in later PRs.


client/src/components/Page.js
```diff
@@ -0,0 +1,55 @@
...
+import { i18n } from "@lingui/core";
+import helpers from "../util/helpers";
+
+const metadataWithoutI18n = {
+  keywords: i18n._(t`pay rent, rent, can't pay rent, may rent, may 1`),
+  description: i18n._(
+    t`Use this free tool from JustFix.nyc to research your building and investigate landlords! We use property ownership mapping to identify a landlord's portfolio and provide data to indicate potential tenant harassment and displacement.`
+  ),
+  siteName: i18n._(t`Who owns what in nyc?`),
+};
```
toolness, line 14: Hmm, so I was about to say that this top-level use if `i18n._()` isn't good, but then I realized that unlike the Tenant Platform, WoW doesn't do server-side rendering, so as long as we assume that the user's page will need to reload when they go from one locale to another (e.g. when they toggle languages it results in a full page reload, not just a react router history push) you should be ok.

That said, if we _do_ want to move to a SSR solution similar to the one that tenants2 has, this kind of code will make that harder to do, so it's up to you whether you want to wrap this kind of code in a function or not!

client/src/components/Page.js
```diff
@@ -0,0 +1,55 @@
...
+
+const getMetadata = () => withI18n()(metadataWithoutI18n);
```
toolness, line 16: Oh funky, I'm not sure if this function is actually doing anything useful, since `metadataWithoutI18n` is already a constant at this point, so its localized values are already in-place at the time the function is called?  Or am I misreading something...

client/src/components/Page.js
sraby: Oh shoot. I thought that since I was actually rendering the `metadata` component as a function and not a constant, we'd be ok here but it sounds like we need to also make `metadataWithoutI18n` into a function as well?

client/src/components/Page.js
sraby: Ooops, yeah you're right. I think I swapped my logic here. The `metadataWithoutI18n` component should be made into a function, not this here. This goes along with my response above.

client/src/components/Page.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Helmet } from "react-helmet";
+import { t } from "@lingui/macro";
+import { withI18n, withI18nProps } from "@lingui/react";
+import helpers from "../util/helpers";
+
+const metadata = {
+  keywords: t`Landlord, Portfolio, Tenant, Displacement, Map, JustFix, NYC, New York, Housing, Who Owns What`,
+  description: t`Use this free tool from JustFix.nyc to research your building and investigate landlords! We use property ownership mapping to identify a landlord's portfolio and provide data to indicate potential tenant harassment and displacement.`,
+  siteName: t`Who owns what in nyc?`,
+};
```
sraby, line 11: Hey @toolness! Check out this implementation of the i18n of our page metadata— it utilizes lingui's notion of [Lazy Translations](https://lingui.js.org/tutorials/react-patterns.html?highlight=strings#lazy-translations). Was wondering if you ever saw this and have concerns about this pattern, as the conciseness of it almost seems too good to be true. That being said, everything seems to be behaving as it should in terms of these strings getting translated 🤷‍♂️

client/src/components/Page.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const FB_APP_ID = "247990609143668";
+const TWITTER_HANDLE = "@JustFixNYC";
+const ORIGIN_URL = "https://whoownswhat.justfix.nyc/";
+const SHARE_IMAGE_URL = "https://i.imgur.com/6WL74DZ.png";
+
```
sraby, line 17: Saving these as constants triggered me to record some issues to address later: #268, #269, and #270

client/src/components/Page.tsx
toolness: oh this is super dope and actually makes the whole `i18n._(t`blah`)` thing make a lot more sense (I always thought it was kind of verbose).  NOICE

client/src/components/Page.tsx
toolness: nooooooooooooooice i dig it!!  Also I find the term `ORIGIN_URL` a tiny bit confusing b/c I think of the word "origin" in the same sense as the [`Origin` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin), which contains no path information, so I would expect something named `ORIGIN_URL` to not contain the trailing slash at the end.

client/src/components/Page.tsx
```diff
@@ -0,0 +1,58 @@
...
+  title?: string;
+  children: React.ReactNode;
+};
+
+const Page = withI18n()((props: PageProps & withI18nProps) => {
+  const i18n = props.i18n;
+  const title = props.title && helpers.titleCase(props.title.trim());
+  const fullTitle = title
+    ? `${title} | ${i18n._(t`Who owns what in nyc?`)}`
+    : i18n._(t`Who owns what in nyc?`);
```
toolness, line 28: This is probably outside the scope of this PR but is there a reason we're not capitalizing NYC here?  I'm wondering if it might also be more likely that a screen reader would pronounce it as letters rather than as some kind of bizarro word that sounds like "nyuk".  I dunno though.

client/src/containers/App.js
```diff
@@ -17,6 +17,22 @@ import ScrollToTop from "components/ScrollToTop";
 import Modal from "components/Modal";
 import SocialShare from "components/SocialShare";
 import MethodologyPage from "./Methodology";
+import { withI18n } from "@lingui/react";
+
+const HomeLink = withI18n()((props) => {
+  const { i18n } = props;
+  return (
+    <Link
+      aria-label={i18n._(t`Who owns what in n y c?`)}
```
toolness, line 26: Probably also out of scope but is there a reason we're not using "NYC" instead of "n y c"?  (I assume the reason for the spaces is so screen readers pronounce each letter?)

client/src/components/Page.tsx
sraby: Ahh yeah, I don't think there's any reason why we need to have that trailing slash there. Let me remove it.

client/src/components/Page.tsx
sraby: It's fully a styling choice (from the old days). I think we should let this sit here until the rebrand happens. And yes, you're right— VoiceOver pronounced it as "nick". That is precisely why I added the aria-label below to make it pronounceable for screen readers.

client/src/containers/App.js
sraby: That is exactly it. Let me add a comment to make that clearer.

PR #271, opened 2020-05-27
Added new styling and html properties for accessibility

sraby:
This PR addresses some of the accessibility concerns covered in our [WOW Accessibility Audit](https://docs.google.com/document/d/1wtyZ81DCZroVavxE0iMnA6oWAH28VPqIJ-TjbirKvYU/edit#). Mainly, this focuses on styling and html property fixes.

Still to do:
- [ ] Configure aria live announcements for new page renders
- [ ] Configure new focus elements for certain page transitions that lose keyboard focus


client/src/components/Indicators.js
```diff
@@ -392,6 +392,9 @@ class IndicatorsWithoutI18n extends Component {
 
                 <div className="Indicators__viz">
                   <button
+                    aria-hidden={
+                      this.state.xAxisStart === 0 || this.state.activeTimeSpan === "year"
+                    }
```
toolness, line 397: Oh neat, I didn't realize you could actually just pass a boolean to this prop. In the tenants platform I have a helper function that converts a boolean to the strings `"true"` and `"false"`, I think.

client/src/components/Indicators.js
```diff
@@ -406,6 +409,10 @@ class IndicatorsWithoutI18n extends Component {
                   </button>
                   <IndicatorsViz {...this.state} />
                   <button
+                    aria-hidden={
+                      this.state.xAxisStart + this.state.xAxisViewableColumns >= xAxisLength ||
+                      this.state.activeTimeSpan
+                    }
```
toolness, line 415: Errr wait, one thing I'm realizing here though is that ... uh ... you're not in TypeScript.  The TypeScript definitions for `aria-hidden` are that it be either a boolean, undefined, `"true"` or `"false"` but it seems like `this.state.activeTimeSpan` can be a string like `"year"`, in which case you might inadvertently be setting the actual value of the `aria-hidden` attribute to `"year"`, which would make no sense to browsers.  You might want to check to see if React is converting the string value to `"true"` under-the-hood, in which case you don't need to worry, or if it's just passing it through to the underlying DOM value, in which case it will confuse screen readers.

toolness: Er the alternative of course is just to change `this.state.activeTimeSpan` to `!!this.state.activeTimeSpan` so it's coerced to a boolean.

Ah this makes me miss TypeScript!!

toolness:
An aside: one thing I'm realizing is that internationalization is kinda providing us with an additional motivation to split our big files up into smaller ones... with big files, there's a good chance that a minor code change might result in the line numbers of lots of translations being changed, resulting in a very noisy diff, but with smaller files that's less likely to happen!  (Or rather, when it does happen, hopefully only a few strings will have their line numbers changed, rather than lots.)

client/src/components/Indicators.js
sraby: ooooh so this is actually a mistake! We're actually checking to see if `this.state.activeTimeSpan` is equal to "year", not checking its existence.

client/src/components/Indicators.js
sraby: Yeah! Same with tabIndex too I think :)

sraby:
@toolness I'm going to push this now, but I'm saving all outstanding issues from the Audit [in this Milestone](https://github.com/JustFixNYC/who-owns-what/milestone/2) so we can keep track.

PR #284, opened 2020-06-11
Initialize GTM in React.

toolness:
This adds a `<GoogleTagManager>` component, which does the following:

```js
/**
 * Google Tag Manager reports the title of the landing page when it
 * initializes, but because we only render the <title> on the client-side
 * via React, we need to wait until it's rendered before we initialize
 * GTM.
 *
 * This inserts the GTM script tag on mount, which is assumed to also
 * be when the <title> tag should be set. This should ensure that
 * GTM reports the correct landing page title.
 */
```

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -66,8 +66,14 @@
     // End Rollbar Snippet
     </script>
 
-    <!-- Global site tag (gtag.js) - Google Analytics -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-67069242-5"></script>
+    <!--
+      Global site tag (gtag.js) - Google Analytics
+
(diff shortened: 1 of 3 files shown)


toolness:
Hmm I don't think so--especially not Rollbar, since there could be errors that occur while the app is initializing, which we want Rollbar to tell us about--so we don't want to block Rollbar on something that could go wrong!

Similarly, for FullStory I could see how it might actually be useful for us to see when the app goes from a "blank page" to a full UI.  If e.g. the user's CPU is super slow and our app initialization is CPU-intensive, that might mean that we see the delay in a fullstory session, alerting us to the problem... otherwise if we only init fullstory after successful app init we'd never see that...

Still, for other things it might make sense!

PR #287, opened 2020-06-22
Reconfigure GTM tag to load with React Helmet component

sraby:
So, it turns out that [React Helmet is very finicky](https://stackoverflow.com/questions/37922183/can-react-helmet-inject-a-javascript-object-into-head-tag) when it comes to adding script tags to it to render in the <head> of a given page. This PR refactors the logic within the `<GoogleTagManager>` component so that it's now part of the `<Page>` component and therefore can properly add our GTM code as a script property to our instance of React `<Helmet>`.


client/public/index.html
```diff
@@ -21,15 +21,6 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
 
-    <!-- Google Tag Manager -->
-    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-      new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-      j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-      })(window,document,'script','dataLayer','GTM-NMPT5JP');</script>
-    <!-- End Google Tag Manager -->
```
toolness, line 30: Err, so this should really stay because while the React app is initializing, we want anything that uses `window.dataLayer` to still work... with this gone, `window.dataLayer` will be undefined, so anything attempting to use GTM before the React app is initialized will throw an exception.  It also means that those initial data layer events that are added in this snippet (`gtm.start` etc) won't be transmitted to GTM.

Err wait, that said, I'm now noticing that this is also adding a `<script>` tag for `gtm.js` to the the page which isn't good... aghc

toolness: Holy cow this is a completely different script tag than the one I saw in https://github.com/JustFixNYC/who-owns-what/pull/284!  It looks like the one you're removing here was originally added 10 months ago in #179, and is explicitly Whole Whale-related, while the one I fiddled with in #284 was originally added two years ago in 5e7bccb40a1b5048c928d1fd453bfc6c853ba47b.  Ahhhh why were there two GTMs on our page

client/public/index.html
```diff
@@ -21,6 +21,11 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
 
+     <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+      new Date().getTime(),event:'gtm.js'});})(window,document,'script','dataLayer','GTM-NMPT5JP');</script>
+    <!-- End Google Tag Manager -->
```
toolness: This looks good but I wonder if we should just simplify it so it's more readable:

```js
window.dataLayer = window.dataLayer || [];
window.dataLayer.push({ "gtm.start": new Date().getTime(), event: "gtm.js" });
```

It might also be worth adding a comment mentioning that loading of the actual GTM script is done in React so that `<title>` is properly set by the time GTM loads, just so future readers won't get too confused?  I dunno.

client/public/index.html
```diff
@@ -71,7 +67,7 @@
 
       Note that we insert the GTM script tag from JS to ensure that
       GTM only initializes itself once we've set the title tag.
-      For more details, see GoogleTagManager.tsx in this repository.
+      For more details, see Page.tsx in this repository.
     -->
     <meta name="jf-gtm-id" content="UA-67069242-5">
```
toolness, line 72: Maybe we should move this and the HTML comment above it to the other GTM script tag above, so they're all in the same place?  Also, shouldn't the `content` attribute here be set to `GTM-NMPT5JP`?  Otherwise that bit of information is being completely lost, since the script tag above isn't actually using it...

sraby: Yes good point! Ok I'll set that. Wait, now that I look at it more closely, the actual URL we use to access the GTM code is different here than it is on our other sites. Org Site and EFNYC make a request to `https://www.googletagmanager.com/gtm.js?id=XXX-XXXXX`
when the current configuration on WOW, the part that renders after the title tag renders, makes a request to `https://www.googletagmanager.com/gtag/js?id=XXX-XXXXX`. Any ideas why that is?

client/public/index.html
sraby: For sure

client/public/index.html
sraby: I just did a quick diff between the code at both end points and they seem slightly different from each other. I wonder if one of the two is an older version of the GTM code? Hmmmm

sraby:
Ok @toolness wanna give this one last look over?

client/public/index.html
sraby: Ok so it looks like I should just use the more recently introduced URL structure for the GTM code, the one that WholeWhale provided for us.

client/public/index.html
```diff
@@ -21,14 +21,29 @@
...
+      Note that we insert the GTM script tag from JS to ensure that
+      GTM only initializes itself once we've set the title tag.
+      For more details, see Page.tsx in this repository.
+    -->
+    <meta name="jf-gtm-id" content="GTM-NMPT5JP">
+    <script>
+      (function (window, dataLayer) {
+        window.dataLayer = window.dataLayer || [];
+        window.dataLayer.push({ "gtm.start": new Date().getTime(), event: "gtm.js" });
+      })(window, document, "script", "dataLayer", "GTM-NMPT5JP");
```
toolness, line 36: Actually, I think you can just remove the [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) wrapper entirely and just have the two lines of the function run as top-level code... that will also avoid the confusion of the fact that the function's arguments aren't used at all too!

client/public/index.html
```diff
@@ -21,14 +21,29 @@
...
+    </script>
     <!-- End Google Tag Manager -->
-  
+
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'UA-67069242-5');
```
toolness, line 45: Wait shouldn't this just be removed?  This was Dan's old code from that commit two years ago, do we actually need it?  I think it just makes things more confusing...

sraby: So, I think that's something we may need to check in with WholeWhale about— not sure if they're still utilizing our old Google Analytics account as a backup or what not

toolness: Oh ok, I guess the thing I'm worried about is if GTM gets confused about seeing data layer events for two different accounts in it, using two different kinds of events?  (it seems like Dan's code adds a `js` event followed by a `config` event, while WW's just adds one `gtm.start` event.)

I think WW told me that they were fine with us still using our old analytics accounts but that it was entirely for our benefit rather than theirs... maybe ask them if it's ok for us to just remove the old stuff entirely?  I don't think it'd be a big deal if the legacy code used Google Analytics instead of an older version of Google Tag Manager--but the fact that it's the latter gives me all kinds of spielkus about GTM getting confused and not knowing which account to send events to...

sraby: Fair point. Asking them about it now!

Issue #276, opened 2020-06-01
Add "aria-live" announcements to key page/tab transitions.


sraby:

toolness:
Note that this kind of thing will also be useful for our "subscribe to our newsletter" form too!

PR #288, opened 2020-06-23
Refactored Mailchimp sign-ups to use new external API

sraby:
This PR replaces our current API for Mailchimp email signups and instead uses the [new API endpoint implemented in the Tenant Platform](https://github.com/JustFixNYC/tenants2/pull/1549/files)! It makes a note that the current server functions to handle Mailchimp signups are deprecated (see #289).


toolness:
Oh, I should also note that while the issue wasn't introduced in this PR, there is still the complication that because we don't have any kind of ARIA live support (#276), screen reader users who fill out this form will be confused because they won't hear any of the response messages.  Oof!

PR #290, opened 2020-07-06
Utilize `DATABASE_URL` env variable in docker compose

sraby:
This PR configures our `docker-compose.yml` file to reference the `DATABASE_URL` env variable when constructing the virtual environment, which allows the developer to dynamically override the database reference if they want to. It also updates our `env.sample` to include the correct default value for this env variable if the user follows the basic Docker instructions on the ReadMe.


sraby:
@toolness when working on this, I noticed that our test data for Docker-based development is a bit out of date with new features. Documented that issue in https://github.com/JustFixNYC/who-owns-what/issues/291

PR #292, opened 2020-07-07
Make running Prettier a build requirement

### LINKED ISSUE #240: Make sure Circle-ci build's fail if code isn't all run through prettier.
Now that we have prettier configured throughout the codebase (thanks to #239), let's make sure we have linting in place to force all code to be pretty for future changes to the codebase.

sraby:
This PR adds a new prettier check to the client build and test scripts that fails if there are any files in client or server that are not properly linted via Prettier. Fixes #240 .


client/package.json
```diff
@@ -60,14 +61,15 @@
   },
   "scripts": {
     "lingui": "lingui",
+    "prettier:check": "cd .. && prettier \"**/*.{ts,tsx,js,scss}\" --check && cd client",
     "contentful": "node pull-contentful-data.js",
```
sraby, line 65: This is not the most DRY solution (as it literally just copies the prettier:check command from the root directory) but given that we are working within a package.json file, I couldn't think of any way to optimize this. Any thoughts?

toolness: yeah i think it is fine! i blame having two package dot json's which is not your fault!

toolness: oh wait i do have one idea, what if you actually changed it to `cd .. && yarn prettier:check && cd client`?

sraby: I tried that and it didn't recognize `yarn prettier:check` even though it's defined in the parent directory's package.json... it seems like some linting mechanism only allows for references to functions that are defined within the same file?

sraby: Oh wait! I just tried it again and it works :). Maybe I was running `cd .. && prettier:check && cd client` before, without the `yarn` which would make sense why it wouldn't be recognized.

PR #293, opened 2020-07-07
Add documentation on duplicate package.json files and env files

### LINKED ISSUE #260: Figure out how to reduce confusion around repository structure (two package.json files)
It is super confusing that we have two separate `package.json` files, one in the route directory and one in `/client`. However, this is apparently the recommended structure for projects using [Create React App](https://reactjs.org/docs/create-a-new-react-app.html). Let's figure out how to reduce confusion— maybe use something called [Yarn Workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) that allows us to have only one `node_modules` folder. Let's also check to make sure that there are no duplicate packages that don't need to be in both server and client.

### LINKED ISSUE #261: Make client env configuration less confusing
So, in our `/client` directory, we have both a `.env` (committed to git) and a `.env.local` (ignored by git). While this is part of the [Create React App](https://reactjs.org/docs/create-a-new-react-app.html) documentation, it can be quite confusing as we now have three areas that configure env variables (1 in root directory, 2 in client).

sraby:
This PR fixes #260 and fixes #261 by adding comments and readme documentation explaining our duplicate package.json files and env files. It also removes any duplicate packages that were accidentally added to either the server or client but don't need to be there.


sraby:
So, as of now, the only packages that are installed in both server and client are `lodash` and `rollbar`, which both make sense to me.

client/.env
```diff
@@ -1,2 +1,13 @@
...
+# ============================================
+# PRODUCTION CLIENT-SIDE ENVIRONMENT VARIABLES 
+# ============================================
+# ============================================
+#
+#
+# We commit some environment variables to git here 
+# for our production instance to reference. 
+# Head over to `env.local.sample` for more documentation:
+
```
sraby, line 11: Is this enough of an explanation here? To be honest, I'm still at a bit of a loss as to _why_ it's strategic to commit these vars to git, and didn't find very good explanations online...

client/.env.local.sample
```diff
@@ -1,3 +1,18 @@
...
+# =================================
+# =================================
+#
+# To see documentation on server-side variables, 
+# head over to the `.env.sample` file in the parent directory
+#
+# Note that some client-side variables are committed to git 
+# via the `.env` file in this directory. This follows the protocol
+# for the "Create React App" framework and allows certain public
+# variables to get used in production more easily.
```
sraby, line 13: Again, not sure if this is the clearest explanation of _why_ we have this extra client-side env file.

client/package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "//*":"NOTE: THIS REPOSITORY CONTAINS TWO PACKAGE.JSON FILES",
+  "//**":"FIND THE SERVER-SIDE PACKAGE.JSON FILE IN THE PARENT DIRECTORY",
```
sraby, line 3: Apparently this is a kooky way to add "comments" to json?

client/.env
toolness: Oh, I think the one advantage of committing them to git is that it makes it easier for developers, especially new contributors.

With the tenant platform, for instance, which has nothing like this, the disadvantage is that new contributors either need to sign up for a new account on every service they want to integrate with, or they need to get a copy of the `.env` file from another developer--which can be especially frustrating if the contributor is an open-source volunteer and doesn't _know_ any other developers on the project.

On the other hand, if the project can just have some default values for the environment variables without introducing any security holes, this can help new developers hit the ground running much quicker.

I guess another advantage is that, if these values are the same ones used in production, it can lead to better [dev/prod parity](https://12factor.net/dev-prod-parity) too.

client/package.json
toolness: haha yeah this is the one thing I _really_ hate about JSON!  If I could go back in time, I'd force the guy who invented it to support the doggone things.

README.md
```diff
@@ -17,9 +17,11 @@ This site is built on top of the critical work done by [@aepyornis](https://gith
 
 Backend logic and data manipulation is largely handled by making calls to PostgreSQL functions and prebuilding results into tables whenever possible to avoid complex queries made per-request. See the [hpd-registration ](https://github.com/nycdb/nycdb/tree/master/src/nycdb/sql/hpd_registrations) scripts of `nycdb` for the SQL code that provides this functionality.
 
+Note that both the backend and the frontend of the app each contain __separate `package.json` configurations__ as well as __separate sets of environment variables__. 
+
```
toolness, line 21: It might be useful to mention that this is the recommended way of doing things according to Create React App, to help convince the reader that we're not just being weird. 😉

client/.env
toolness: It might also help to point the reader to CRA's documentation on adding custom environment variables too, in part for their own edification but also to help convince them that we're not just being weird: https://create-react-app.dev/docs/adding-custom-environment-variables/

toolness: Err also it might be worth calling this file "DEFAULT CLIENT-SIDE ENVIRONMENT VARIABLES" rather than production?  I'm not sure what terminology CRA uses, actually, but as I mentioned in an earlier comment, I think the biggest benefit comes from these being useful development defaults, rather than necessarily production values (I might be wrong but I think that it might be possible to override these values with actual environment variables we set through e.g. Heroku, and in some cases we might end up doing that).

README.md
sraby: haha good point

client/.env
sraby: Ok great! This definitely makes more sense now. Also— this helps explain too why the one var we commit to git is precisely the one "required" env var we have on on the client side. All the other vars are optional, which means that a new contributor wouldn't need to do anything with those to get the codebase up and running.

PR #299, opened 2020-07-14
Refactor Routes

### LINKED ISSUE #263: Document and potentially refactor how we define all Routes
Currently, it's very confusing how all routes in WOW are generated and which pages correspond to which routes. Let's at the very least add documentation to explain this, but also consider refactoring the Route component definitions in [App.js](https://github.com/JustFixNYC/who-owns-what/blob/master/client/src/containers/App.js) into a new `Routes.tsx` file that allows us to easily track where each route is referenced. Since WOW relies a lot on parameterized routes, we would define new functions that generate the routes for you given parameters as arguments (I.e. `createBblLinkRoute(boro, block, lot) { ... }`)

sraby:
This PR fixes #263 by refactoring our Routes structure into its own file, complete with helper functions to help any redirects around our site link to the correct route. It also factors out all of the paths for routes of the site so we can reference them from a central place.


client/src/routes.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <Route
+      path={addressPageRouteWithParams + "/portfolio"}
+      render={(props) => <AddressPage currentTab={2} {...props} />}
+    />
+    <Route
+      path={addressPageRouteWithParams + "/summary"}
+      render={(props) => <AddressPage currentTab={3} {...props} />}
+    />
+    <Route path={localizeRoute("/bbl/:boro/:block/:lot")} component={BBLPage} />
+    <Route path={localizeRoute("/bbl/:bbl")} component={BBLPage} />
```
sraby, line 42: Since these parametrized routes are not actually generated anywhere in the code (for a Redirect or whatnot), I didn't bother making their own `createRoute...` function like I did with the Address Page Routes

client/src/components/SocialShare.tsx
```diff
@@ -98,9 +100,11 @@ const SocialSharePortfolioWithoutI18n: React.FC<{
...
       url={encodeURI(
-        `https://whoownswhat.justfix.nyc/address/${addr.boro}/${addr.housenumber}/${
-          addr.streetname
-        }${location === "summary-tab" ? "/summary" : ""}`
+        `https://whoownswhat.justfix.nyc${createRouteForAddressPage(
+          addr.boro,
+          addr.streetname,
+          addr.housenumber
+        )}${location === "summary-tab" ? "/summary" : ""}`
       ).replace(" ", "%20")} // Support for Android
```
toolness, line 108: Hmm I wonder if we should also factor out an `absoluteURL()` function that makes a relative route into an absolute one?  That way we also wouldn't have to hard-code the URL to WoW here... Also we should probably generate this URL's protocol/host from `window.location` so that it works regardless of the domain it's hosted on. I guess we could also do that in a separate PR if you want though.  But it seems useful since there are a bunch of other places where `whoownswhat.justfix.nyc` is hard-coded in this PR (obviously it's not your fault since it's from legacy code!).

Err also, I think you probably want to be using `encodeURIComponent()` for the street name and/or house number in  `createRouteForAddressPage`, rather than `encodeURI()` on the whole thing--that will _probably_ fix the "support for android" workaround you have in there, and even if it doesn't, I think that workaround should be done in `createRouteForAddressPage`, since that's really the source of any space characters that appear in the URI... Otherwise you'll have to do that workaround everywhere you use `createRouteForAddressPage()` which isn't very DRY!

client/src/containers/HomePage.tsx
```diff
@@ -74,9 +75,9 @@ class HomePage extends Component<HomePageProps, State> {
       searchAddress: makeEmptySearchAddress(),
       results: null,
       sampleURLs: [
-        "/address/BROOKLYN/89/HICKS%20STREET",
-        "/address/QUEENS/4125/CASE%20STREET",
-        "/address/BROOKLYN/196/RALPH%20AVENUE",
+        encodeURI(createRouteForAddressPage("BROOKLYN", "HICKS STREET", "89")),
+        encodeURI(createRouteForAddressPage("QUEENS", "CASE STREET", "4125")),
+        encodeURI(createRouteForAddressPage("BROOKLYN", "RALPH AVENUE", "196")),
```
toolness, line 80: If you follow my suggestion to use `encodeURIComponent()` in `createRouteForAddressPage` you won't have to use `encodeURI` here either.

client/src/routes.tsx
```diff
@@ -0,0 +1,70 @@
...
+import AddressPage from "./containers/AddressPage";
+import BBLPage from "./containers/BBLPage";
+import AboutPage from "./containers/AboutPage";
+import HowToUsePage from "./containers/HowToUsePage";
+import MethodologyPage from "./containers/Methodology";
+import TermsOfUsePage from "./containers/TermsOfUsePage";
+import PrivacyPolicyPage from "./containers/PrivacyPolicyPage";
+
+export const createRouteForAddressPage = (boro: string, streetname: string, housenumber?: string) =>
+  `/address/${boro}/${housenumber || " "}/${streetname}`;
```
toolness, line 13: this is where you can just wrap the arguments in `encodeURIComponent()` to avoid having to wrap the whole URL in `encodeURI()`. (Honestly in practice I never use `encodeURI()`, only `encodeURIComponent()`.)

client/src/routes.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const addressPageRouteWithParams = createRouteForAddressPage(
+  ":boro",
+  ":streetname",
+  ":housenumber"
+);
```
toolness, line 19: Err, using `encodeURIComponent()` will break this though, as it'd escape the `:`; you'd want to just replace this with the raw string `/address/:boro/:streetname/:housenumber` rather than calling out to `createRouteForAddressPage()`.

Honestly though, IMO the street name and house number (possibly also the boro) should never have been path components, but rather querystring arguments.  Path components should only ever contain strings that don't need any escaping, like slugs, numbers, etc.  But I guess we are stuck with what we've got since we don't want to break existing links...

client/src/containers/NychaPage.js
```diff
@@ -313,12 +314,11 @@ class NychaPageWithoutI18n extends Component {
...
-                        usersInputAddress.boro +
-                        "/" +
-                        usersInputAddress.housenumber +
-                        "/" +
-                        usersInputAddress.streetname
+                      `https://whoownswhat.justfix.nyc${createRouteForAddressPage(
+                        usersInputAddress.boro,
+                        usersInputAddress.streetname,
+                        usersInputAddress.housenumber
+                      )}`
```
toolness, line 321: I'm noticing there are a bunch of places where you're essentially passing as arguments to `createRouteForAddressPage` an object that contains `boro`, `streetname`, and `housenumber` properties, but you have to split them up into sequential arguments.  I can think of a few ways to make the calls to this code a bit more concise:

* Change `createRouteForAddressPage` to take an object containing these properties, rather than sequential arguments.  This has the additional benefit of reducing the possibility for confusion of argument order: for example, if a developer mixed up the order and passed in `boro, housenumber, streetname` that would be bad, but it's much harder to make that mistake with named arguments.  The downside of this approach, of course, is that calls become a bit more verbose when you don't already have an object matching this schema on-hand.

* Overload `createRouteForAddressPage` to take _either_ sequential arguments _or_ an object with these properties.  This would require some [TypeScript overloading](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) which gets a bit messy, though, and ideally you'd probably want to add some unit tests just to make sure the function behaves as you expect.

* Create a separate `createRouteForAddressPageFromObject()` function that takes the object with these properties.

Of course, you can also do nothing and keep these calls as they are, which is also totally fine!

client/src/routes.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const createWhoOwnsWhatRoutePaths = (prefix?: string) => {
+  const pathPrefix = prefix || "";
+  return {
+    home: `${pathPrefix}/`,
+    addressPageOverview: `${pathPrefix}${addressPageRouteWithParams}`,
+    addressPageTimeline: `${pathPrefix}${addressPageRouteWithParams}/timeline`,
+    addressPagePortfolio: `${pathPrefix}${addressPageRouteWithParams}/portfolio`,
+    addressPageSummary: `${pathPrefix}${addressPageRouteWithParams}/summary`,
+    bbl: `${pathPrefix}/bbl/:boro/:block/:lot`,
+    bblWithFullBblInUrl: `${pathPrefix}/bbl/:bbl`,
```
sraby, line 30: I'm gonna consider adding comments here to specify what these routes are.

client/src/containers/BBLPage.tsx
```diff
@@ -159,11 +159,11 @@ export default class BBLPage extends Component<BBLPageProps, State> {
...
-            pathname: createRouteForAddressPage(
-              addressForURL.boro,
-              addressForURL.streetname,
-              addressForURL.housenumber
-            ),
+            pathname: createRouteForAddressPage({
+              boro: addressForURL.boro,
+              streetname: addressForURL.streetname,
+              housenumber: addressForURL.housenumber,
+            }),
```
toolness, line 166: err is there a reason you're not just passing `addressForURL` here, e.g. `createRouteForAddressPage(addressForURL)`?  That would make it a lot more concise...

client/src/containers/HomePage.tsx
```diff
@@ -136,11 +148,11 @@ class HomePage extends Component<HomePageProps, State> {
...
-            pathname: createRouteForAddressPage(
-              this.state.searchAddress.boro,
-              this.state.searchAddress.streetname,
-              this.state.searchAddress.housenumber
-            ),
+            pathname: createRouteForAddressPage({
+              boro: this.state.searchAddress.boro,
+              streetname: this.state.searchAddress.streetname,
+              housenumber: this.state.searchAddress.housenumber,
+            }),
```
toolness, line 155: Similarly, this would be a lot easier to read as `createRouteForAddressPage(this.state.searchAddress)`!

client/src/containers/NotRegisteredPage.js
```diff
@@ -292,14 +292,12 @@ export default class NotRegisteredPage extends Component {
...
-                        usersInputAddress.streetname,
-                        usersInputAddress.housenumber
-                      )}`
-                    ).replace(" ", "%20")
-                  } // Support for Android
+                    `${getSiteOrigin()}${createRouteForAddressPage({
+                      boro: usersInputAddress.boro,
+                      streetname: usersInputAddress.streetname,
+                      housenumber: usersInputAddress.housenumber,
+                    })}`
```
toolness, line 299: Similarly, this would be a lot easier to read as `createRouteForAddressPage(usersInputAddress)`!

client/src/containers/NychaPage.js
```diff
@@ -313,14 +313,12 @@ class NychaPageWithoutI18n extends Component {
...
-                        usersInputAddress.streetname,
-                        usersInputAddress.housenumber
-                      )}`
-                    ).replace(" ", "%20")
-                  } // Support for Android
+                    `${getSiteOrigin()}${createRouteForAddressPage({
+                      boro: usersInputAddress.boro,
+                      streetname: usersInputAddress.streetname,
+                      housenumber: usersInputAddress.housenumber,
+                    })}`
```
toolness, line 320: Similarly, this would be a lot easier to read as `createRouteForAddressPage(usersInputAddress)`!

client/src/components/SocialShare.tsx
sraby: All set! I went ahead and factored out all `encodeURI()` instances while we're at it

client/src/routes.tsx
sraby: Sure thing

client/src/containers/NychaPage.js
sraby: I ended up going with option one, which is a lot cleaner :) thanks

PR #301, opened 2020-07-15
Consolidated formatDate and formatPrice functions 

Sjones21:
I merged all formatDate functions into one, with an "options" object passed in to specify the length of year, month, and day.


sraby:
@Sjones21 taking a look at this now! Also just so you know, we fixed some issues with old packages so I'm updating this branch from `master`.

client/src/components/DetailView.js
```diff
@@ -150,19 +143,47 @@ class DetailViewWithoutI18n extends Component {
...
                             <Trans>Last registered:</Trans>
                           </b>{" "}
-                          {this.formatDate(this.props.addr.lastregistrationdate)}
+                          {Helpers.formatDate(
+                            this.props.addr.lastregistrationdate,
+                            {
+                              year: "numeric",
+                              month: "short",
+                              day: "numeric",
+                            },
```
sraby, line 152: This object of date options seems to be used a bunch within the DetailView component. One suggestion for making this more DRY ("Don't Repeat Yourself") would be to actually save this object as a variable within the file, much like you do with `locale`. That way, you can just call it each time that you need it without spelling everything out again.

client/src/components/IndicatorsViz.js
```diff
@@ -305,7 +305,7 @@ class IndicatorsVizImplementation extends Component {
             } else if (timeSpan === "month") {
               // Make date value include day:
               var fullDate = this._data.labels[tooltipItem[0].index].concat("-15");
-              return Helpers.formatDateForTimeline(fullDate, locale);
+              return Helpers.formatDate(fullDate, { year: "numeric", month: "long" }, locale);
```
sraby, line 308: Same as above— it might make sense to save the options object as a variable somewhere, as you use it twice. If you choose to do so, you might want to name it something specific as we do seem to use other date options within this component, like `{ month: "short" }`, and would want to be clear about that distinction.

client/src/util/helpers.ts
```diff
@@ -109,6 +109,11 @@ export default {
     return null;
   },
 
+  formatPrice(locale: string, amount: number) {
+    const formatPrice = new Intl.NumberFormat(locale);
+    return formatPrice.format(amount);
+  },
+
```
sraby, line 116: Two things here:
1. It might be good to make locale an optional variable (by adding a ? at the end of the name) and including the default locale string within the body of the function, in the same way that `formatDate` works.
2. It would be good to specify the locale variable as a `SupportedLocale` here, as that will allow us to type check and make sure we are inputting a locale that we actually have enabled on the site, and not just any random string. Again, `formatDate` is a good example of that.

sraby:
Really great job on this @Sjones21! Appreciate your thoughtful work on this. I have some small suggestions above so feel free to ask me to clarify anything that's not clear.

Other general thoughts:

1.  We have a lot of different date option objects that we feed into the `formatDate` function— some of which we use in multiple places. For example:
```
{
   year: "numeric",
   month: "short",
   day: "numeric",
}
```
is an input we use both on the `PropertiesList` and `DetailView` components. For cases like these, it may be cleaner just to save the object as a variable— either In helpers.ts or in one of the files it shows up in, and export that variable so it can be referenced around the codebase. Up to you, though!

2. In the future, try to include commit messages that are a bit more detailed if you can. "Update helpers.ts" will not be super helpful to you if you want to go back and look at specific changes you made.

client/src/util/helpers.ts
```diff
@@ -109,6 +115,11 @@ export default {
     return null;
   },
 
+  formatPrice(amount:number, locale?: SupportedLocale): string {
+    const formatPrice = new Intl.NumberFormat(locale);
```
sraby, line 119: One quick thing: make sure you account for the case where locale is not defined!

PR #305, opened 2020-07-17
Remove babel and bluebird from server.

toolness:
This simplifies our server by removing babel and bluebird, since it seems like Node 12 supports everything that we used from those packages.


toolness:
Hmm, I think I am just gonna close this in factor of rewriting the whole doggone thing in Django (#306) but maybe we'll come back to this if that doesn't work out.

PR #306, opened 2020-07-17
[WIP EXPERIMENT] Rewrite server in Django

### LINKED ISSUE #90: Remove landlord contact info and endpoint
The SQL and endpoint added in #40 can be removed because it was only used by the tenants app, and as of https://github.com/JustFixNYC/tenants2/pull/380, that app is getting the information directly from NYCDB. So we should go ahead and remove the dead code from this repository.

### LINKED ISSUE #195: "Export Data" button seems to require GeoClient authentication in order to work
Discovered through investigation into #194

### LINKED ISSUE #289: Remove Deprecated Mailchimp server functions
These functions were rendered obsolete by #288. Let's remove them after we let our new implementation bake in the sun for a little bit

toolness:
This is an attempt to rewrite the server in Django, and have the front-end served entirely via CDN (e.g. [Netlify](https://create-react-app.dev/docs/deployment/#netlify) which supports client-side routing).

All the non-deprecated API endpoints are implemented.

Fixes #289.
Fixes #90.
Fixes #195.

## Rationale

* The current server wasn't written by us, so there's not really any institutional memory embedded in it.  It's also uses old versions of dependencies, some of which probably aren't actually needed (see #305), and has a bunch of dead code that needs to be removed, It also has absolutely no tests.

* Right now we only have one other server--the [tenant platform](https://github.com/justFixNYC/tenants2)--and it's in Django.  Even though Django is kind of overkill for an API server as simple as WoW's, using the same framework will make it much easier for developers to switch between projects.

* Doing this also makes it easy for us to eventually integrate the WoW server _into_ the tenant platform, which is something we've occasionally considered for various sundry reasons.

* Having two `package.json` files is [extremely confusing](https://github.com/JustFixNYC/who-owns-what/issues/260).  This is a very minor reason, of course, but it shouldn't be understated how utterly confusing it is.

* Hosting the front-end via CDN is actually independent of rewriting the server in Django; i.e., we could still keep the server in JS, but host the static files on a CDN and have the JS server used as an API backend only.  However, since we're migrating to Django in this PR, we might as well move the front-end to a CDN too, as it will relieve us of having to deal with static file serving from the Django app, will improve scalability, simplifies deployment to Heroku, and avoids the use of CRA's confusing [proxy](https://create-react-app.dev/docs/proxying-api-requests-in-development/) functionality.  ~~Perhaps most importantly, it will allow out-of-date clients to still fetch old versions of front-end files, thereby constituting a more robust fix for #169 (we still get lots of errors from Rollbar because this isn't currently supported)~~ (Update: Netlify doesn't actually support this, and deletes old files as soon as we deploy new ones!  Oof).

## Notes

* The CSV endpoint only takes a `bbl` argument, rather than street name/house number/borough.  This relieves the back-end of having to ever contact any kind of geosearch API (this is done entirely on the front-end).  It fixes #195 but also means that out-of-date clients will break.  My hope is that this functionality isn't used _too_ often and that people who encounter errors will just reload the page and everything will work.

* For similar reasons, this removes `/api/address` support for street name/house number/borough.  As far as I can tell, the front-end never actually uses this endpoint, but we should make sure.

* The CSV exports are slightly different.  Specifically ,dates are represented as e.g. `2019-08-20` rather than `"Tue Aug 20 2019 00:00:00 GMT+0000 (Coordinated Universal Time)"`.  The order of the colums _seems_ to be preserved but I'm not 100% sure.

## To do

Some of this can be punted to future PRs.

- [x] Resolve critical TODOs in `settings.py`.
- [x] Remove the `"proxy"` entry from `client/package.json`.
- [x] Remove all references to geoclient environment variables from `.env.sample`.
- [x] Change deployment so it uses a Python buildpack and all static content is hosted via Netlify.
- [ ] Consider adding a more stringent CORS policy.
- [ ] Add comprehensive tests.
- [x] Add Rollbar integration.
- [x] Remove the `/api/hello` endpoint.
- [x] Return HTTP 400 for all bad requests.
- [x] Ensure that 500 errors result in JSON responses for API endpoints.  The JS server returns an object with an `error` property but sets the status code to 200, which I disagree with.  IMO this should be changed to 500.
- [x] Remove the `server` directory, root `package.json`, and any other remnants of the JS server.
- [ ] Remove code to access `/api/address` with street name/house number/borough from front-end.
- [x] Change the `Procfile` to run `gunicorn` instead of node.
- [ ] Add a linting step that checks for unused imports (possibly PEP-8 compliance too).
- [x] Update the README to reflect this new reality.
- [ ] Remember to change `HEROKU_API_KEY` on CircleCI to use a long-lived token created via `heroku authorizations:create` (right now it's using a short-lived one created via `heroku auth:token`).


server/controllers/address.js
```diff
@@ -1,110 +0,0 @@
...
-      address: {
-        geosupportReturnCode: "00",
-        bbl: query.borough + query.block + query.lot,
-      },
-    });
-    // return geo.requestBBL(query).then(formatData);
-  } else {
-    throw new Error("API query param mismatch");
-  }
-};
```
sraby, line 38: Hey @toolness! Reviewing this again, and yes it absolutely does not make sense for us to re-fetch the bbl from a housenumber, street, and borough combination. A user has access to an "Export Data" button only if they have located their address's bbl and are on the Address Page component.

sraby: This also makes me wonder if we even need to request anything again from the database— given that this function is only available on the Address Page component (for addresses that have a record in the `wow_bldgs` table), we should already have all of the exportable data stored in state when a user clicks on "Export Data" button. Therefore, we may only need to reformat and restructure the data in state at that time which wouldn't require a database call at all. Let's discuss today at our check-in!

PR #307, opened 2020-07-19
Fix bug on indicator vis tooltip for "quarter" view

sraby:
This PR fixes a bug where the abbreviation of months on the timeline tab would show up as invalid values— specifically when viewing the data by quarters on Safari or Firefox. 

image


sraby:
@Sjones21, just FYI I fixed this bug on the timeline tab— which may have implications for the work you are doing in #301. Specifically, my fix here creates a new helper function that references the `formatMonthAbbreviationForTimeline` function within it. Since it seems that your work is actually getting rid of the `formatMonthAbbreviationForTimeline` in lieu of a clearer, more usable function definition, there will likely be a merge conflict we need to figure out. Let's find a time to work through that together!

PR #331, opened 2020-07-23
Convert most remaining files to typescript

sraby:
This converts most of the remaining files to TS, except AddressPage and a few others.

### DIFF
--- a/client/src/containers/NotRegisteredPage.tsx
+++ b/client/src/containers/NotRegisteredPage.tsx
@@ -1,17 +1,29 @@
 import React, { Component } from "react";
 import { LocaleLink as Link } from "../i18n";
-import Modal from "components/Modal";
-import LegalFooter from "components/LegalFooter";
-import Helpers from "util/helpers";
-import APIClient from "components/APIClient";
-import SocialShare from "components/SocialShare";
 
 import "styles/NotRegisteredPage.css";
 import { Trans } from "@lingui/macro";
 import { createRouteForAddressPage, getSiteOrigin } from "../routes";
+import Modal from "../components/Modal";
+import LegalFooter from "../components/LegalFooter";
+import Helpers from "../util/helpers";
+import APIClient from "../components/APIClient";
+import SocialShare from "../components/SocialShare";
+import { SearchAddress } from "../components/AddressSearch";
+import { GeoSearchData, BuildingInfoRecord } from "../components/APIDataTypes";
 
-export default class NotRegisteredPage extends Component {
-  constructor(props) {
+type Props = {
+  geosearch?: GeoSearchData;
+  searchAddress?: SearchAddress;
+};
+
+type State = {
+  showModal: boolean;
+  buildingInfo: BuildingInfoRecord[] | null;
+};
+
+export default class NotRegisteredPage extends Component<Props, State> {
+  constructor(props: Props) {
     super(props);
 
     this.state = {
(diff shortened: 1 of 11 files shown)


client/src/containers/NotRegisteredPage.tsx
```diff
@@ -21,7 +36,7 @@ export default class NotRegisteredPage extends Component {
   }
 
   componentDidMount() {
-    if (this.props.geosearch && this.props.geosearch.bbl && !this.state.buildingInfo) {
+    if (this.props.geosearch && !this.state.buildingInfo) {
```
toolness, line 39: Huh the removal of the `this.props.geosearch.bbl` in this conditional is interesting... I wonder if it might affect #340?  Not sure but the error in that issue is coming from very near this line of code.

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -289,10 +301,11 @@ export default class NotRegisteredPage extends Component {
                   <Trans>Share this page with your neighbors</Trans>
                 </p>
                 <SocialShare
-                  location="nycha-page"
+                  location="not-registered-page"
```
toolness, line 304: Interesting, is there a reason this string is changing from `nycha-page` to `not-registered-page`?

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -39,20 +54,19 @@ export default class NotRegisteredPage extends Component {
...
+          housenumber: searchAddress.housenumber || " ",
+          streetname: searchAddress.streetname || " ",
+        }
+      : buildingInfo
+      ? {
+          boro: buildingInfo.boro,
+          housenumber: buildingInfo.housenumber || " ",
+          streetname: buildingInfo.streetname || " ",
+        }
+      : null;
```
toolness, line 69: I know this PR is just for typescript-ification but it seems like we could just factor out this object notation into a helper function to make things more DRY.  If I'm correct maybe we should make an issue for it or something?  Not really a big deal either way I guess.

client/src/containers/NychaPage.tsx
```diff
@@ -312,8 +330,9 @@ class NychaPageWithoutI18n extends Component {
                 <SocialShare
                   location="nycha-page"
                   url={
-                    usersInputAddress &&
-                    `${getSiteOrigin()}${createRouteForAddressPage(usersInputAddress)}`
+                    usersInputAddress
+                      ? `${getSiteOrigin()}${createRouteForAddressPage(usersInputAddress)}`
+                      : undefined
```
toolness, line 335: Also out-of-scope of this PR, but it seems like both the generated URL and the ternary here are the same as in the not registered page, which makes me wonder if it'd be worth refactoring something to be more DRY... Not a huge deal though.

client/src/containers/NotRegisteredPage.tsx
toolness: Actually I wonder if it makes sense to, inside this conditional, do a `if (!this.props.geosearch.bbl) rollbar.error("Geosearch result has no BBL!");` just to make sure that we're notified if our assumptions about our data types are incorrect?  I guess we could also move this to the `ApiClient.ts` file or whevever we're originally getting that `geosearch` object too, though.

sraby: Yes, good call!

client/src/containers/NotRegisteredPage.tsx
sraby: Yes, this is used for utm reporting in our social share— this should appropriately be not-registered to be a better label for this page.

client/src/containers/NotRegisteredPage.tsx
sraby: Ahhh so actually there is an issue here—  `searchAddress.streetname` is typed to never be null so there isn't exact parity between the construction of the first object and the construction of the second one here.

sraby: I just updated the code to reflect that.

client/src/containers/NychaPage.tsx
sraby: Sweet! Will address this in a resulting PR

PR #332, opened 2020-07-23
Factor out APIClient.searchForAddressWithGeosearch().

### LINKED ISSUE #317: Remove code to access /api/address with street name/house number/borough from front-end
In #306 we dropped support for it, but there's actually a code path that attempts to use it.  We should remove it from there (we might have to look up the BBL based on the address and then make the network request for that, I guess).

### LINKED ISSUE #329: Clean up use of GeoRequester in AddressPage.js
Sam was trying to convert this file to TypeScript and realized we're not using the class' public interface properly.  We need to refactor how we're using it!

toolness:
This fixes #329 by factoring out a new function called `APIClient.searchForAddressWithGeosearch()` which first uses the GeoSearch API (properly) to get the BBL for an address, and then passes the BBL on to the WoW API to get portfolio information.

It also modifies `APIClient.searchForAddress()` to use this function if it's not given a BBL, which fixes #317.

### DIFF
--- a/client/src/components/APIClient.ts
+++ b/client/src/components/APIClient.ts
@@ -6,18 +6,48 @@ import {
   WithBoroBlockLot,
 } from "./APIDataTypes";
 import { SearchAddress } from "./AddressSearch";
+import { GeoSearchRequester } from "@justfixnyc/geosearch-requester";
 
 // API REQUESTS TO THE DATABASE:
 
+function searchForAddressWithGeosearch(q: {
+  housenumber?: string;
+  streetname: string;
+  boro: string;
+}): Promise<SearchResults> {
+  let addr = `${q.streetname}, ${q.boro}`;
+  if (q.housenumber) {
+    addr = `${q.housenumber} ${addr}`;
+  }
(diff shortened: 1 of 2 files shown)


client/src/components/APIClient.ts
```diff
@@ -6,18 +6,48 @@ import {
...
+    req.changeSearchRequest(addr);
+  });
+}
+
+function splitBBL(bbl: string): { boro: string; block: string; lot: string } {
+  return {
+    boro: bbl.slice(0, 1),
+    block: bbl.slice(1, 6),
+    lot: bbl.slice(6, 10),
+  };
```
toolness, line 43: Hmm, I just noticed there's a function in `helpers.ts` called `splitBBL()` which is _extremely_ similar to this, but differs in some weird ways.  I wonder if we should coalesce the two?

sraby: So strange! Gonna boot this to a separate issue.

toolness:
Cool thanks sir!

PR #334, opened 2020-07-24
Allow for age to be null in aggregate endpoint

toolness:
While WoW never seems to ping the `/api/aggregate` endpoint for BBLs that have no HPD registration, DAP does (albeit at the synonymous `/api/dap-aggregate` endpoint), and it sometimes asks about buildings that apparently either have no PLUTO data, or for which PLUTO reports the building's age as being `NULL`.  Regardless, the value our NYCDB returns is `NULL`, which our typecasting code introduced in #326 barfs on, so this fixes it.

The offending BBLs are very small buildings so I'm not comfortable posting the examples here, but they can be seen on Rollbar at https://rollbar.com/justfix-nyc/who-owns-what/items/1957/.

### DIFF
--- a/client/src/components/APIDataTypes.ts
+++ b/client/src/components/APIDataTypes.ts
@@ -75,7 +75,7 @@ type HpdViolationsAddress = AddressLocation & {
 };
 
 type SummaryStatsRecord = {
-  age: number;
+  age: number | null;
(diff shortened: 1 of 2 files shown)


sraby:
Looks great, thanks @toolness!

Issue #340, opened 2020-07-27
"TypeError: Cannot get property 'bbl' of undefined or null" from AddressPage.js's handleResults()


toolness:
Not sure what's going on here but it might be linked to #318, possibly.

toolness:
Note that since #346 this will be reported as an explicit exception with the message `Address results do not contain geosearch results`.

PR #346, opened 2020-07-27
Convert AddressPage.js to TSX.

toolness:
This converts `AddressPage.js` to TSX.  ~~Crucially, though, most of the address-related state fields are typed to `any`, which makes some of this incomplete.  Those fields are going to take a _lot_ of work to properly type because of how they're being used, and how they interact with some components further down in the component hierarchy (some of which are still untyped). It might ultimately be easier to just refactor the data model to be more comprehensible, rather than attempting to add types as-is.~~

Note that this introduces an explicit `throw` for the situation where an address search has no geosearch information; this essentially means that the error signature for #340 will now be different, as it will be an explicit exception rather than `Cannot get property 'bbl' of undefined or null`.


client/src/components/PropertiesList.tsx
```diff
@@ -10,35 +10,17 @@ import { t } from "@lingui/macro";
...
-  rsunits2007: number;
-  rsunits2017: number;
-  openviolations: number;
-  totalviolations: number;
-  evictions: string | null;
-  ownernames: { title: string; value: string }[];
-  lastsaledate: string;
-  lastsaleamount: string;
-  lastsaleacrisid: string;
-};
```
toolness, line 31: Well, at least we're getting rid of this duplicate typing.

client/src/containers/AddressPage.tsx
```diff
@@ -1,35 +1,64 @@
...
+
+export default class AddressPage extends Component<AddressPageProps, State> {
+  constructor(props: AddressPageProps) {
     super(props);
 
     this.state = {
       searchAddress: { ...props.match.params }, // maybe this should be
       userAddr: {}, // merged together?
       hasSearched: false,
-      geosearch: {},
```
toolness, line 32: I'm a bit queasy about this change but hopefully it should work.  We should probably kick the tires a bit just to make sure nothing explodes...

sraby:
@toolness this is ready for your eyes again!

toolness:
awesome thanks @sraby !

PR #347, opened 2020-07-27
Added Typescript to remaining files (except Address Page and Timeline Pages)

sraby:
This PR adds typescript all remaining pages in the WOW front-end (with the exception of the Address Page and Timeline Pages which will be saved for separate PRs).

### DIFF
--- a/client/src/components/APIDataTypes.ts
+++ b/client/src/components/APIDataTypes.ts
@@ -34,6 +34,8 @@ export type AddressRecord = {
   lastsaledate: string | null;
   lat: number | null;
   lng: number | null;
+  /** This property gets assigned in the PropertiesMap component, not from our API */
+  mapType?: string;
   openviolations: number;
   ownernames: HpdOwnerContact[] | null;
   registrationenddate: string;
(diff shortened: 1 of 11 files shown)


client/src/components/APIDataTypes.ts
```diff
@@ -20,23 +20,25 @@ export type GeoSearchData = {
...
+export type AddressRecord = {
   bbl: string;
   bin: string;
   boro: Borough;
   businessaddrs: string[] | null;
   corpnames: string[] | null;
   evictions: number | null;
   housenumber: string;
-  lastregistrationdate: Date;
+  lastregistrationdate: string;
```
toolness, line 31: One thing that might help here for documentation is to create a type called e.g. `APIDate` that is just a string, but where you document the type as adhering to a particular format, e.g. YYYY-MM-DD.  For example, [tenants2 has this](https://github.com/JustFixNYC/tenants2/blob/f9247d9666973fac5c2a28d0b9e53eaefd58059a/frontend/lib/networking/custom-graphql-scalars.d.ts#L1-L5).

sraby: Nice! That's great.

PR #350, opened 2020-07-28
Improve ChartJS typings.

toolness:
This replaces the `any` typing for our ChartJS data to be a `ChartData` instead.


sraby:
Noice! looks awesome— thank you for this

PR #354, opened 2020-07-29
Fix #351.

### LINKED ISSUE #351: Timeline title is just "0" when a dataset has no entries
Here's an example: http://localhost:3000/en/address/BROOKLYN/90/WILSON%20AVENUE/timeline

Click "Building Permit Applications".

This was likely introduced by #349, as the bug doesn't exist before our recent TypeScript-ification.

toolness:
In the case where an indicator's total was `0`, this was causing a short-circuited expression to incorrectly short-circuit.  Now we explicitly test for a value to be `null` rather than simply falsy.


sraby:
Noice!

PR #356, opened 2020-07-29
Set baseUrl instead of changing NODE_PATH.

toolness:
This uses a [baseUrl](https://www.typescriptlang.org/docs/handbook/module-resolution.html#base-url) instead of changing `NODE_PATH`, which has been deprecated.


sraby:
Cool cool!

PR #363, opened 2020-07-29
Make request for Aggregate data before summary tab is visible

sraby:


sraby:
We don't need this anymore, thanks to #364 :)

PR #368, opened 2020-08-03
Add timeline, summary child states to "portfolioFound" state

toolness:
We're implementing child states so we might as well use a child pull request.

### DIFF
--- a/client/src/state-machine.ts
+++ b/client/src/state-machine.ts
@@ -18,7 +18,6 @@ import APIClient from "components/APIClient";
 import helpers, { assertNotUndefined } from "util/helpers";
 
 import _find from "lodash/find";
-import { useMachine } from "@xstate/react";
 
 type WowState =
   | { value: "noData"; context: {} }
--- a/client/src/state-machine.ts
+++ b/client/src/state-machine.ts
@@ -1,11 +1,4 @@
-import {
-  createMachine,
-  assign,
-  DoneInvokeEvent,
-  State,
-  TransitionsConfig,
-  interpret,
-} from "xstate";
+import { createMachine, assign, DoneInvokeEvent, State, TransitionsConfig } from "xstate";
 import {
   SearchAddressWithoutBbl,
   AddressRecord,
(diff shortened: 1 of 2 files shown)


client/src/state-machine.ts
```diff
@@ -43,17 +52,55 @@ type WowState =
...
+      context: WowPortfolioFoundContext;
+    }
+  | {
+      value: { portfolioFound: { summary: "pending" } };
+      context: WowPortfolioFoundContext;
+    }
+  | {
+      value: { portfolioFound: { summary: "error" } };
+      context: WowPortfolioFoundContext;
+    }
```
toolness, line 86: For some reason TypeScript doesn't let me write `summary: "noData"|"pending"|"error"` without introducing bizarre type errors elsewhere in the code, even though the docs actually recommend such a pattern.  Sigh.

client/src/state-machine.ts
```diff
@@ -74,12 +121,10 @@ type PortfolioData = {
 };
 
 type TimelineData = {
-  timelineBbl: string;
```
toolness, line 77: I'm removing this for now because keeping track of it is a bit of a hassle and it's easier to just set up our state machine so we always reset the timeline/summary states to `noData` whenever our detail address changes.

client/src/state-machine.ts
```diff
@@ -255,10 +296,77 @@ export const wowMachine = createMachine<WowContext, WowEvent, WowState>({
...
         ...handleSearchEvent,
+        VIEW_TIMELINE: {
+          target: [".timeline.pending"],
+        },
+        VIEW_SUMMARY: {
+          target: [".summary.pending"],
+        },
         SELECT_DETAIL_ADDR: {
-          target: "portfolioFound",
+          target: [".summary.noData", ".timeline.noData"],
```
toolness, line 369: Oof, I'm showing this to @sam now and he pointed out that we don't need to set `.timeline.noData` here b/c that information isn't actually specific to the detail address!!!

PR #374, opened 2020-08-05
Cleaned up our helper functions and added full test coverage

sraby:
This PR consolidates some duplicate helpers (like `splitBBL`), moves some function declarations around so that helpers.ts only includes globally-relevant helper functions, and adds full test coverage for all.


toolness:
Noice!

PR #377, opened 2020-08-05
Added more robust State machine tests

### LINKED ISSUE #262: Add robust front-end tests
Let's add some front end tests to WOW! High priority ones are:
- test all the different pathways for when an address is searched (to address page, nycha page, not registered page)
- test different conditionals on Summary tab

sraby:
This PR adds more tests to our XState Machine so that we now cover all state changes within WoW. For now, fixes #262.

### DIFF
--- a/client/src/state-machine.test.ts
+++ b/client/src/state-machine.test.ts
@@ -3,12 +3,18 @@ import { interpret } from "xstate";
 import { GEO_AUTOCOMPLETE_URL } from "@justfixnyc/geosearch-requester";
 import { waitUntilStateMatches, mockJsonResponse, mockResponses } from "tests/test-util";
 import GEOCODING_EXAMPLE_SEARCH from "./tests/geocoding-example-search.json";
-import { SearchResults, BuildingInfoResults, IndicatorsHistoryResults } from "components/APIDataTypes";
+import {
+  SearchResults,
+  BuildingInfoResults,
+  IndicatorsHistoryResults,
+  SummaryResults,
+} from "components/APIDataTypes";
 import helpers from "util/helpers";
 import {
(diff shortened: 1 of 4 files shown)


sraby:
@toolness this seemed like a good stopping point for now, but let me know if you think there are any other major state machine functions we should add tests for.

client/src/state-machine.test.ts
```diff
@@ -59,31 +94,104 @@ describe("wowMachine", () => {
...
+
+    const wm = interpret(wowMachine).start({ portfolioFound: { summary: "noData" } });
+    wm.state.context = PORTFOLIO_FOUND_CTX;
+
+    wm.send({ type: "VIEW_SUMMARY" });
+    await waitUntilStateMatches(wm, { portfolioFound: { summary: "success" } });
+  });
+
+  it("should deal w/ summary data request errors", async () => {
+    mockResponses({ [PORTFOLIO_URLS.SUMMARY_URL]: { status: 500 } });
```
sraby, line 189: Mocking a 500 status here actually triggers a console.log of the error in APIClient.ts... so while these tests pass they do clutter our testing feed with error messages like
```
HTTPError: HTTP Error Internal Server Error
        at checkStatus (/Users/sam/Sites/who-owns-what/client/src/components/APIClient.ts:149:17) {...
```

Is there any way to disable those for when we are running the tests?

toolness: Hmm good question, honestly I think we should just remove the console.log in this case--it's not particularly helpful, at least with 500's, since the server will be logging it on its end, and with a lot more information about what caused the error?

sraby: Yeah definitely. Ok good!

Issue #381, opened 2020-08-06
Add error-ish edge case URLs to /dev/ page.


toolness:
We should add an example error BBL route that e.g. goes to `/bbl/NA`, and an example error address page that goes to e.g. `/address/BROOKLYN/999/BOOP ROAD`.

toolness:
Also BBLs not in PLUTO! (See #385.)

PR #385, opened 2020-08-07
Deal semi-gracefully with BBLs not in PLUTO.

toolness:
Apparently there are some addresses that map to actual BBLs in GeoSearch and represent real buildings, but which PLUTO doesn't have entries for.

An example of one (at the time of this writing) is 77 Park Place in Manhattan, which [exists according to StreetEasy](https://streeteasy.com/building/the-griffon) and has BBL 1008940201 according to NYC GeoSearch, but which apparently PLUTO 19v2 has no information about.  Because of this, the API request for building info returns an empty array, which the `NotRegisteredPage` component chokes on (currently with an exception of `Cannot get property 'charAt' of undefined or null`).

(Interestingly [ZoLa chokes on this address as well](https://zola.planning.nyc.gov/l/lot/1/894/201).)

I'm not sure what the best fix for this is, but this PR basically makes WoW act as though it can't find the BBL for the address, so it will respond with a page that says "No address found", which seems at least somewhat honest since we can't find the address in PLUTO...

### DIFF
--- a/client/src/state-machine.test.ts
+++ b/client/src/state-machine.test.ts
@@ -112,6 +112,26 @@ describe("wowMachine", () => {
     await waitUntilStateMatches(wm, "unregisteredFound");
   });
 
+  it("should deal w/ unregistered addresses not in PLUTO", async () => {
+    mockResponses({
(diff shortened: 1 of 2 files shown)


toolness:
Thanks! And good idea, I'll add a case to the `/dev` page when I tackle #381!

Issue #361, opened 2020-07-29
Coalesce our `generateBaseUrl` function and `createRouteForAddressPage` functions.


sraby:
We have an exported function in routes.tsx called `createRouteForAddressPage` that generates the proper route structure given the correct route params for an address page. We also have a function in the AddressPage component itself called `generateBaseUrl` that performs nearly the identical function and gets (annoyingly) passed down to child components. I'm wondering if there is a way we can eliminate the latter function using the helper instead.

toolness:
Fixed by #390!

PR #390, opened 2020-08-10
Factor out createAddressPageRoutes() and add a 404 page.

### LINKED ISSUE #389: WoW doesn't have a 404 page
I just realized that WoW doesn't have a 404 page, and just shows a blank page with the WoW banner at the top if a non-existent route is visited, e.g.:

image

It might be good to actually put something in there, like the tenant platform's "Alas.
Sorry, the page you are looking for doesn't seem to exist.".  Or at least something that might not make the user wonder if the page is still loading or broken or something.

toolness:
This factors out a new function called `createAddressPageRoutes()` which cleans up the creation of links to sub-pages of the address page, making them more type-safe and DRY.

It also moves the `WhoOwnsWhatRoutes` component into `App.tsx` to avoid circular imports.  This is also more ideal: `routes.tsx` should be only concerned with routing pathnames, rather than what routes map to what components, making it as dependency-free as possible so that anything that imports it doesn't also end up importing every module in the codebase (this will come in handy with Jest's detection of dependent tests, as well as if/when we decide to do code splitting).

Unrelatedly, it also removes the double quotes around the word "owns" in the Twitter sharing message, because that was really weird.  (This does mean we'll have to re-translate though, so I could always move it to a different PR...)

Finally, it adds a very spartan 404 page, which fixes #389.  The text on the page is taken from the tenant platform so it's easy to localize.

### DIFF
--- a/client/src/components/DetailView.tsx
+++ b/client/src/components/DetailView.tsx
@@ -14,7 +14,7 @@ import { isPartOfGroupSale } from "./PropertiesList";
 import { Link } from "react-router-dom";
 import { LocaleLink } from "../i18n";
 import BuildingStatsTable from "./BuildingStatsTable";
-import { createWhoOwnsWhatRoutePaths } from "../routes";
+import { createWhoOwnsWhatRoutePaths, AddressPageRoutes } from "../routes";
 import { AddressRecord } from "./APIDataTypes";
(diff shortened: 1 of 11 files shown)


toolness:
Ooo yeah I didn't even know about #361! Indeed, this fixes it!

PR #392, opened 2020-08-10
Updated copy for moratorium banner and reran lingui extract

sraby:
This PR updates the text on our Eviction Moratorium banner on the homepage. It specifically removes a mention of any date so that the messaging is more "evergreen" and doesn't need to be updated every time the Moratorium is extended.


sraby:
Since this is a small text change, I'm gonna go ahead and merge myself.

PR #393, opened 2020-08-10
Add Feature Callout widget to the app header menu

Sjones21:
This PR builds on the work of @Sjones21 in that it implements a "Feature Callout" widget that users can access via an icon from the header nav bar. When clicked, a dropdown menu will appear that walks users through a series of GIFs and descriptions that highlight new changes to the tool. 

TODO:
- [x] Open the widget by default 50% of the time

Screen Shot 2021-08-09 at 11 27 06 AM
Screen Shot 2021-08-09 at 11 27 22 AM

In addition to Sydney's work, @sraby added a feature flag ENV variable that allows us to enable/disable the presence of this widget without writing new code. 

_Sydney's previous comments from 7/23/2020:_ 

> This is the most recent conception of the Feature Callout widget.
> 
> I modified the files from FeatureCalloutWidgetFinal branch to be compatible with TSX files and merged all conflicts.
> 
> -I used the existing media schema "for-phone-only" to switch to a full-screen modal view, however the "for-phone-only" width seems fairly wide, so let me know what you think.
> 
> -Feel free to comment on any bugs. I am especially interested in making the accessibility schema conform to what is already present on WOW. Right now on WOW, screenreaders can only tab-through links, but I added "tabIndex = 0" to the buttons and header on my widget to allow keyboard access.


old_node_modules/.yarn-integrity
```diff
@@ -0,0 +1,20 @@
+{
```
toolness, line 1: Oops, it looks like this is in a folder called `old_node_modules`--did you commit that accidentally?

client/src/containers/App.tsx
```diff
@@ -42,13 +48,169 @@ export default class App extends Component<Props, State> {
...
+          description:
+            "It's now possible to share links to specific tabs (Overview, Timeline, Portfolio, & Summary).",
+          img: URLS_gif,
+        },
+        {
+          index: "4 of 4",
+          title: "Last Sold",
+          description: "The Overview and Portfolio tabs now display the date and price from when your building was last sold.",
+          img: LastSold_gif,
+        },
```
toolness, line 78: So I assume that ideally this information should be pulled from Contentful, right?  Maybe with the exception of `index`, which we can generate on-the-fly?

client/src/containers/App.tsx
```diff
@@ -42,13 +48,169 @@ export default class App extends Component<Props, State> {
...
+    let infoButton = document.querySelector(".widget-button-info");
+    if (widget.style.display === "none") {
+      widget.style.display = "inline-block";
+      if (tooltipTriangle) tooltipTriangle.classList.add("toggled");
+      if (infoButton) infoButton.classList.add("pressed");
+    } else {
+      widget.style.display = "none";
+      if (tooltipTriangle) tooltipTriangle.classList.remove("toggled");
+      if (infoButton) infoButton.classList.remove("pressed");
+    }
```
toolness, line 106: Hmm, so this code will be the easiest to understand and modify if you do things 100% using DOM APIs or 100% React, rather than mixing the two.  Right now this seems to be mixing the two, in that it's creating the widget using React but then modifying it using DOM APIs, which will end up confusing React because it's [virtual DOM](https://reactjs.org/docs/faq-internals.html) now deviates from the real DOM.  Instead of that, you'll likely want to do things 100% the React way by modifying the component's state when something is clicked, and then changing the `render()` method to apply the appropriate classes and styling depending on the state.  The [classnames](https://www.npmjs.com/package/classnames) package can be particularly useful in generating the final list of classes, I'm not sure if we use it yet in this project or not.

(I'm realizing now that we talked about doing parts of the first version of the widget using DOM APIs because of how hard it is to "point at" elements in completely different parts of a react app, but now that we're not doing the "pointing at" thing and just embedding animated GIFs, I don't think that mixing the two styles is very helpful... sorry if that's confusing, happy to talk it over more!)

client/src/containers/App.tsx
```diff
@@ -75,6 +237,7 @@ export default class App extends Component<Props, State> {
                   </span>
                 )}
                 <nav className="inline">
+                  {featureCalloutWidget}
```
toolness, line 240: So you might want to consider actually moving all the feature callout widget logic out of `App.tsx` and into a separate component in a different file, e.g. `FeatureCalloutWidget.tsx`.  Then this line would become `<FeatureCalloutWidget />` instead of `{featureCalloutWidget}` and it'd be a lot easier to understand the widget's logic because it wouldn't be intermingled in all the general WoW app logic.  It would also make the feature callout widget a lot easier to reuse in our other products!

old_node_modules/.yarn-integrity
Sjones21: removed!

client/src/containers/App.tsx
Sjones21: I attempted this, please lmk if it actually is React-only now

client/src/containers/App.tsx
Sjones21: I did this in a recent commit!

sraby:
[ch7651]

client/public/index.html
```diff
@@ -6,6 +6,10 @@
     <meta name="theme-color" content="#000000">
     <meta name="version" content="%REACT_APP_VERSION%">
 
+    <!--
+      Google Material UI icons
+    -->
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
toolness, line 12: It looks like this points to a CSS file that has a WOFF font in it that's a 100K download, are we comfortable with adding that to the site's page weight?  How many icons are we actually using?  If it's only like, one icon, we might want to just get its SVG and use that...

client/src/components/FeatureCalloutWidget.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const [entryIndex, setEntryIndex] = useState(0);
+
+  const { i18n } = props;
+
+  const content = getFeatureCalloutContent();
+  const numberOfEntries = content.length;
+
+  let goToNextEntry = () => {
+    entryIndex === numberOfEntries - 1 ? setEntryIndex(0) : setEntryIndex(entryIndex + 1);
```
toolness, line 22: It might be a bit less verbose if you put `setEntryIndex` on the outside of this, e.g.:

```js
setEntryIndex(entryIndex === numberOfEntries - 1 ? 0 : entryIndex + 1)
```

Although for that matter, an even more idiomatic way of doing this is to use the modulus operator, e.g.

```js
setEntryIndex((entryIndex + 1) % numberOfEntries)
```

You can apply similar logic to `goToPrevEntry` below.

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
```
toolness, line 3: Er, why is a `package-lock.json` being added to the repo?  Should this be removed, since we use yarn instead of npm?

toolness (approved):
Hey this looks great overall!  Just a few suggestions/notes below.

toolness:
Er also, some of what I commented on in my review might not actually be about code you wrote @sraby!  Sorry about that.

client/src/components/FeatureCalloutWidget.tsx
sraby: math math math math math

package-lock.json
sraby: gah!

client/public/index.html
sraby: Removing it entirely!

sraby:
@samaratrilling I added some screenshots to the PR description! You can also check out the [LIVE DEMO LINK](https://deploy-preview-393--wow-django-dev.netlify.app/).

samaratrilling:
Thanks!

PR #397, opened 2020-08-17
Replaced COVID banner with component from monorepo

sraby:
This PR makes use of the new react-common package out of our @JustFixNYC monorepo on npm- we now pull the content for the COVID Eviction Moratorium banner from there.


sraby:
Since this is a quick change, I'm gonna go ahead and merge myself

PR #402, opened 2020-08-19
Added spaces back into the Data Request URL

sraby:
This PR adds in some spaces into the "full address" parameter we include in the URL to our Data Request Google Form. It fixes an issue where addresses would show up on the form improperly formatted, like `921EAST 145TH STREET,BRONX`.


toolness:
umm i thought u were on vacation

PR #408, opened 2020-09-14
Clean up DetailView component 

### LINKED ISSUE #409: DetailView component doesn't reference our state machine props
Let's refactor this component like we did other AddressPage child components so we don't have to pass it the `addr` and `addrs` props down from its parent.

sraby:
This PR fixes #409 and #410— namely, it refactors the DetailView component like we did other AddressPage child components so we don't have to pass it the addr and addrs props down from its parent. It also introduces a new handler called `onOpenDetail` that makes sure to bring in the DetailView on mobile when a user mounts the address page for the first time.


sraby:
@toolness this PR is pretty straightforward maintenance stuff, so I'm gonna go ahead and merge it, but feel free to take a gander if you are interested!

PR #411, opened 2020-09-17
Updated footer copy

sraby:
Changed the footer of WOW to read:
"JustFix, Inc. is a registered 501(c)(3) nonprofit organization."
instead of "JustFix.nyc is a registered 501(c)(3) nonprofit organization."


sraby:
Quick fix, so I'm gonna go ahead and merge!

PR #412, opened 2020-10-05
Add new rentstab data to WOW sql queries

sraby:
This PR updates our SQL code for creating the `wow_bldgs` table and our aggregate function for summary, in order to include the new 2019 rent stabilized unit counts we now have access to. In doing so, this PR also removes the `percentchange` field from our wow bldgs table set up, which was never used and would introduce unneeded complexity to recreate with the new data.


wow/views.py
```diff
@@ -38,7 +38,6 @@ def clean_addr_dict(addr):
         "bin": str(addr['bin']),
         "lastsaleamount": int_or_none(addr['lastsaleamount']),
         "registrationid": str(addr['registrationid']),
-        "rspercentchange": float_or_none(addr['rspercentchange']),
```
toolness, line 41: Oh ok, it looks like the TypeScript typing changes to this, and the removal of `avrspercent`, are in #414.  Are we removing these b/c the frontend doesn't use them at all?

That's fine by me--I forget if this endpoint is the one used by DAP portal though, do you think it uses these fields?

toolness (approved):
Nice, this looks good to me!  Just one question in the comments.

wow/views.py
sraby: That is correct! And hmmm good point about DAP Portal— they don't use the wow table but they do use our aggregate function. I just pinged Jade to make sure they know that we are removing that field. They are def not using any rs datapoints on DAP Portal from us but you're right that we should flag it for them so they can prepare for it in case there is any validation in place for the API response they get from us.

toolness:
Oh, one quick note... while writing https://github.com/JustFixNYC/tenants2/pull/1690#discussion_r500562743 I realized that idea could be applied here too.  What if instead of having separate fields for 2017 and 2019 (and next year, 2020), we instead just had a `unitsStabLatest` and `unitsStabLatestYear`?  The former could be the count of the RS units, and the latter could be the count of the year that the `unitsStabLatest` count represents (i.e. 2019 for now, and 2020 when we update the data next year).  This would allow us to update this count in future years without needing to make any front-end changes (e.g. #414) at all!

sraby:
oh hmmmm @toolness ok let me give it a shot

toolness (approved):
Oh nice, this is great!

PR #413, opened 2020-10-06
Add 'extra_nycdb_test_data' section to YML config.

toolness:
It's super confusing to add new NYCDB dataset dependencies to WoW because of how our tests work, so this moves the mention of any CSVs that we need to directly copy over into a new `extra_nycdb_test_data` entry in the `who-owns-what.yml` file.  Now whenever we add a new nycdb dataset dependency, hopefully we can just specify any additional CSVs in this area so tests know how to scaffold the dataset for tests.


sraby (approved):
Looks great! So simple

PR #414, opened 2020-10-06
Update front-end to show rent stabilized unit counts for 2019

sraby:
This PR updates all of our front-end code to make use of the new 2019 rent stabilized unit counts we now have access to in nycdb. While updating our tooltips for the rent stab data box on the Detail View, this PR also fixes a minor bug where the Δ symbol in the tooltip wasn't rendering properly. 

NOTE: this PR is dependent on #412 being productionalized, as it requires the `rsunitslatest` and `rsunitslatestyear` fields to be present in our `wow_bldgs` table.


toolness (approved):
Nice!

toolness (approved):
NOICE

PR #417, opened 2020-10-28
Remove old unitsstab2017 field from wow_bldgs table

sraby:
This PR removes an unused and outdated field from our `wow_bldgs` sql table, and also removes the field from our type definitions within our state machine. It also updates our "exported test data" that our test suite uses, to reflect the changes to the columns of our `wow_bldgs` table.


sraby:
[ch5044]

toolness (approved):
Cool, this looks great!

PR #422, opened 2020-12-03
Add definition of `anyarray_uniq` function to helpers sql

sraby:
So, it turns out that this function `anyarray_uniq` is NOT a pre-loaded array function that we [get for free through the pg_catalog](https://www.postgresql.org/docs/9.4/functions-array.html). This morning, we started seeing this critical error: 
```
function anyarray_uniq(integer[]) does not exist LINE 8: unnest(anyarray_uniq(array_cat_agg(merged.uniqregids... ^ HINT: No function matches the given name and argument types...
```
and it turns out that `anyarray_uniq` was no longer defined. Given that this function lies within the definition of other vital functions that power key WOW interactions, like `get_assoc_addrs_from_bbl`, our search function.

Since `anyarray_uniq` does not get created through any script that powers our auto-updating instance of nycdb, I suspect that somehow this function got deleted during our rebuild and there was no code to recreate it.

So, this PR adds a function definition to our `helpers.sql` file to explicitly build `anyarray_uniq` whenever we rebuild the db.  It also reorders our code a bit to make sure every function declaration runs in the proper order.

### DIFF
--- a/sql/helper_functions.sql
+++ b/sql/helper_functions.sql
@@ -1,3 +1,65 @@
+-- https://github.com/JDBurnZ/postgresql-anyarray/blob/master/stable/anyarray_uniq.sql
+DROP FUNCTION IF EXISTS anyarray_uniq(anyarray);
+CREATE OR REPLACE FUNCTION anyarray_uniq(with_array anyarray)
+	RETURNS anyarray AS
+$BODY$
+	DECLARE
+		-- The variable used to track iteration over "with_array".
+		loop_offset integer;
+
+		-- The array to be returned by this function.
+		return_array with_array%TYPE := '{}';
+	BEGIN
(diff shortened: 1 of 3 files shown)


toolness:
Hmm, so I don't believe that adding this function in WoW is the answer... the explosion last night actually occurred because the auto-update of the `hpd_registrations` dataset failed, which left our nycdb instance in an inconsistent state.  See this issue for more details: https://github.com/JustFixNYC/nycdb-k8s-loader/issues/60

I'm going to close this PR because I feel pretty strongly that copying `hpd_registrations`'s functions over to WoW isn't the right solution to this problem--the problem is that some of `hpd_registrations`'s functions are being deleted by our auto-updating nycdb and not being restored, so we need to fix that. 😣

PR #423, opened 2020-12-08
Remove trailing slash from parseLocale function

sraby:
This PR fixes a bug where WOW would redirect paths without a trailing slash incorrectly. For example, any url ending in `/en` or `/en?utm....` would be interpreted as not having a locale defined, and then get redirected to `/en/en...` which was not a correct route. We reconfigured our `parseLocale...` code to recognize locales in the pathname even if there is no trailing slash present.


sraby:
@toolness we reviewed this together so gonna merge myself.

PR #428, opened 2021-02-22
Update content in our SocialShare component

sraby:
This PR adds new content to our SocialShare component— specifically the content that gets pre-filled when a user tries to share the tool via email or twitter from the homepage, Overview tab, Summary tab, or elsewhere.

### DIFF
--- a/client/src/components/SocialShare.tsx
+++ b/client/src/components/SocialShare.tsx
@@ -14,7 +14,7 @@ import { getSiteOrigin, createAddressPageRoutes } from "../routes";
 const DEFAULT_TWEET = t`Who’s responsible for issues in your apartment and building? Use #WhoOwnsWhat, a free tool to research property owners in NYC using public, open data. Built by @JustFixNYC, it functions on any device with an internet connection. Search your address here: `;
 const DEFAULT_EMAIL_SUBJECT = t`All the public info on your landlord`;
 const getDefaultEmailBody = (url: string) =>
-  t`Who Owns What is a free tool built by JustFix.nyc to research property owners in NYC. It has helped over 200,000 New Yorkers find out who really owns their buildings, what other buildings that their landlord or management company owns, and other critical information about code violations, evictions, rent stabilized units, and so much more in any given building. You can look up any residential building located in NYC, even public housing (NYCHA) buildings! Search your address here: ${url}`;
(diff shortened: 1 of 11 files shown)


client/src/react-social.d.ts
```diff
@@ -5,8 +5,15 @@ module "react-social" {
...
+    /**
+     * For EmailButton components, this prop becomes the BODY of the email itself,
+     * so it can be just a url or a larger string of text.
+     */
     url: string;
     windowOptions?: string[];
+    /**
+     * For EmailButton components, this prop becomes the SUBJECT of the email.
+     */
     message?: string;
```
sraby, line 17: @toolness so frustratingly, these `url` and `message` properties have different uses depending on what component is interacting with them. For tweets, `message` is the body of the tweet with `url` being a trailing url at the end of the tweet. However, for emails, `message` is actually the subject of the email when `url` corresponds to the body of the email, and therefore can be more than just a simple URL. 

I tried adding comments here and elsewhere to make this distinction more clear.

sraby:
[ch6213]

client/src/react-social.d.ts
toolness: Oh weird!  Thanks for clarifying this in the code!

client/src/components/SocialShare.tsx
```diff
@@ -102,12 +110,31 @@ const SocialSharePortfolioWithoutI18n: React.FC<{
...
       location={location}
       url={url}
-      twitterMessage={i18n._(
-        t`The ${buildingCount} buildings that my landlord owns 👀... #WhoOwnsWhat @JustFixNYC`
-      )}
-      emailMessage={i18n._(
-        t`The ${buildingCount} buildings owned by my landlord (via JustFix's Who Owns What tool)`
-      )}
+      twitterMessage={
+        location === "summary-tab"
```
toolness, line 114: I'm noticing there's a number of places where this code says `location === "summary-tab"`... it might make things a bit more concise and also less error-prone (e.g. if one of them is misspelled as "summary-tb" that would be bad) to define a boolean variable called `isSummaryTab` that represents this condition?  Eh, just a thought, but not a big deal.

toolness: Also, I feel like ideally the actual content of all this would just be passed in as a prop by the summary tab page itself (or defined by a context or something in that page), so that this component was generic and didn't need tons of branching logic--it'd also make more sense to find summary tab-specific content in the summary tab page, rather than here, I think... but anyways, I know a lot of this code is legacy and does this kind of thing all over the place already so it's not a priority.

toolness (approved):
Cool looks good!  Made a few minor suggestions but nothing major.

client/src/components/SocialShare.tsx
sraby: yes, both good points! Yeah I think it would be worth it to refactor

PR #435, opened 2021-03-03
Show complete count of evictions since 2017 across WOW

sraby:
This PR updates the WOW `create_bldgs_table` SQL function to make use of the new `marshal_evictions_all` table in NYCDB. This means that `create_bldgs_table` will now have cumulative and daily-updated evictions data going back to 2017. I also updated our front-end content to reflect this change.

NOTE: For NYCHA pages, we cache the eviction data by development for 2019 in a json stored in the client, so that count is remaining unchanged in this PR.


toolness (approved):
Whoa nice, good job getting the tests to work!!

sraby:
[ch6374]

PR #439, opened 2021-03-11
Close DetailView on mobile for every new address search

sraby:
This PR fixes an important but hard to reproduce bug where Mobile users couldn't use the Overview tab if they were to search ANOTHER address after their initial search. What was happening is that the "detailMobileSlide" state on the AddressPage component was pre-configured, so when the user returned to the AddressPage after searching a second address, it was not properly retriggering the opening of the DetailView slide and so users could only view the property map and not see any details. This PR fixes this issue by making sure the DetailView slide is "closed" every time a new address search is conducted.


sraby:
@toolness this is a quick bug fix (1 line of code) but take a look if you're interested!

PR #441, opened 2021-03-24
Stop using psycopg2-binary and use psycopg2 with a higher version.

samaratrilling:
Previously, who-owns-what tests were failing with the error "AttributeError: module 'psycopg2.errors' has no attribute 'UndefinedTable'".

They were failing because they used syntax that only works if you have psycopg2 v 2.8.6 installed, but that wasn't explicitly stated as a dependency. The computer the tests were written on 8 months ago actually had 2.8.6 installed instead of the expected 2.7.7. It had 2.8.6 installed because nycdb's requirements.txt asks for nycdb >= 2.7, and whenever WOW's requirements-dev.txt gets loaded, it fetches the latest version of psycopg2.

We thought we had pinned psycopg2 to version 2.7.7, but instead we had psycopg2-binary pinned to 2.7.7, which was ineffective at preventing arbitrary library updates from its dependencies (i.e. nycdb).

This makes it so we're actually pinning psycopg2 to 2.8.6.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # This only contains dependencies for running the API
 # server. For all other dependencies, see requirements-dev.txt.
-psycopg2-binary==2.7.7
+psycopg2==2.8.6
 django==3.1.6
 rollbar==0.14.5
 dj-database-url==0.5.0


samaratrilling:
Comments on my explanation are welcome!

toolness (approved):
This is great, and an excellent explanation too--thanks!! 🎉 

BTW, when you squash+merge, can you copy/paste your explanation into the commit description? (You may already know this since it's in the engineering handbook but I wanted to be redundant since undoing stuff in git is so complicated!)

samaratrilling:
I think it's in there! Let me know if you meant something else tho. https://github.com/JustFixNYC/who-owns-what/pull/441/commits/465b8eae85f81881ad048a8a64d9b522cff181ab

toolness:
> I think it's in there! Let me know if you meant something else tho. [465b8ea](https://github.com/JustFixNYC/who-owns-what/commit/465b8eae85f81881ad048a8a64d9b522cff181ab)

Yup you did it well, thanks!!

toolness:
Oh @samaratrilling last thing to do is just make sure you delete the branch once you're done merging it:

> image

Thankfully GitHub makes it easy to restore the branch if we ever need to, so deleting it once we're done is helpful so our list of active branches doesn't get out of control:

> image

Er... the above list should actually be shorter probably, we might not have been deleting branches as diligently as we ought to.

PR #443, opened 2021-03-30
Add new internationalized AirTable Data Request form to Summary tab

sraby:
Our new WOW Data Request form is built on AirTable and is available in [English](https://airtable.com/shruDUDHViainzUBB) and in [Spanish](https://airtable.com/shrQudDbhUVz5J83u)! This PR updates our outbound link on the Summary tab with the new form in its full i18n splendor.

### DIFF
--- a/client/src/components/PropertiesSummary.tsx
+++ b/client/src/components/PropertiesSummary.tsx
@@ -13,8 +13,9 @@ import { SocialShareAddressPage } from "./SocialShare";
 import { withMachineInStateProps } from "state-machine";
 import { NetworkErrorMessage } from "./NetworkErrorMessage";
 import { AddressRecord } from "./APIDataTypes";
-import { I18n } from "@lingui/react";
-import { defaultLocale, SupportedLocale } from "i18n-base";
+import { defaultLocale, isSupportedLocale } from "i18n-base";
+import { I18n } from "@lingui/core/i18n";
+import { I18n as I18nComponent } from "@lingui/react";
 
 type Props = withMachineInStateProps<"portfolioFound"> & {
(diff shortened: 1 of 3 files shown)


sraby:
[ch6502]

client/src/components/PropertiesSummary.tsx
```diff
@@ -12,15 +12,50 @@ import { StringifyListWithConjunction } from "./StringifyList";
...
+  return (
+    <I18n>
+      {({ i18n }) => (
+        <a
+          onClick={() => {
+            window.gtag("event", "data-request");
+          }}
+          href={`${getDataRequestUrlFromLocale(
+            i18n.language as SupportedLocale
+          )}?prefill_Custom+Data+Request=${i18n._(t`BUILDING:`)}+${fullAddress}`}
```
toolness, line 46: this feels like a bit of a mouthful code-wise, I wonder if it'd be easier to read if factored out into a `generateLinkToDataRequestForm` function, like the one that got removed in this PR?

Also when I look closely at this I'm realizing that the value of the localized `BUILDING:` text should be wrapped in `encodeURIComponent`, I think?

toolness (approved):
This looks good!  Just one comment below... Also, this PR makes me realize that it could be useful to make our `<LocalizedOutboundLink>` in the tenant platform into a npm package so this code (and our org site) can use it, so I filed that as https://github.com/JustFixNYC/tenants2/issues/2020.

client/src/components/PropertiesSummary.tsx
sraby: yes! good point

Issue #337, opened 2020-07-24
Consider moving all NYCHA data to the back-end


toolness:
I noticed that we've got a relatively large (137k uncompressed) JSON file at `client/src/data/nycha_bbls.json` that is included in our JS bundle.  It seems to only be looked up one BBL at a time, for the current building the user is looking at, so it might make sense to move this file into the back-end, and automatically include the data in address search query results, so that clients don't need to make extra network requests to obtain it.  Doing this should result in a much smaller JS bundle.

sraby:
Fixed by #447

PR #447, opened 2021-04-22
Grab development-level statistics for NYCHA buildings live from nycdb

sraby:
This PR makes use of our new `nycha_bbls_18` table in nycdb, allowing us to detect whether a building address is part of NYCHA, and then subsequently find data about the development (like # of evictions to date) directly via a live connection to the nycdb database. 

While there are some minor changes to the front end, the meat of this PR is in the evolution of the API endpoint at `api/address/buildinginfo`, which when given a bbl (like `api/address/buildinginfo?bbl=2037250001`), will run an [updated SQL query](https://github.com/JustFixNYC/who-owns-what/blob/08e20a20c2101531a394169e821aa980f1cae092/wow/sql/address_buildinginfo.sql) to grab statistics on the bbl's NYCHA development if it exists.

There is a rare edge case (less than 10 bbls city-wide) where one particular BBL is shared between two different NYCHA developments. For example, bbl `1018550001` is part of both `DOUGLASS I` and `DOUGLASS II`. We handle these cases by aggregating eviction/unit counts across _both_ developments, and showing both development names listed like "DOUGLASS I / DOUGLASS II" on the top of the page.


sraby:
Fixes #337

wow/sql/address_nychastats.sql
```diff
@@ -0,0 +1,29 @@
+WITH DEVS AS (
+	SELECT 
+		DISTINCT DEVELOPMENT 
+	FROM NYCHA_BBLS_18
+	WHERE DEVELOPMENT !~ 'POLICE SERVICE AREA'
```
samaratrilling, line 5: A question
`!~` means `does not match regular expression, case sensitive`, right? Just wondering if the data ever would be expected to have a lowercase or mixed case version of this, in which case we might want `!~*`
https://www.postgresql.org/docs/8.1/functions-matching.html#FUNCTIONS-POSIX-TABLE

Also this is probably premature optimization but I think if the list of developments ever gets huge, using `LIKE` instead of a regex might be faster.

wow/sql/address_nychastats.sql
```diff
@@ -0,0 +1,29 @@
...
+		N.BBL,
+		P.UNITSRES, 
+		E.EVICTIONS
+	FROM NYCHA_BBLS_18 N
+	LEFT JOIN PLUTO_19V2 P USING(BBL)
+	LEFT JOIN (
+		SELECT BBL, COUNT(*) EVICTIONS FROM MARSHAL_EVICTIONS_ALL 
+		WHERE RESIDENTIALCOMMERCIALIND = 'RESIDENTIAL'
+		GROUP BY BBL
+	) E USING(BBL)
```
samaratrilling, line 21: What does this E mean?

samaratrilling (approved):
Just some questions for my own edification, nothing blocking!

client/src/components/APIDataTypes.ts
```diff
@@ -126,6 +126,18 @@ export type BuildingInfoResults = {
   result: BuildingInfoRecord[];
 };
 
+// TYPES ASSOCIATED WITH NYCHA STATS QUERY:
+
+export type NychaStatsRecord = {
+  development: string | null;
```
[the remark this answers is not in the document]
sraby: yes! Let me add some more detailed comments here for the fields as they're a bit ambiguous

wow/sql/address_nychastats.sql
sraby: So `E` is the name I'm assigning to the subquery inside the preceding parentheses (short for "Evictions"). I tried to keep it consistent with the alias names I was assigning the other tables in the query (`N` and `P`):
```
FROM NYCHA_BBLS_18 N
LEFT JOIN PLUTO_19V2 P...
```

client/src/state-machine.ts
```diff
@@ -179,8 +170,9 @@ async function getSearchResult(addr: SearchAddressWithoutBbl): Promise<WowState>
     };
   } else if (apiResults.addrs.length === 0) {
     const buildingInfoResults = await APIClient.getBuildingInfo(apiResults.geosearch.bbl);
-    const nychaData = getNychaData(apiResults.geosearch.bbl);
     const buildingInfo = buildingInfoResults.result[0];
+    const nychaStatsResults = await APIClient.getNychaStats(apiResults.geosearch.bbl);
```
toolness, line 174: Hmm, does this effectively mean that we're making two separate API calls to the server for every building a user looks up?  I'm wondering if we actually want to just add NYCHA info to the `getBuildingInfo()` API endpoint instead, but that's potentially a premature optimization... Might be worth filing an issue to consider combining the two endpoints in the future, though?

Also note that because this code consists of two _sequential_ `await` calls, this effectively means that we're making one request, waiting for the result, and then making the other request, which could be slower than e.g. using [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) to run them in parallel, which could be faster (although it depends on how our server processes requests too).  Again, probably not a big deal in the grand scheme of things, but it might be worth doing a bit of measurement up-front using e.g. the [Chrome network panel](https://developer.chrome.com/docs/devtools/network/) to see if/how performance is affected.

toolness (approved):
Nice, this looks great!  A few comments below but no blockers.

client/src/state-machine.ts
sraby: @toolness I was inspired by this comment and decided to include the nycha stats SQL query as part of the building info query so we only need to make one network request. Do you mind reviewing my work once more?

client/src/components/APIDataTypes.ts
```diff
@@ -120,25 +120,16 @@ export type BuildingInfoRecord = {
...
-export type BuildingInfoResults = {
-  result: BuildingInfoRecord[];
-};
-
-// TYPES ASSOCIATED WITH NYCHA STATS QUERY:
-
-export type NychaStatsRecord = {
-  /** The name of the NYCHA development */
+  /** The name of the NYCHA development (e.g. "SOTOMAYOR HOUSES"). NULL if building is not part of NYCHA. */
   development: string | null;
```
toolness: Oof, since this is now part of general building info, do you think these NYCHA-specific fields should have slightly different names?  For instance, what if they were called `nycha_development`, `nycha_dev_evictions`, and `nycha_dev_unitsres`, just so clients (including e.g. Jade, who might be using this endpoint) aren't confused?

I guess that begs the question of why we don't put this information under a `nycha` object, which I would be an additional hassle for us right now, but might make things easier down the road... for example, we'd just have to test if `nycha` is `null` rather than having to assert that all three NYCHA-specific fields are non-null.  I'll leave those decisions up to you though.

toolness (approved):
Thanks, this looks good!  Just  one comment, but it's not a blocker.

client/src/components/APIDataTypes.ts
sraby: Good point about appending `nycha_` to each nycha-specific field. For clarification about your second point, we actually don't check if _all three_ NYCHA-specific fields are non-null, we only check `developments` as the other two can conceivably be NULL simply if there are no evictions or no res. units listed on PLUTO. So, I'm gonna file this as an issue now but since we only check to see if `developments` is populated it doesn't seem to add much complexity from my end IMO. Let me know if I'm missiong something

client/src/components/APIDataTypes.ts
```diff
@@ -120,6 +120,15 @@ export type BuildingInfoRecord = {
...
   formatted_address: string;
+  /** The name of the NYCHA development (e.g. "SOTOMAYOR HOUSES").
+   * NULL if building is not part of NYCHA */
+  nycha_development: string | null;
+  /** Total executed residential evictions (since 2017) accross the building's entire NYCHA development.
+   * NULL values either mean 0 evictions took place across the development or building is not part of NYCHA. */
+  nycha_dev_evictions: number | null;
+  /** Total residential units accross the building's entire NYCHA development.
+   * NULL values either mean PLUTO listed 0 residential units across the development or building is not part of NYCHA. */
+  nycha_dev_unitsres: number | null;
```
sraby, line 131: @toolness I updated these comments to be more descriptive of what NULL values mean in each field here

toolness: Ohh nice that's great!  Cool, you can disregard my other comment about having a `nycha` object then, thanks!

sraby: oh ok!

PR #448, opened 2021-04-26
Use assertion helpers from justfix-ts.

### LINKED ISSUE #395: Use assertNotUndefined() from @justfixnyc/util.
I moved the assertion helpers into the package in https://github.com/JustFixNYC/justfix-ts/pull/9, so we should start using them from that package and get rid of them in this codebase.  I think that the only one this project uses is `assertNotUndefined()`.

samaratrilling:
Fixes #395


sraby (approved):
Awesome, thank you!

toolness (approved):
This is great, thanks!

PR #457, opened 2021-05-20
Add hpd_complaints data to wow_bldgs table

sraby:
This adds three new columns to the `wow_bldgs` table that we use to source all the data presented on the "overview" tab. The three columns are:
- `totalcomplaints`: total hpd_complaints received by 311 ever
- `recentcomplaints`: total hpd_complaints received by 311 over the last 3 years
- `recentcomplaintsbytype`: an array of complaint _types_ and their frequency, ordered from most frequent to least, over the last 3 years

Note that I chose the 3 year cutoff totally arbitrarily. My thinking was that folks would want to know what kinds of complaints have been issued most recently, but that's obviously subjective...

### DIFF
--- a/sql/create_bldgs_table.sql
+++ b/sql/create_bldgs_table.sql
@@ -70,7 +70,7 @@ select distinct on (registrations.bbl)
   registrations.*,
   coalesce(violations.total, 0)::int as totalviolations,
   coalesce(violations.opentotal, 0)::int as openviolations,
-  coalesce(complaints.totalcomplaints, 0):: int as totalcomplaints,
+  coalesce(complaints.totalcomplaints, 0)::int as totalcomplaints,
   complaints.complaintsbytype,
   pluto.unitsres,
   pluto.yearbuilt,
--- a/sql/create_bldgs_table.sql
+++ b/sql/create_bldgs_table.sql
@@ -35,12 +35,43 @@ rentstab as (
     coalesce(uc2019, 0) - coalesce(unitsstab2007, 0) rsdiff
   from rentstab_summary
   left join rentstab_v2 using(ucbbl)
+),
+
+complaints as (
+  select 
+    bbl, 
+    sum(countcomplaints) as totalcomplaints, 
+    json_agg(
+        json_build_object(complainttype,countcomplaints) 
+        order by countcomplaints desc) complaintsbytype
+  from 
+    -- ----------
+    -- For every bbl, grab frequency of complaints by type 
+    -- ----------
+    (
+        select 
+            bbl, 
+            case 
+                when majorcategory = any('{UNSANITARY CONDITION,GENERAL}') then minorcategory
+                else majorcategory end 
+            as complainttype,
+            count(*) countcomplaints
+        from hpd_complaints h
+        left join hpd_complaint_problems using(complaintid)
+        -- Limit complaints to those received within the last 3 years
+        where h.receiveddate > CURRENT_DATE - '3 YEARS'::INTERVAL
+        group by bbl, complainttype
+    ) subtable
+    -- ----------
+  group by bbl
 )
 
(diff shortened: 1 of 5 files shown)


sraby:
[ch7157]

sql/create_bldgs_table.sql
```diff
@@ -35,12 +35,43 @@ rentstab as (
...
   left join rentstab_v2 using(ucbbl)
+),
+
+complaints as (
+  select 
+    bbl, 
+    sum(countcomplaints) as totalcomplaints, 
+    json_agg(
+        json_build_object(complainttype,countcomplaints) 
+        order by countcomplaints desc) complaintsbytype
```
sraby, line 46: Soooo, I was on the fence regarding exactly what structure I wanted to use in building these json outputs. I originally structured these outputs to have separate key/value pairs for both the complainttype and the count, so outputs in the SQL table would look like this:
```
[{type: "HEAT/HOT WATER", count: 4}, {type: "PESTS", count: 2}]
```
This seemed to be the standard way to set up the json object given that both the type and count fields are considered "data." 

However, I also took into account that whatever raw format the json object is in is what users see when they click the "Export Data" button and download the raw data for a given portfolio. Given that, I thought it would be a lot more digestible and readable for a user to see this json object in a more condensed form, with the complainttypes as keys and the counts as values:
```
[{"HEAT/HOT WATER": 4}, {"PESTS": 2}]
```
This is what I landed on, but I just wanted to check whether this would make sense in the long run. I didn't see any major issue in trying to pull these values into our front-end, but lmk!

samaratrilling (approved):

sql/create_bldgs_table.sql
toolness: Hmm, so there are a few advantages I see with the "standard" (first) approach you described:

1. It's really straightforward to define this kind of type in TypeScript.  It's also a natural fit for an algebraic type, so if for whatever reason you wanted to add extra fields that are specific to particular types of repairs, it would be really easy.

2. In general, it's easy to add extra information to each blob in a forwards-compatible way. For instance, if for some reason you wanted to mention, say, the date of the most recent complaint in a particular category, you could just add a `latestComplaintDate` field without any issue. Out-of-date front-ends would still be fine with parsing the new object. This is in contrast to the second format you provided, where adding any additional information would either require the value of each object to change from a number to an object, or something else that out-of-date front-ends might explode on.

That said, readability of the export data feed also sounds important, so I'm not really sure which way to go here.  I guess I might lean towards an internal representation that's more verbose (the first option) but have the exported CSV perhaps contain the more condensed form, if that's even a possibility, but I forget enough about the actual implementation right now that I'm not sure if it is!

sraby: Ok, given your two comments, I decided to refactor the SQL to spit out the "standard" json format, and will work to make sure the output in the data export is readable (I think we have some mechanism to do this for other fields too).

sql/create_bldgs_table.sql
```diff
@@ -35,12 +35,43 @@ rentstab as (
...
+            bbl, 
+            case 
+                when majorcategory = any('{UNSANITARY CONDITION,GENERAL}') then minorcategory
+                else majorcategory end 
+            as complainttype,
+            count(*) countcomplaints
+        from hpd_complaints h
+        left join hpd_complaint_problems using(complaintid)
+        -- Limit complaints to those received within the last 3 years
+        where h.receiveddate > CURRENT_DATE - '3 YEARS'::INTERVAL
```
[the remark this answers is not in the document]
sraby: Good point @toolness! To account for this, I will make sure to include `totalcomplaints` and `recentcomplaints` as two separate fields in the wow_bldgs table.

samaratrilling (approved):

PR #460, opened 2021-05-26
Improve format for `recentcomplaintsbytype` column in CSV export

sraby:
So, given that #457 has been merged and deployed to our auto-updating instance of NYCDB, the new hpd_complaints data that we added is now, by default, part of the CSV data export that users receive when clicking on the "Export Data" button on WOW. We can think of this as a nice little "soft launch" (or even an "easter egg") of the new data before we actually bake it in to our front-end somewhere.

In the meantime, though, I want to format the `recentcomplaintsbytype` column as the default json structure is hard to read.

### DIFF
--- a/wow/views.py
+++ b/wow/views.py
@@ -131,8 +131,10 @@ def address_export(request):
     first_row = addrs[0]
 
     for addr in addrs:
-        addr['ownernames'] = csvutil.stringify_owners(addr['ownernames'])
-        addr['recentcomplaintsbytype'] = csvutil.stringify_complaint_types(addr['recentcomplaintsbytype'])
+        addr["ownernames"] = csvutil.stringify_owners(addr["ownernames"])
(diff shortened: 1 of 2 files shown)


wow/views.py
```diff
@@ -131,7 +131,10 @@ def address_export(request):
     first_row = addrs[0]
 
     for addr in addrs:
-        addr['ownernames'] = csvutil.stringify_owners(addr['ownernames'])
+        addr["ownernames"] = csvutil.stringify_owners(addr["ownernames"])
```
samaratrilling, line 134: not a huge deal but the rest of this file uses single quotes, so maybe we could continue in that tradition instead of the double quotes here.

samaratrilling (approved):
One small comment, otherwise LGTM

wow/views.py
sraby: oh weird! sorry about that, it must've been the Python formatter I was using. Will fix now

PR #461, opened 2021-05-26
Add hpd complaints data to Portfolio Tab

sraby:
This PR adds a new section to the react table on the Portfolio Tab called "HPD Complaints", that includes 3 new columns of data:
- **Total**: count of all complaints ever called in to 311
- **Last 3 Years**: count from just last 3 years
- **Top Complaint**: most common complaint type from last 3 years

One funky thing that this PR does is [add a new function to our helpers.ts file](https://github.com/JustFixNYC/who-owns-what/pull/461/files#diff-a7c84dff5c3235f5b41a5d1a818116d17fd0f0edb64977c68ab99a3fe97d660c) that internationalizes the complaint types that show up in open data.

### DIFF
--- a/client/src/util/helpers.ts
+++ b/client/src/util/helpers.ts
@@ -228,7 +228,7 @@ export default {
     return arr.join(" ");
   },
 
-  getI18nComplaintType(complaintType: string, i18n: I18n) {
+  getTranslationOfComplaintType(complaintType: string, i18n: I18n) {
     const translatedType = hpdComplaintTypeTranslations.get(complaintType);
     if (!translatedType) {
       reportError(`The HPD Complaint type called "${complaintType}" doesn't have a translation.`);
--- a/client/src/util/helpers.ts
+++ b/client/src/util/helpers.ts
@@ -230,8 +230,9 @@ export default {
 
   getI18nComplaintType(complaintType: string, i18n: I18n) {
     const translatedType = hpdComplaintTypeTranslations.get(complaintType);
-    if (translatedType) {
-      return i18n._(translatedType);
-    } else return complaintType;
+    if (!translatedType) {
+      reportError(`The HPD Complaint type called "${complaintType}" doesn't have a translation.`);
+      return complaintType;
+    } else return i18n._(translatedType);
   },
 };
--- a/client/src/util/helpers.ts
+++ b/client/src/util/helpers.ts
@@ -231,7 +231,7 @@ export default {
   getTranslationOfComplaintType(complaintType: string, i18n: I18n) {
     const translatedType = hpdComplaintTypeTranslations.get(complaintType);
     if (!translatedType) {
-      reportError(`The HPD Complaint type called "${complaintType}" doesn't have a translation.`);
+      reportError(`The HPD Complaint type "${complaintType}" isn't internationalized`);
       return complaintType;
     } else return i18n._(translatedType);
   },
--- a/client/src/util/helpers.ts
+++ b/client/src/util/helpers.ts
@@ -6,6 +6,8 @@ import { deepEqual as assertDeepEqual } from "assert";
 import { SupportedLocale } from "../i18n-base";
 import { SearchAddressWithoutBbl } from "components/APIDataTypes";
 import { reportError } from "error-reporting";
+import { t } from "@lingui/macro";
+import { I18n } from "@lingui/core";
 
 /**
  * An array consisting of Who Owns What's standard enumerations for street names,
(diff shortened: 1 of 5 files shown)


sraby:
[ch7313]

client/src/util/helpers.ts
```diff
@@ -188,4 +227,12 @@ export default {
     });
     return arr.join(" ");
   },
+
+  getTranslationOfComplaintType(complaintType: string, i18n: I18n) {
+    const translatedType = hpdComplaintTypeTranslations.get(complaintType);
+    if (!translatedType) {
+      reportError(`The HPD Complaint type "${complaintType}" isn't internationalized`);
```
sraby, line 234: @samaratrilling, I remember you bringing up a good point on _when_ we should check whether there is a new complaint type from HPD data that we don't have translated/internationalized—specifically, that we should throw this kind of error from our [nycdb-k8s-loader](https://github.com/JustFixNYC/nycdb-k8s-loader/) when we actually generate the `wow_bldgs` table every morning vs. waiting for the issue to emerge from a user interacting with the front end. I thought this solution could work for now, so I'll file an issue for your solution once this is merged.

client/src/util/helpers.ts
```diff
@@ -188,4 +227,12 @@ export default {
     });
     return arr.join(" ");
   },
+
+  getTranslationOfComplaintType(complaintType: string, i18n: I18n) {
+    const translatedType = hpdComplaintTypeTranslations.get(complaintType);
```
samaratrilling, line 232: I'm curious - instead of having the map above, could you just do

```
const translatedType = t`{complaintType}`
```
or something like that? Would that work?

client/src/util/helpers.ts
```diff
@@ -26,6 +28,43 @@ const hpdNumberTransformations = [
   ["TENTH", "10"],
 ];
 
+const hpdComplaintTypeTranslations = new Map([
```
samaratrilling, line 31: This feels iffy to me, but having read https://github.com/lingui/js-lingui/issues/426, I guess it's not an uncommon problem.

Eventually we might want to think about storing these in a database instead of in code kind of like in the issue above. Is this the first time we've ever had to translate raw data? It feels like this is setting a precedent and it would be great to have just a smidge longer eng-specific chat on how we want to handle this in the future.

sraby: Oh nice good find on that issue from lingui! And that totally makes sense, let's flag this as something to discuss next week at engineering check in.

client/src/util/helpers.ts
toolness: Ah unfortunately we can't do <code>t`{complaintType}`</code> because (IIRC) `t` is actually a [babel macro](https://babeljs.io/blog/2017/09/11/zero-config-with-babel-macros) provided by lingui, which basically means that it doesn't behave like a normal piece of code--as @sraby mentioned, it's used when we run `yarn lingui:extract` to extract literal strings to be translated, and it basically needs to have every single potential string spelled out for it.

client/src/util/helpers.ts
toolness: Hmm, I am intrigued by the possibility of doing this translation on the server-side rather than the client-side.

In the tenant platform, our APIs are actually prefixed by a locale identifier (e.g. `/en/graphql` or `/es/graphql`) which allows the API response to be localized.  However, in WoW, it's not prefixed by a locale (it's just `/api/blah`), which makes it harder to do the localization on the server-side.  Still, we could potentially migrate to that kind of URL structure, or we could have the locale get passed in via a querystring argument...

Anyways, I don't think it's _that_ big a deal because our list of strings to be localized isn't too huge.

Storing the translations in a database is possible but comes with its own headaches--most significantly, since dev/staging/production use entirely different databases, we'd have potentially different behavior from them, and it raises the issue of keeping them "in sync"... we could do fancy things with migrations to make sure they are in sync, but then that's functionally the same thing as just translating them via Crowdin.  I think it might be a bigger deal if this list of categories was constantly changing and we wanted to be able to localize it without having to issue a PR, but since the list appears to be fairly stable, I think translating the strings in Crowdin is probably fine.

Regardless of the outcome though I think this is a great thing to flag @samaratrilling, thanks for doing tha!

sraby: awesome, thanks for the feedback both of you! Given this analysis, I think it's safe to merge this for now although I will make a separate issue to explore doing the translation on the server-side as mentioned.

samaratrilling (approved):

PR #469, opened 2021-06-09
Add `housingtype` API endpoint to the WOW back-end

sraby:
This PR adds a new WOW API endpoint at `/api/address/housingtype` that takes a parameter `bbl` and determines from data on nycdb whether the bbl could be classified as 'NYCHA',  'RENT-STABILIZED', 'MARKET-RATE', or 'UNKNOWN'. 

Note that the response from this API endpoint will provide one of three things:
1. An object array with one element that indicates the given housing type for a bbl.
Ex: `/api/address/housingtype?bbl=4082490088` --> `{"result": [{"housingtype": "MARKET-RATE"}]}`
2. An error response, from a network error or because the bbl provided in the request wasn't 10 digits.
Ex: `/api/address/housingtype?bbl=408249008X` --> `{"error": "Bad request", "validationErrors": {"bbl": [{"message": "This should be a 10-digit padded BBL.", "code": "invalid"}]}}` 
3. An empty array, which should indicate that the bbl provided in the request was of valid format, but that ultimately the PLUTO dataset from Dept. of City Planning didn't have a record of that BBL existing.
Ex: `/api/address/housingtype?bbl=4082491000` -->`{"result": []}`


sraby:
[ch7368]

samaratrilling (approved):
This looks great! I'm having some trouble getting wow back up and running on my machine so I haven't tested it yet but code-wise it looks good.

toolness:
This is looking great!

One question I have is whether we want this API endpoint to live on WoW or the tenant platform.  The reason I ask is because, if this endpoint is being made specifically for the NYCx bot (and has no immediate use by WoW itself), there's a possibility that we might eventually want the NYCx bot to talk to the Tenant Platform too--e.g., so that it can look up / modify user data and such.

The TP is already coupled to WoW's database, so coupling it further by putting this API endpoint on it wouldn't significantly complicate things, whereas having NYCx be coupled to _both_ WoW and the TP might not be great.  Anyways, just a thought... since the TP and WoW both use Django it's not that big a deal anyways, since we can fairly easily move code between the two codebases.

wow/tests/test_views.py
```diff
@@ -90,6 +90,23 @@ def test_it_works(self, db, client):
...
+        res = client.get('/api/address/housingtype?bbl=4082490088')
+        assert res.status_code == 200
+        assert res.json()['result'] is not None
+
+    def test_it_returns_empty_result_for_unrecognized_bbl(self, db, client):
+        res = client.get('/api/address/housingtype?bbl=4082491000')
+        assert res.status_code == 200
+        assert res.json()['result'] == []
+
+
```
toolness, line 109: noooice tests

sraby:
good point @toolness, I made a flag to discuss this point tomorrow during engineering review

toolness:
Oh I should also add that yet _another_ option is simply to give the serverless NYCx code (at https://github.com/JustFixNYC/dialogflow-fun, unless @samaratrilling has forked it) read-only access to the NYCDB/WoW database.  I generally feel like it's okay to couple our tools directly to NYCDB/WoW db since their schema doesn't change often and it's not a security hazard (it would be if we had user data in those databases, but we don't).

sraby:
As @samaratrilling said: 
> We agreed to instead have the fulfillment server query nycdb via sql directly. This is so we have greater flexibility in changing the sql query without needing a PR on a different repo, and also minimizes latency (tho this may not be a problem with our small scale user base).

Closing this PR for now.

PR #471, opened 2021-06-14
Update FullStory script tag

sraby:
We got this message from FullStory via email:

> We are preparing to update some of our infrastructure, and as a result if you haven’t updated your snippet since October 30, 2019 you will need to update the FullStory snippet on your site by July 13, 2021 to avoid a disruption in session recording.

So, this PR is simply updating the snipped that FullStory provided.


sraby:
[ch7395]

PR #474, opened 2021-06-15
Handle CSV export when ownernames is null.

toolness:
Sometimes the `ownernames` field of the query used to export CSVs is `NULL`, which we haven't been accounting for until now.


toolness:
[ch7379]

sraby (approved):
Noice!!

PR #477, opened 2021-06-15
Add `allcontacts` field to registration_with_contacts table

sraby:
In preparation for implementing Tahnee's new redesign of "Landlord Info" on the WOW Overview Tab, this PR adds a new column to the `registration_with_contacts` custom table that includes _all_ HPD contact names and corporation names, with associated addresses, in one single json object. 

In turn, this PR will subsequently add this new field called `allcontacts` to the end of the `wow_bldgs` table, since the SQL to generate the `wow_bldgs` table arbitrarily selects all columns from the `registration_with_contacts` table. Therefore, we should be able to start using this new column of data on the front-end right away.

### DIFF
--- a/sql/registrations_with_contacts.sql
+++ b/sql/registrations_with_contacts.sql
@@ -80,14 +80,16 @@ LEFT JOIN (
         jsonb_build_object(
           'title', type, 
           'value', (firstname || ' ' || lastname),
-          'address', jsonb_build_object(
-            'housenumber', businesshousenumber,
-            'streetname', businessstreetname, 
-            'apartment', businessapartment, 
-            'city', businesscity,
-            'state', businessstate,
-            'zip', businesszip
-          )
+          'address', case when businessstreetname is not null then 
+	          jsonb_build_object(
+	            'housenumber', businesshousenumber,
+	            'streetname', businessstreetname, 
+	            'apartment', businessapartment, 
+	            'city', businesscity,
+	            'state', businessstate,
+	            'zip', businesszip
+	          ) 
+	        else null end
         )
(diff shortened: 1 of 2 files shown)


sql/registrations_with_contacts.sql
```diff
@@ -71,7 +72,38 @@ LEFT JOIN (
...
-      AS ownernames,
+    AS ownernames,
+
+    coalesce(
+      jsonb_agg(
+        distinct
+        jsonb_build_object(
+          'title', type, 
+          'value', (firstname || ' ' || lastname),
+          'address', nullif(concat_ws(' ', businesshousenumber, businessstreetname, businessapartment, businesszip), '')
```
sraby, line 83: Do we think it's worth it to define this `address` prop as it's own object with keys for each part of the address? Perhaps we will need bits and pieces of the address handy in the future? The downside is that it adds unneeded complexity to the overall json, and will affect readability in the export CSV (although it's pretty bad to begin with...). What do folks think?

toolness: interesting--regarding the export CSV, i wonder if we could join together all the fields into a string, analogous to the way we join some arrays of objects into comma-separated strings in the CSV export?  (except it'd be an object, not an array, we're joining, but same general concept)

toolness: That said, overall I do like the idea of showing the individual fields.

toolness (approved):
Nice!  Do you think there's any easy way to test this code, e.g. given some initial raw data, showing what the resultant `allcontacts` field should look like? Not really a big deal if not.

PR #478, opened 2021-06-16
Assign and fix type of `allcontacts` field in front-end and back-end

sraby:
This PR adds type casting in our back-end python code and and new type definitions to the front-end APIDataTypes.txt file to accommodate the new `allcontacts` feild added to the `wow_bldgs` table. In doing so, this PR also adds some tooling to format this new column in the "Export Data" feature.


client/src/components/APIDataTypes.ts
```diff
@@ -19,6 +19,17 @@ type HpdOwnerContact = {
   value: string;
```
toolness, line 19: I know this isn't part of the original PR, but calling this `value` is weird to me, I feel like it should be `fullName` or something.  But if we've already got production code looking at this field then definitely don't worry about it!!

wow/tests/test_datautils.py
```diff
@@ -16,3 +16,18 @@ def test_str_or_none_works(self):
...
+        assert json_or_none(sample_json_string) == [{
+            "foo": "bar",
+            "beep": {
+                "boop": "blah"
+            }}
+        ]
+        assert json_or_none(sample_json_string_with_null) == [{
+            "foo": "bar",
+            "beep": None
+        }]
```
toolness, line 33: AW YEAH THE TESTS NOICE

toolness: Oh this would be an excellent candidate for the use of [`pytest.mark.parametrize`](https://docs.pytest.org/en/6.2.x/parametrize.html) by the way.  The nice thing is that it actually will turn this one test into three separate tests with a minimum of fuss (which makes it easier to figure out what's going wrong if any of them ever break).

sraby:
[ch7440]

wow/views.py
```diff
@@ -39,6 +39,7 @@ def clean_addr_dict(addr):
         "bin": str(addr['bin']),
         "lastsaleamount": int_or_none(addr['lastsaleamount']),
         "registrationid": str(addr['registrationid']),
+        "allcontacts": json_or_none(addr['allcontacts'])
```
toolness, line 42: Er wait... why is this coming out of the database as stringified JSON (rather than parsed JSON) again?  I remember this is what you pinged me about on Slack but I'm unclear on why it's happening.  Maybe document it in the code so that future readers aren't befuddled?

toolness (approved):
Cool this looks good!  Just a few comments and a question.

client/src/components/APIDataTypes.ts
sraby: @toolness I agree! it's a horribly vague name for this property that unfortunately predates my time working on WOW... maybe we can file an issue to try and address this in the future?

toolness: Oh gotcha yeah no worries!!

wow/views.py
toolness: Ooo I think it _might_ have something to do with this: https://code.djangoproject.com/ticket/31991

While we are executing raw SQL, we're doing it through Django as an intermediary, which might be the culprit.  Gah so annoying.

wow/tests/test_datautils.py
sraby: you got it!

wow/views.py
toolness: Figured out a better fix, just filed #480!

PR #479, opened 2021-06-16
Restructure how landlord contact info is displayed on the Overview Tab

sraby:
This PR implements [Tahnee's designs](https://www.figma.com/file/w2eo2tq6YEbVW06rFF0jA4/Summer-2021-Sprints?node-id=47%3A2) that more clearly organize and spell out all of the contact names and corporations associated with a building on WOW. While mostly a styling and html structure change, this PR makes use of #477 in that it properly shows all entity names with their associated business addresses (unlike before).

Beyond updating the Detail View on the overview tab, this PR changes the look and feel of the "useful links" section to make it more compact, which shows up on both the overview and timeline tabs.

One small win: the various expandable menus and text boxes in this PR were ALL implemented with the `<summary><details>` native HTML elements! No special JS needed! Also they are all accessible via tab focus.

### DIFF
--- a/client/src/components/Accordion.tsx
+++ b/client/src/components/Accordion.tsx
@@ -3,14 +3,31 @@ import React from "react";
 import chevron from "../assets/img/accordionchevron.svg";
 import "styles/Accordion.scss";
 
-export const Accordion = (props: { question: string; children: React.ReactNode }) => {
+type AccordionProps = {
+  title: string;
+  /**
+   * Alternate text to show as the "title" part of the accordion,
+   * once it has been opened.
+   */
+  titleOnOpen?: string;
+  children: React.ReactNode;
+};
+
+export const Accordion: React.FC<AccordionProps> = ({ title, titleOnOpen, children }) => {
   return (
-    <details className="Accordion">
+    <details className="accordion">
       <summary>
-        {props.question}
+        {titleOnOpen ? (
+          <>
+            <span className="title-default">{title}</span>
+            <span className="title-on-open">{titleOnOpen}</span>
+          </>
+        ) : (
+          title
+        )}
         <img src={chevron} className="icon mx-1 float-right" alt="Open" />
       </summary>
-      <div>{props.children}</div>
+      <div>{children}</div>
     </details>
   );
 };
--- a/client/src/components/Accordion.tsx
+++ b/client/src/components/Accordion.tsx
@@ -1,0 +1,16 @@
+import React from "react";
+
+import chevron from "../assets/img/accordionchevron.svg";
+import "styles/Accordion.scss";
+
+export const Accordion = (props: { question: string; children: React.ReactNode }) => {
+  return (
+    <details className="Accordion">
+      <summary>
+        {props.question}
+        <img src={chevron} className="icon mx-1 float-right" alt="Open" />
+      </summary>
+      <div>{props.children}</div>
+    </details>
+  );
+};
--- a/client/src/components/DetailView.tsx
+++ b/client/src/components/DetailView.tsx
@@ -173,7 +173,18 @@ class DetailViewWithoutI18n extends Component<Props, State> {
                           </ul>
                         </div>
                       </div>
-                      <Accordion question="MAGGIE MCCORMICK">Landlord</Accordion>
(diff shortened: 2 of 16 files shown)


client/src/components/Accordion.tsx
```diff
@@ -0,0 +1,33 @@
...
+   * once it has been opened.
+   */
+  titleOnOpen?: string;
+  children: React.ReactNode;
+};
+
+export const Accordion: React.FC<AccordionProps> = ({ title, titleOnOpen, children }) => {
+  return (
+    <details className="accordion">
+      <summary>
```
toolness, line 19: NOICE `<details>` and `<summary>` FTW

client/src/components/DetailView.tsx
```diff
@@ -44,6 +47,28 @@ const SocialShareDetailView = () => (
...
+          <p>
+            While the legal owner of a building is often an “LLC” company, these names and business
+            addresses registered with HPD offer a clearer picture of who the landlord really is.
+          </p>
+          <p>
+            People listed here as “Head Officer” or “Owner” usually have ties to building ownership,
+            while “Site Managers” are part of management. That being said, these names are self
+            reported, so they can be misleading.
+          </p>
+          <p>Learn more about HPD registrations and how they power this tool on the About page.</p>
```
toolness, line 65: Should "About page" be hyperlinked to the about page?

client/src/components/DetailView.tsx
```diff
@@ -172,44 +195,56 @@ class DetailViewWithoutI18n extends Component<Props, State> {
...
+                              // Group all contact info by the name of each person/corporate entity
+                              Object.entries(_groupBy(detailAddr.allcontacts, "value"))
+                                // Move head officers and owners to the top of the list
+                                .sort((contact) =>
+                                  contact[1].find(
+                                    (o) => o.title === "HeadOfficer" || o.title.includes("Owner")
+                                  )
+                                    ? -1
+                                    : 0
+                                )
```
toolness, line 217: It might make the code more readable to factor this out into a separate well-named function, e.g. `sortContactsByImportance()` or something?  That would also make it easier to unit test and such, if someone ever wanted to do that.

client/src/components/DetailView.tsx
```diff
@@ -172,44 +195,56 @@ class DetailViewWithoutI18n extends Component<Props, State> {
...
+                                              Helpers.formatHpdContactAddress(info.address)
+                                                .addressLine1
+                                            }
+                                            <br />
+                                            {
+                                              Helpers.formatHpdContactAddress(info.address)
+                                                .addressLine2
+                                            }
+                                          </>
+                                        )}
```
toolness, line 236: It might make the code a bit more readable to factor this out into a `<FormattedContactAddress>` component or something but really up to you.  That function could also just bind a variable to `Helpers.formatHpdContactAddress(info.address)` so you wouldn't have to call it twice.

wow/views.py
```diff
@@ -34,7 +34,6 @@ def log_unsupported_request_args(request):
 
 
 def clean_addr_dict(addr):
-    print("UM", addr)
```
toolness, line 37: hahaha that was probably accidentally left there by me, thanks for removing it!

toolness (approved):
This looks great!  I have a few comments on the code, and one concern regarding the visual design:

> image

In the screenshot above, "Learn more" is an interactive element, but it's styled like static text, which feels inconsistent.  It makes me wonder if fewer people might click on it b/c they might not think it's interactive, even though the label clearly indicates that it is.

Anyways, no blockers here, just some comments/suggestions!

toolness:
One last thing... this is what the new content looks like on IE11, which doesn't support `<details>`:

> image

I think it's actually fine, but the one thing that might be potentially confusing is the fact that "Learn more" doesn't actually do anything because the content is already expanded.  There might be ways to alter this so that the text is invisible if `<details>` isn't supported, but it's probably not a big deal--I think IE11 already reached end-of-life anyways.

client/src/components/DetailView.tsx
sraby: yes!!

PR #480, opened 2021-06-16
Automatically deserialize jsonb in SQL queries.

toolness:
This builds upon #478 by modifying the underlying database cursor to properly deserialize jsonb content from Postgres.  This is actually the default behavior of Python's Postgres package, psycopg2, but it's [overridden by Django][1]; this restores the default behavior within the scope of a single cursor.

A regression test has also been added to ensure that we'll know if this fix ever breaks (which could happen if Django decides to change its implementation somehow).

[1]: https://code.djangoproject.com/ticket/31991

### DIFF
--- a/wow/datautil.py
+++ b/wow/datautil.py
@@ -1,5 +1,4 @@
-import json
-from typing import List, Optional, Any
+from typing import Optional, Any
 
 
 def int_or_none(val: Any) -> Optional[int]:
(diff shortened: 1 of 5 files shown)


sraby (approved):
Wow amazing!! Thank you so much for your investigative work @toolness. I'm glad we got to the bottom of this :)

PR #482, opened 2021-06-17
Pull content for moratorium banner from Contentful.

toolness:
This is analogous to https://github.com/JustFixNYC/tenants2/pull/2125, but for our static sites (WoW's front-end is a static site).

Note that pulling from Contentful for common strings for our static sites is a bit unusual because:

* Not all our static sites use the same framework. For example, WoW uses Create React App, while our other sites use Gatsby.
* While all our static sites currently use Contentful in some capacity, they don't use the same Contentful _space_.  The common strings are all stored in a separate Contentful space to keep things DRY.

This attempts to deal with these issues by adding some code that can be encapsulated in its own module in the `@justfixnyc` npm namespace.

## To do

- [ ] Add documentation
- [ ] Pull most of the code out into its own module (moving it to the [justfix-ts](https://github.com/justFixNYC/justfix-ts) repo)


client/src/contentful-common-strings/pull.js
```diff
@@ -0,0 +1,71 @@
+//@ts-check
+
+/**
```
samaratrilling, line 3: Did you want to leave this comment in here? Is this some kind of Javascript typing convention I'm not aware of?

client/src/contentful-common-strings/pull.js
```diff
@@ -0,0 +1,71 @@
...
+ * @typedef {import("./index").ContentfulCommonStrings} CommonStrings
+ */
+
+let https = require("https");
+
+const ORIGIN = "https://cdn.contentful.com";
+const COMMON_STRING_TAG = "common";
+const SPACE_ID = "markmr2gi204";
+const ACCESS_TOKEN = "Fli_OMdKgUFw6tEX3uv6HqvptuG6A6jn9bZVPlHZj8E";
```
samaratrilling, line 12: Ah, this was what the email was complaining about?

client/yarn.lock
```diff
@@ -2244,11 +2244,6 @@
   resolved "https://registry.yarnpkg.com/@justfixnyc/geosearch-requester/-/geosearch-requester-0.0.6.tgz#af3144a8b599e45ad3ce5c2190fc60da81106475"
   integrity sha512-wHsY9DbiQUp7H/KiQ4+ejTShR6lY3iagiFhw4VoP1kljyem+WIwEkiZNnvUHkOAQIkTtPulLaiKHEMpzaw9xZQ==
 
-"@justfixnyc/react-common@^0.0.12":
```
samaratrilling, line 2247: Huh, so will this obviate the need for react-common across TP *and* WOW? is that the only thing react-common was doing??

samaratrilling (commented):
Thanks for the thorough description!

1. I'm certainly in favor of automatically re-deploying our static sites using a webhook whenever contentful strings change. Will this ever cause downtime (potentially a problem if we do it a lot), or does Contentful guarantee smooth rebuilds?

2. I'm asking myself, scale of 1-10, how kludge-y does this feel? How much extra is this getting us?
3. Some thoughts:
1. It is a major improvement in that non-engineers can at least edit the text for the banner themselves.

2. If there were another situation where we wanted the same string across all of our websites, would we feel comfortable following this convention of putting the string in contentful and then using `getContentfulCommonString`?
3. You mentioned we might need to manually trigger WOW's rebuild. Is there a reason we couldn't use webhooks for that too, or do something more similar to the tenant platform?

toolness:
> 1. I'm certainly in favor of automatically re-deploying our static sites using a webhook whenever contentful strings change. Will this ever cause downtime (potentially a problem if we do it a lot), or does Contentful guarantee smooth rebuilds?

Nope, Contentful guarantees smooth rebuilds!

> 2. I'm asking myself, scale of 1-10, how kludge-y does this feel? How much extra is this getting us?

Yeah, I don't feel like it's particularly kludge-y, but I _do_ feel like it's complex... however, given the different ways our sites work, it's also the simplest solution I could think of.  (The current solution of using a shared npm module was the simplest solution at the time too, but we definitely hadn't anticipated needing to change the banner several times over the span of a day!)

> 5. If there were another situation where we wanted the same string across all of our websites, would we feel comfortable following this convention of putting the string in contentful and then using `getContentfulCommonString`?

Yeah, that was my hope!  I'm not sure exactly what such a situation would be though...

> 6. You mentioned we might need to manually trigger WOW's rebuild. Is there a reason we couldn't use webhooks for that too, or do something more similar to the tenant platform?

Oh right, that actually has to do with some of my goals in adding Contentful integration to WoW... it's documented fully in #205, here's an excerpt:

> However, unlike our Gatsby-based sites, I'd like it to be _optional_ for developers to require access to Contentful in order to develop the site, rather than a requirement, particularly because the amount of long-form content is very minimal, well-defined, and unlikely to be updated frequently.
>
> Because of this, we're storing the JSON representation of our Contentful pages in the `client/data/` directory, and can re-pull from Contentful using the new `yarn contentful` script.

So I had heebie jeebies about tightly coupling WoW to Contentful (this is still something that concerns me--we're essentially relying on the charity of Contentful to give us a free account right now, and were they suddenly to rescind that, it'd put us in a bad spot).  And also, at the time of writing that, I had assumed that putting Contentful's access key in the repo would be bad, and I didn't want to bar volunteer contributors in particular from getting up-and-running solely because they lacked Contentful credentials.

That said, particularly if we decide that we want to update WoW's secondary pages more frequently, it won't be too hard to abandon this approach and instead follow the same one that this PR uses, i.e. always pull from the latest Contentful data at static front-end build time, and set up a webhook to rebuild the front-end whenever Contentful changes!

client/yarn.lock
toolness: Heh, for now it would, yeah.  The original goal was for `react-common` to contain all common JustFix-specific react components shared between all our products, but the only thing in there at the moment is the Covid Moratorium Banner.  But perhaps if we have a new need for a shared React component it might come back.  (We could also just have separate NPM packages for each individual component of course--I just find creating new NPM packages w/ their concomitant transpilation build pipelines + test suite configurations + etc to be a big hassle, even with our monorepo's DRY enhancements, so I decided to just put all the components in one package and let tree-shaking excise whatever a project decided not to reuse.)

client/src/contentful-common-strings/pull.js
toolness: Oh yeah, so `//@ts-check` is a magical comment that lets TypeScript know that even that this is a JS file, it (and its jsdoc comments) should be type-checked.  (I didn't write this in TS because I didn't want to have to deal with running it through a transpile pipeline to run it from the command-line, but I also wanted some of the benefits of TS, which TS happily provides--see their docs on [JS projects utilizing TypeScript](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html) for more info!)

client/src/contentful-common-strings/pull.js
toolness: yupppppppp 😠

toolness:
> > 1. I'm asking myself, scale of 1-10, how kludge-y does this feel? How much extra is this getting us?
> 
> Yeah, I don't feel like it's particularly kludge-y, but I _do_ feel like it's complex... however, given the different ways our sites work, it's also the simplest solution I could think of. (The current solution of using a shared npm module was the simplest solution at the time too, but we definitely hadn't anticipated needing to change the banner several times over the span of a day!)

Oh, I should mention that I'm feeling conflicted about whether it might be less complex if we just do the content fetching in the front-end at page load, rather than at build time.  In some ways it would be less complex (no need to set up any webhooks or have a script that pulls from contentful) but in other ways it could introduce new complexity--e.g. having to do things to avoid layout instability, dealing with the possibility of rate limiting from Contentful--so I'm kind of torn.

PR #493, opened 2021-06-22
Add hamburger menu to header on mobile

sraby:
This PR replaces our standard navbar of links in the header with a hamburger menu once the viewport is smaller than a desktop or landscape tablet. The hamburger menu includes all original links from the navbar, with the only difference being that instead of the `EN/ES` toggle, we now show the full language name of whatever language the user is _not_ currently seeing on the site (see the new `<LocaleSwitcherWithFullLanguageName>` component). I also added an overlay to gray out the main content of the page when the user has the hamburger menu open. 

TODO:
- [x] Decide whether to match styling between the menu overlay and the modal overlay, or keep them different

### DIFF
--- a/client/src/components/Modal.tsx
+++ b/client/src/components/Modal.tsx
@@ -19,8 +19,7 @@ const Modal = (props: ModalProps) => {
       left: 0,
       right: 0,
       bottom: 0,
-      backgroundColor: "rgba(255, 255, 255, 0.75)",
-      //backgroundColor: "rgba(69, 77, 93, 0.6)", // rgba equivalent of the $dark color (#454D5D)
+      backgroundColor: "rgba(69, 77, 93, 0.6)", // rgba equivalent of the $dark color (#454D5D)
     },
     content: {
       position: "absolute",
--- a/client/src/components/Modal.tsx
+++ b/client/src/components/Modal.tsx
@@ -20,6 +20,7 @@ const Modal = (props: ModalProps) => {
       right: 0,
       bottom: 0,
       backgroundColor: "rgba(255, 255, 255, 0.75)",
+      //backgroundColor: "rgba(69, 77, 93, 0.6)", // rgba equivalent of the $dark color (#454D5D)
     },
     content: {
       position: "absolute",
--- a/client/src/components/Modal.tsx
+++ b/client/src/components/Modal.tsx
@@ -41,7 +41,14 @@ const Modal = (props: ModalProps) => {
   };
 
   return (
-    <ReactModal isOpen={props.showModal} contentLabel="Modal" style={styles}>
+    <ReactModal
+      isOpen={props.showModal}
+      contentLabel="Modal"
+      style={styles}
+      onRequestClose={props.onClose}
+      shouldCloseOnOverlayClick={true}
+      shouldCloseOnEsc={true}
+    >
       <button className="ReactModal__Close btn btn-link" onClick={(e) => props.onClose(e)}>
         [ x ]
       </button>
--- a/client/src/containers/App.tsx
+++ b/client/src/containers/App.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from "react";
 import { BrowserRouter as Router, Switch, Route } from "react-router-dom";
 import { Trans, t } from "@lingui/macro";
+import FocusTrap from "focus-trap-react";
 
 import "styles/App.css";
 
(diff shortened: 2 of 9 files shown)


sraby:
[ch7477]

toolness:
Hmm, so one funky thing about this right now is that keyboard focus is a bit wonky:

https://user-images.githubusercontent.com/124687/123128320-0c443680-d419-11eb-976c-df26b2232618.mp4

When I use <kbd>Tab</kbd> to move focus past the end of the menu, either the menu should collapse and the translucent background should disappear and focus should go to the next focusable element outside of the menu (this is effectively what's done on app.justfix.nyc), OR the hamburger should act like a [focus trap/loop](https://medium.com/@im_rahul/focus-trapping-looping-b3ee658e5177) and focus should go to the beginning of the hamburger.  But either way, it should never be the case that keyboard focus is inside the dimmed background area!

client/src/containers/App.tsx
```diff
@@ -85,83 +85,126 @@ const WhoOwnsWhatRoutes: React.FC<{}> = () => {
...
+    <LocaleNavLink to={paths.about} key={2}>
+      <Trans>About</Trans>
+    </LocaleNavLink>,
+    <LocaleNavLink to={paths.howToUse} key={3}>
+      <Trans>How to use</Trans>
+    </LocaleNavLink>,
+    <a href="https://www.justfix.nyc/donate" key={4}>
+      <Trans>Donate</Trans>
+    </a>,
+  ];
```
toolness, line 103: Rather than returning an array, which requires you to add a `key` prop to each child, you could return a [React fragment](https://reactjs.org/docs/fragments.html) with e.g.:

```jsx
return <>
  <p>child one</p>
  <p>child two</p>
  <p>etc...</p>
</>;
```

client/src/containers/App.tsx
```diff
@@ -85,83 +85,126 @@ const WhoOwnsWhatRoutes: React.FC<{}> = () => {
...
+                <div className="dropdown dropdown-right show-lg">
+                  <button
+                    tabIndex={0}
+                    className={
+                      "btn btn-link dropdown-toggle m-2" + (isDropdownVisible ? " active" : "")
+                    }
+                    onClick={() => toggleDropdown()}
+                  >
+                    <i className={"icon " + (isDropdownVisible ? "icon-cross" : "icon-menu")}></i>
+                  </button>
```
toolness, line 167: You might want to add `aria-label="menu"` and set the `aria-expanded` attribute to `"true"` or `"false"` here depending on whether the menu is collapsed or not.  This will ensure that on screen readers, focusing on the hamburger will say "menu" (instead of "button", which is currently the case on Edge + Narrator), and it will also tell the screenreader user whether the menu is currently open or closed.  See the [tenants2 code](https://github.com/JustFixNYC/tenants2/blob/66ebc930237bec3c687649864c5381cce4c269fe/frontend/lib/ui/navbar.tsx#L172-L173) for an example.

toolness (approved):
This seems fine for users with a mouse, but I think there are issues with keyboard and screen reader support--see my comments for more details!

toolness:
Also I highly recommend trying out the hamburgler on Voiceover to see how it feels for you!

client/src/containers/App.tsx
sraby: hmmm so the issue here is that I want to be able to iterate over these link elements in a `map()` function later, so I can place them within menu item components. Is it possible to do that with a react fragment?

client/src/components/Modal.tsx
```diff
@@ -20,6 +20,7 @@ const Modal = (props: ModalProps) => {
       right: 0,
       bottom: 0,
       backgroundColor: "rgba(255, 255, 255, 0.75)",
+      //backgroundColor: "rgba(69, 77, 93, 0.6)", // rgba equivalent of the $dark color (#454D5D)
```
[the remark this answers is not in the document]
sraby: Waiting on Tahnee to decide which background color she wants, then will only be keeping one of these!

client/src/containers/App.tsx
toolness: Oh I didn't realize you wanted to iterate over them--yeah if you want to iterate over them, keep this as-is!

sraby:
> Hmm, so one funky thing about this right now is that keyboard focus is a bit wonky...
> 

So @toolness I think I figured out a workaround for this! I set an `onFocus` event that closes the dropdown menu when either the App body comes into focus, or any of the elements before the nav on the header. So, basically, if you tab down past the last element or the menu, or you reverse up past the menu close button, that should trigger the menu to close. Let me know what you think about this solution.

toolness:
Nice, this is much better!  But now the only problem is that the main `<div class="App__body">` appears to be keyboard-focusable, which isn't great because (A) it's not an interactive element ([only interactive elements should be keyboard-focusable](https://webaim.org/techniques/keyboard/#:~:text=Items%20that%20should%20not%20receive%20keyboard%20focus)) and (B) when it's focused there's no focus outline, so it looks like focus has disappeared, which would be very confusing for users.

Rather than attaching event listeners on either side of the hamburger, you might consider adding a document-level event handler for the `focus` event, which catches any change in focus that bubbles up the DOM, and then uses [`Node.contains()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) to see if it's within the hamburger menu... that's just one way, though.

Another might be to use [focus-trap-react](https://github.com/focus-trap/focus-trap-react#usage), which provides a nice component that you can use.  Um, if you have trouble getting TypeScript typings for it, you can actually potentially use [@justfixnyc/react-aria-modal](https://github.com/JustFixNYC/justfix-ts/tree/master/packages/react-aria-modal), where I [forked it](https://github.com/JustFixNYC/justfix-ts/blob/master/packages/react-aria-modal/src/focus-trap-react.ts) a while back...

sraby:
omg @toolness focus-trap-react comes with typings and is great!! I just used it and it seems to do exactly what we want to do.

toolness:
Oh nice!  This is looking much better... just two things:

1. It seems that `<div class="App__body">` is still keyboard-focusable, which is unfortunate because it makes focus disappear when the user is on the hamburger menu (or "ES" if on desktop) and presses <kbd>Tab</kbd>.  I had thought the `onFocus` handler was what made it keyboard-focusable but perhaps it was something else?

   Oh wait... I'm using the production site now and it looks like it has the same problem, so this isn't a bug introduced by your PR.  We should file a bug for it to fix it at some point though, because it's quite a bummer.

2. I notice that when I press <kbd>Esc</kbd>, the focus trap seems to disable itself, but the hamburger doesn't actually collapse.  This means that once I press that key, everything behaves like it does in the movie I posted above.  Is there some easy way to make <kbd>Esc</kbd> collapse the hamburger?  Or alternatively, make the focus trap not disable itself when the user presses that key?  (Though I recommend the former, since it's a common way to dismiss a modal, which is effectively what the hamburger is.)

sraby:
@toolness just got the "ESC" key to work with closing the menu. Will file a bug once it's merged for the first point. Lmk if there's anything else!

toolness:
Nice, this is splendid!!  Feel free to merge away.

Also, I just noticed that the disappearing focus issue only appears to be happening on my Firefox (and I'm on Windows, dunno if that matters).  When I use Chrome, it works fine--it's like that `<div class="App__body">` is only keyboard-focusable on Firefox, which is weird...

PR #497, opened 2021-06-24
Remove widows from title link on navbar

sraby:
This PR makes use of the [widont package](https://www.npmjs.com/package/widont) (that we use on the justfix-website) to remove any single-word second lines from the title link on the header nav bar. This problem specifically arose with the _Spanish_ title as it was too long to fit on one line. Now, instead of looking awkward with a trailing "York" on the second line, it looks much better: 

image


toolness (approved):
noice

PR #499, opened 2021-06-24
Timeline page v2 updates

sraby:


sraby:
These additions have been incorporated in #500, so closing this!

PR #500, opened 2021-06-24
Add DOB datasets to the timeline tab

sraby:
This PR adds two new datasets to our `indicatorhistory` api call—DOB and ECB Violations—which subsequently adds this data as options on the Timeline Tab of Who Owns What. While this data comes from two sources, we encapsulate both datasets within one grouping called "DOB Violations" which includes both regular DOB Violations and ECB Violations. This PR is also a good illustration of all the moving parts that need to be adjusted or implemented in order to bring new datasets to the Timeline Tab in general, which highlights that the process of adding these is still quite complicated! 

Still TODO:
- [x] Rename json fields to be more descriptive (I.e. "violations" instead of "viols") but keep old fields too for a month until user's cache's have reset
- [x] Consider merging both DOB and ECB violations
- [x] Look into if there is a lag for violation data in the new year with DOB NOW
- [x] Rename dataset "DOB/ECB Violations"
- [x] Consider just breaking it up into DOB Violations and ECB 
- [x] Rework description for dataset


client/src/components/Indicators.tsx
```diff
@@ -218,21 +218,14 @@ class IndicatorsWithoutI18n extends Component<IndicatorsProps, IndicatorsState>
...
-                      id="viols"
-                      activeId={activeVis}
-                      onChange={this.handleVisChange}
-                    />
-                    <IndicatorsDatasetRadio
-                      id="permits"
-                      activeId={activeVis}
-                      onChange={this.handleVisChange}
-                    />
+                    {Object.keys(INDICATORS_DATASETS).map((datasetKey, i) => (
```
sraby, line 221: Hrmmmm I couldn't think of a better way to map over all the possible values of the `IndicatorsDatasetId` type... any suggestions?

client/src/components/IndicatorsDatasets.tsx
```diff
@@ -69,68 +114,101 @@ export const INDICATORS_DATASETS: IndicatorsDatasetMap = {
...
+          rel="noopener noreferrer"
+        >
+          official NYC Buildings page
+        </a>
+        .
+      </Trans>
+    ),
+  },
+  dobviols: {
+    // TODO: Update This
```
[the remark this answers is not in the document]
sraby: Oooo! Thanks for catching. So I actually DID update this with the text I need, just didn't remove the "TODO"

client/src/components/Indicators.tsx
sraby: that's awesome, thank you @samaratrilling! I think I prefer the solution listed in there tbh

samaratrilling (approved):
I think this looks really good! Thank you so much!
Also, now that Atul has added a git SHA to our Amplitude user profile, we should able to use that to tell how many users are on which version of our code and therefore when you can change that function!

PR #501, opened 2021-06-25
Only load FullStory if users view the Timeline tab.

toolness:
This wraps FullStory's script injection in an idempotent global function called `justfix_startFullStoryRecording()`.  Our front-end code can call this anywhere it wants, as a way to limit our use of FullStory only to interactions that we're interested in (otherwise we deplete our monthly allowance of sessions very quickly).

Right now this PR only activates FullStory when the timeline tab is visited, because I think this was the event of interest mentioned in a recent meeting, but it's easy to change.


client/src/components/Indicators.tsx
```diff
@@ -190,6 +190,8 @@ class IndicatorsWithoutI18n extends Component<IndicatorsProps, IndicatorsState>
 
       const datasetDescription = INDICATORS_DATASETS[this.state.activeVis];
 
+      justfix_startFullStoryRecording();
```
toolness, line 193: This is being called in `render()`, which I don't like, because it's supposed to be a "pure" function that has no side effects.  But when I put it in `componentDidMount()`, it got called even when the user wasn't actually on the timeline tab... Anyways, since this just starts recording and isn't particularly related to the lifecycle of the component, I think it's okay to just leave it in `render()`, but whatevs.

If we want to get really fancy we could make a separate `<StartFullStory/>` component that calls `justfix_startFullStoryRecording()` on mount, but that might be going overboard.

kiwansim:
@austensen thinks we stopped using fullstory altogether

PR #502, opened 2021-06-29
Enhance Portfolio Tab experience for mobile users

sraby:
This PR makes various front-end improvements to the Portfolio Tab, with the main intention of improving our mobile experience. These include:
- Fix in place the first "Address" column
- Decrease width of Address column to ~90% (130px)
- Wrap text of Address entries onto 2+ lines
- Add Up down arrows in headers to indicate that data can be re-ordered
- Include fade out rectangle on viewport edge for horizontal scroll (and remove it when reaching end of scroll)
- Make all column headers fully shown 
- Make all columns only resizable on desktop

UPDATE: This PR now includes some tooling to set the height of the first column header as fixed to avoid some very distracting jank on mobile devices.

### DIFF
--- a/client/src/components/PropertiesList.tsx
+++ b/client/src/components/PropertiesList.tsx
@@ -63,9 +63,10 @@ const PropertiesListWithoutI18n: React.FC<
     <div className={classnames("PropertiesList", isLastColumnVisible && "last-column-visible")}>
       <ReactTableFixedColumns
         data={addrs}
-        minRows={Browser.isMobile() ? 5 : 10}
+        minRows={10}
         defaultPageSize={addrs.length}
         showPagination={false}
+        resizable={!Browser.isMobile()}
         columns={[
           {
             fixed: "left",
@@ -81,7 +82,6 @@ const PropertiesListWithoutI18n: React.FC<
                 id: "address",
                 width: 130,
                 fixed: "left",
-                resizable: false,
                 style: {
                   textAlign: "left",
                   whiteSpace: "unset",
--- a/client/src/components/PropertiesList.tsx
+++ b/client/src/components/PropertiesList.tsx
@@ -17,6 +17,8 @@ import Helpers, { longDateOptions } from "../util/helpers";
 import { AddressRecord, HpdComplaintCount } from "./APIDataTypes";
 import { withMachineInStateProps } from "state-machine";
 import { AddressPageRoutes } from "routes";
+import { createRef } from "react";
+import classnames from "classnames";
 
 export const isPartOfGroupSale = (saleId: string, addrs: AddressRecord[]) => {
   const addrsWithMatchingSale = addrs.filter((addr) => addr.lastsaleacrisid === saleId);
@@ -54,8 +56,11 @@ const PropertiesListWithoutI18n: React.FC<
   const addrs = props.state.context.portfolioData.assocAddrs;
   const rsunitslatestyear = props.state.context.portfolioData.searchAddr.rsunitslatestyear;
 
+  const lastColumnRef = createRef<any>();
+  const isLastColumnVisible = Helpers.useOnScreen(lastColumnRef);
+
   return (
-    <div className="PropertiesList">
+    <div className={classnames("PropertiesList", isLastColumnVisible && "last-column-visible")}>
       <ReactTableFixedColumns
         data={addrs}
         minRows={Browser.isMobile() ? 5 : 10}
@@ -368,7 +373,11 @@ const PropertiesListWithoutI18n: React.FC<
             ],
           },
           {
-            Header: i18n._(t`View detail`),
+            Header: (
+              <div ref={lastColumnRef}>
+                <Trans>View detail</Trans>
+              </div>
+            ),
             accessor: (d) => d.bbl,
             columns: [
               {
--- a/client/src/components/PropertiesList.tsx
+++ b/client/src/components/PropertiesList.tsx
@@ -268,7 +268,7 @@ const PropertiesListWithoutI18n: React.FC<
                 ),
                 accessor: (d) => d.openviolations,
                 id: "openviolations",
-                width: getWidthFromLabel(i18n._(t`Open`), 80),
+                width: getWidthFromLabel(i18n._(t`Open`), 85),
               },
               {
                 Header: (
@@ -279,7 +279,7 @@ const PropertiesListWithoutI18n: React.FC<
                 ),
                 accessor: (d) => d.totalviolations,
                 id: "totalviolations",
-                width: getWidthFromLabel(i18n._(t`Total`), 80),
+                width: getWidthFromLabel(i18n._(t`Total`), 85),
               },
             ],
           },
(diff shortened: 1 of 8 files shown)


client/src/components/PropertiesList.tsx
```diff
@@ -34,39 +53,68 @@ const PropertiesListWithoutI18n: React.FC<
...
+                Header: (
+                  <>
+                    <Trans>Zipcode</Trans>
+                    <ArrowIcon />
+                  </>
+                ),
                 accessor: (d) => d.zip,
                 id: "zip",
-                width: 75,
+                width: i18n.language === "es" ? 125 : 85,
```
toolness, line 104: I am curious if these kinds of conditionals would be better expressed as e.g. some kind of type, e.g.:

```typescript
type LocaliedLabelWidths = {
  zipcode: number,
  borough: number,
  /* Et cetera... */
};
```

then the localized metrics could just be passed into this function and the `width` could just be referred to by `widths.zipcodeWidth` etc, which would make it really easy to add more locales in the future... Hmm.  I also wonder if the width could just be a function of the actual number of characters in the localized string... anyhoo, just some thoughts.

sraby: @toolness I tried doing your latter recommendation of making width a function of label length! check it out

toolness:
Cool!  One thing I'm noticing as I use this on Firefox, Chrome, and IE11 on Windows is that there seem to be _two_ horizontal scrollbars at the bottom of the page; the bottom one doesn't actually have any scrollable area, so it seems like it should be hidden somehow.  I'm guessing you're not seeing this on Mac because of how Macs hide their scrollbars by default.

> image

toolness:
Oh weird, lol, I guess this PR is actually an improvement over the current production deployment, where there are _three_ scrollbars at the bottom of the page:

> image

So, uh, consider my above comment a non-blocker I guess.

client/src/util/helpers.ts
```diff
@@ -305,4 +306,26 @@ export default {
...
   },
+
+  /**
+   * Detects whether a given DOM element is visible on screen.
+   * Borrowed from https://stackoverflow.com/questions/45514676/react-check-if-element-is-visible-in-dom
+   */
+  useOnScreen(ref: React.RefObject<any>) {
+    const [isIntersecting, setIntersecting] = useState(false);
+
+    const observer = new IntersectionObserver(([entry]) => setIntersecting(entry.isIntersecting));
```
toolness, line 317: Hmm, I'm noticing that the current portfolio tab seems to work (mostly) on IE11 but this PR makes the portfolio tab crash on this line, because IE11 doesn't support `IntersectionObsever`.  In general, [older browsers from 2018 don't support it](https://caniuse.com/intersectionobserver) so it might be worth using a progressive enhancement strategy here, e.g. checking to see if `typeof IntersectionObserver === "undefined"` and returning some kind of default value if it's unsupported.

toolness (approved):
Nice, this looks great! Tried it on all my Windows-based browsers and Chrome on Android and it's nice.  The new `getWidthFromLabel()` function is great too!

Just one comment regarding your use of `IntersectionObserver` below but that's it.

client/src/util/helpers.ts
sraby: Good catch! Will add this in now, and adjust the description of the function to match.

sraby:
[ch7486]

client/src/styles/PropertiesList.scss
```diff
@@ -2,16 +2,53 @@
 @import "_scrollbar.scss";
 @import "_button.scss";
 
+$table-border-light: 1px solid $gray-light;
+$table-border-dark: 1px solid $dark;
+
 .PropertiesList {
   height: 100%;
   width: 100%;
+  // This makes sure that the fade effect from the "::after" component
```
samaratrilling, line 11: Can you clarify that it's the next two items, not just one?

client/src/styles/PropertiesList.scss
```diff
@@ -2,16 +2,53 @@
...
+        transform: translateY(0.5px);
+      }
+    }
+
+    .rt-thead {
+      max-height: 30px;
+      .rt-tr .rthfc-th-fixed.rthfc-th-fixed-left {
+        position: fixed;
+        // This display prop makes sure that the fixed table cells
+        // fill up their entire alloted spaces
```
samaratrilling, line 36: Tiny nit but `alloted` -> `allotted`

client/src/util/helpers.ts
```diff
@@ -305,4 +306,67 @@ export default {
...
+
+  /**
+   * Detects whether a given DOM element is visible on screen.
+   *
+   * Note: for older browsers that do not support IntersectionObserver, this
+   * hook will always return FALSE by default.
+   *
+   * Borrowed from https://stackoverflow.com/questions/45514676/react-check-if-element-is-visible-in-dom
+   */
+  useOnScreen(ref: React.RefObject<any>) {
```
samaratrilling, line 318: should this be `isOnScreen` since it returns a boolean? wait, does it return a boolean?

samaratrilling (approved):

client/src/util/helpers.ts
toolness: oh i think it might be a [React hook](https://reactjs.org/docs/hooks-intro.html), those always start with `use` by convention...

sraby: yes ditto what @toolness said, I remember being a bit confused by that too

client/src/styles/PropertiesList.scss
sraby: heheh ty

client/src/components/PropertiesList.tsx
```diff
@@ -30,76 +67,174 @@ const PropertiesListWithoutI18n: React.FC<
...
+   * Let's hide the fade out on the right edge of the table if:
+   * - We've scrolled to the last column OR
+   * - We're using an older browser that cannot detect where we've scrolled
+   */
+  const hideScrollFade = isOlderBrowser || isLastColumnVisible;
+
+  const tableRef = useRef<HTMLDivElement>(null);
+  const isTableVisible = Helpers.useOnScreen(tableRef);
+
+  const [headerTopSpacing, setHeaderTopSpacing] = useState<number | undefined>();
```
toolness, line 93: Maybe add a comment here explaining why we're tracking this, and why it's important, from an end-user point of view?

client/src/util/helpers.ts
```diff
@@ -82,6 +83,27 @@ export const longDateOptions = { year: "numeric", month: "short", day: "numeric"
 export const mediumDateOptions = { year: "numeric", month: "long" };
 export const shortDateOptions = { month: "short" };
 
+/**
+ * Delay the action of a certian function by a set amount of time.
+ *
+ * Originally copied from:
+ * https://gist.github.com/gragland/4e3d9b1c934a18dc76f585350f97e321#gistcomment-3073492
+ */
+const debounce = (delay: number, fn: any) => {
```
sraby, line 92: btw @toolness I tried using the react hook version of this function that you implemented in tenants2, but given that I wanted to use it inside the `useWindowSize` hook function below, it ended up causing some rule-of-hooks type errors. I reverted back to this implementation.

toolness (approved):
Hmm so code-wise this looks good, but FYI while things are a bit more responsive on IE11 on my system, resizing the window still feels very laggy, which is a bummer...

PR #504, opened 2021-07-12
Upgrade Lingui to most recent minor release

sraby:
This fixes an issue where running `yarn lingui extract` crashes because it cannot parse through a object chaining statement (ex: `BABEL_PARSE_ERROR` logged on trying to parse `tableRef?.current?.offsetTop`).


sraby:
this is a super small package upgrade so im gonna merge meself

PR #506, opened 2021-07-13
Attempt to memoize the portfolio table.

toolness:
This memoizes the portfolio table via [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) in an attempt to improve performance, particularly on IE11.

As far as I can tell via IE11's profiler, the majority of delay is now caused by layout:

> image

I'm honestly not even sure how much it helps to make the table a pure component, though I guess it can't really hurt.  I suppose it does make the code a bit more complex, though.

### DIFF
--- a/client/src/components/PropertiesList.tsx
+++ b/client/src/components/PropertiesList.tsx
@@ -112,7 +112,7 @@ const PropertiesListWithoutI18n: React.FC<
       className={classnames("PropertiesList", hideScrollFade && "hide-scroll-fade")}
       ref={tableRef}
     >
-      <TheTable
+      <TableOfData
         addrs={addrs}
         headerTopSpacing={headerTopSpacing}
         i18n={i18n}
@@ -136,11 +136,15 @@ type TheTableProps = {
   addressPageRoutes: AddressPageRoutes;
 };
 
-const TheTable = React.memo(
+/**
+ * This component memoizes the portfolio table via React.memo
+ * in an attempt to improve performance, particularly on IE11.
+ */
+const TableOfData = React.memo(
   React.forwardRef<HTMLDivElement, TheTableProps>((props, lastColumnRef) => {
     const { addrs, headerTopSpacing, i18n, locale, rsunitslatestyear } = props;
 
-    console.log("Rendering <TheTable>", headerTopSpacing);
+    console.log("Rendering <TableOfData>", headerTopSpacing);
 
     return (
       <ReactTableFixedColumns
--- a/client/src/components/PropertiesList.tsx
+++ b/client/src/components/PropertiesList.tsx
@@ -126,7 +126,7 @@ const PropertiesListWithoutI18n: React.FC<
   );
 };
 
-type TheTableProps = {
+type TableOfDataProps = {
   addrs: AddressRecord[];
   headerTopSpacing: number | undefined;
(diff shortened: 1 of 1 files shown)


toolness:
Hmm, @sraby, once I saw that most of the time spent during resizing is on layout, I tried removing the `PropertiesList` class from the React table and resizing was _much_ more responsive on IE11, but the table lost its scrollbars.  I tried various things to make the scrollbars come back, but none of them seemed to fully work (at one point I had a vertical scrollbar that was somehow partially obscured by the table content).

I suspect there's a way to have either IE11-specific CSS, or universal CSS that's more IE11-friendly, to make this easier on IE11 users.  But given that a tiny fraction of our users are on IE11 and the browser [reaches end-of-life in under a year](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge#what-is-the-lifecycle-policy-for-internet-explorer-), I'm not sure how much it's worth, unless perhaps fixing the layout issues on IE11 happens to benefit other browsers/devices in ways that we don't know about.

client/src/components/PropertiesList.tsx
```diff
@@ -73,7 +72,7 @@ const PropertiesListWithoutI18n: React.FC<
   const addrs = props.state.context.portfolioData.assocAddrs;
   const rsunitslatestyear = props.state.context.portfolioData.searchAddr.rsunitslatestyear;
 
-  const lastColumnRef = createRef<any>();
+  const lastColumnRef = useRef<HTMLDivElement>(null);
```
toolness, line 75: This is the one thing I found that's potentially a bug.  `createRef` is really for class-based components, and will return a brand-new reference every time it's called, whereas `useRef` will return the same one every render.

sraby:
Thank you so much for your work on this @toolness! To be clear, are you recommending that I just fix the one bug that you found and NOT merge this PR in as it adds unnecessary complexity?

toolness:
> Thank you so much for your work on this @toolness! To be clear, are you recommending that I just fix the one bug that you found and NOT merge this PR in as it adds unnecessary complexity?

Sure thing!  Hmm, it is up to you really.  If you think this PR doesn't add too much complexity, there is a possibility it might make things a bit faster.  Otherwise you can just fix the one bug I found.

sraby:
ok sounds good @toolness! In that case, I'll merge it once the tests pass. I added a code comment and changed the component name for clarity.

toolness:
Cool thanks!

I also wanted to note one thing that motivated me to memoized the table component: when I originally profiled the performance of the portfolio tab's rendering, a fair amount of it _was_ on the JS side.  But it turned out that this was because I was profiling our development build, and an inordinate amount of time was actually being spent on our I18n translations (lingui stuff).  However, I believe this was because lingui actually behaves very differently during development rather than production: during development it does a lot of parsing on-the-fly to make development easier, but it also makes things run slower.  In production, though, it uses pre-compiled stuff that was generated at build time, so that extra cost isn't present.  I verified this by profiling our production builds/netlify preview deploys and indeed the JS ran much faster, with negligible time being spent on I18n.

Anyways, I thought I'd mention that just in case you end up profiling stuff yourself in the future: development JS doesn't necessarily have the same performance characteristics as production JS!

PR #517, opened 2021-08-03
Show full list of useful links on Not Registered pages

sraby:
This PR is a restructuring of the UI of our "Not Registered" page. Specifically, it replaces the buttons under the "Useful Links" header with our `<UsefulLinks>` component. This is a response to a request from the Program team about having the HPD Building Profile link available, even for non-registered properties.

Note that this PR also fixes a minor bug where we were not including a proper google event tag with these outbound useful links on the Not Registered pages.

### DIFF
--- a/client/src/containers/NotRegisteredPage.tsx
+++ b/client/src/containers/NotRegisteredPage.tsx
@@ -7,10 +7,11 @@ import { AddressPageUrlParams } from "../routes";
 import Modal from "../components/Modal";
 import LegalFooter from "../components/LegalFooter";
 import Helpers from "../util/helpers";
-import { SocialShareAddressPage } from "../components/SocialShare";
+import SocialShare, { SocialShareAddressPage } from "../components/SocialShare";
 import { Nobr } from "../components/Nobr";
 import { withMachineInStateProps } from "state-machine";
 import Page from "components/Page";
+import { UsefulLinks } from "components/UsefulLinks";
 
 type Props = withMachineInStateProps<"unregisteredFound">;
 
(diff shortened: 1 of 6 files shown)


sraby:
[ch7712]

samaratrilling:
Looks ok to me! Can you add a screenshot to the description of what it looks like now?

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -42,19 +42,20 @@ export default class NotRegisteredPage extends Component<Props, State> {
...
-          streetname: searchAddrParams.streetname,
-        }
-      : buildingInfo
-      ? {
-          boro: buildingInfo.boro,
-          housenumber: buildingInfo.housenumber || " ",
-          streetname: buildingInfo.streetname || " ",
-        }
-      : null;
+    const usersInputAddress =
```
samaratrilling, line 45: This is still a little confusing to me, but I guess it's just if you have both `searchAddrParams` and `searchAddrBbl` use them, otherwise if you have building info use that.

What happens if you don't have `searchAddrBbl` but you do have `buildingInfo`? Should the `: buildingInfo` bit really be `:buildingInfo && searchAddrBbl`?

samaratrilling: Or will that never happen?

samaratrilling (approved):
Looks ok, with one question. Great job!

sraby:
@samaratrilling here is what it looks like (you can get to this page by searching `75 HENDRIX` street on the deploy preview link) 
image
Screen Shot 2021-08-04 at 1 38 26 PM

client/src/containers/NotRegisteredPage.tsx
sraby: Oh good point! yeah this logic is unnecessarily confusing. The idea is that if the user didn't type in the address themselves from the homepage (like if they redirect from a BBL-based url), we will need to rely on the "buildingInfo" query to grab an address for the property in question as a fallback.

sraby: I simplified the logic and also added a comment to explain

PR #519, opened 2021-08-10
Preliminary adding of amplitude to WOW

samaratrilling:
This PR sets up Amplitude tracking on WOW and adds a few custom events and property sets to the Feature Callout Widget component added in #391. Most notably, this PR sets a new client-side git-committed env variable called  `REACT_APP_AMPLITUDE_API_KEY` that can be overridden in production so we can have both a default dev api key present, as well as a production api key.


samaratrilling:
https://analytics.amplitude.com/justfixnyc/settings/projects

sraby:
[ch7243]

sraby (approved):

client/src/components/Amplitude.ts
```diff
@@ -0,0 +1,12 @@
+import amplitude from "amplitude-js";
+
+// Initiating Amplitude inside this helper file seems to work better with the Create React App framework than
+// adding a script tag to our index.html file.
+//
+// See https://javascript.plainenglish.io/adding-analytics-to-your-react-application-b584265f9fae for more details
```
samaratrilling, line 6: nice

client/src/components/FeatureCalloutWidget.tsx
```diff
@@ -71,7 +87,10 @@ const FeatureCalloutWidget = withI18n()((props: withI18nProps) => {
           focusTrapOptions={{
             clickOutsideDeactivates: true,
             returnFocusOnDeactivate: false,
-            onDeactivate: () => setWidgetVisibility(false),
+            onDeactivate: () => {
+              amplitude.getInstance().logEvent("closeFeatureCalloutWidget");
```
samaratrilling, line 91: Since we have more events now, I'm thinking about the best way to standardize this (e.g. a WOW-specific amplitude wrapper where you can just call `logEvent(eventType)` where `eventType` is some big list of strings). That way you won't accidentally pass a misspelled string and have 2 different event types going

samaratrilling (commented):
Looks good! See the note about standardization - I'm curious to hear your thoughts.

PR #523, opened 2021-08-16
Incorporate new WOWZA Portfolio mapping into the WOW back-end and front-end 

sraby:
This PR implements a new graph-based portfolio mapping between landords, business addresses and buildings. It builds off of the work implemented in #524 (and the accompanying https://github.com/JustFixNYC/nycdb-k8s-loader/pull/81) by pulling in data from the newly created `wow_portfolios` table as an alternative source of grouping properties together under shared ownership. 

Using the feature flag `REACT_APP_ENABLE_NEW_WOWZA_PORTFOLIO_MAPPING`, you can enable UI for the user to toggle between the old and this new method of portfolio mapping. The new method, in addition to finding a more robust set of associated buildings on the AddressPage, also shows an interactive network graph at the top of the summary tab.

NOTE: these changes are not quite ready for the public (for example, the Summary tab still shows statistics using the old portfolio method even if the "new" one is toggled), but committing this will allow us deploy these changes to our demo link for testing and research.


toolness:
Yeee when I look up 654 park place I get this dialog:

> image

It does _eventually_ load though.

The portfolio also appears to be 4,335 buildings, which is almost twice as large as any of the portfolios in [hpd-graph-fun](https://justfixnyc.github.io/hpd-graph-fun/)... hmm.  I think hpd-graph-fun culls out hpd registrations that expired more than 90 days ago, rather than 365 as #524 does things, but I am not sure if that setting made _that_ much of a difference.  Hmm.

client/src/components/APIDataTypes.ts
```diff
@@ -80,9 +80,29 @@ export type AddressRecord = {
...
+export type PortfolioGraphEdge = {
+  from: number;
+  to: number;
+  reg_contacts: number;
+};
+
+export type RawPortfolioJson = {
+  nodes: PortfolioGraphNode[];
+  edges: PortfolioGraphEdge[];
+};
```
toolness, line 100: I wonder if this should be called `RawPortfolioGraphJson`, just so it's not confused with the legacy JSON structure that describes portfolios?

Also, I updated this typing in 2cabd8d7b817f238da8586c977c0e3af977afa24 so it doesn't have any `any`'s anymore.

client/src/containers/AddressPage.tsx
```diff
@@ -38,6 +38,7 @@ type RouteState = {
 type AddressPageProps = RouteComponentProps<RouteParams, {}, RouteState> &
   withMachineProps & {
     currentTab: number;
+    requestNewPortfolioData?: boolean;
```
toolness, line 41: When I see `requestNewPortfolioData` my first thought is that it is used to clear some kind of cache, e.g. "don't use the old portfolio data you might have cached, request it anew so our data is fully up to date".  It took looking at the implementation to realize that it's about requesting the new wowza/graph structured format of portfolio data, rather than the legacy algorithm's.

I wonder if this should be called `requestPortfolioGraphData`, or `requestWowzaPortfolioData`, or `useNewPortfolioMethod` (as it is some other places in the code) or something else, to make it less ambiguous?

toolness (approved):
The code here looks fine, I just have concerns about the performance (which I commented on separately).  After doing a brief profiling of page load it looks like the vast majority of initial time (a good 3-4 seconds) is taken up just by Lingui's parsing (this is all done in a build step during production and skipped at runtime, which is part of why production runs so much faster than development).  So it might just be due to the fact that the portfolio is so much bigger than any of the ones in current WoW.

Anyways, the good news is that it doesn't seem like much of the time is spent on the server-side, which would be more of a concern for merging (since we don't want visitors on the WoW demo site to accidentally DoS it, even if it is just the demo site and not production).  So I am okay with merging this!

toolness:
Er also, even though the backend doesn't spend too much compute time on getting a portfolio graph, the graphs for the largest portfolios are _huge_:

```
"GET /api/address/wowza?block=01238&lot=0016&borough=3 HTTP/1.1" 200 8748138
```

So that's an **8 megabyte payload** for the 654 park place portfolio (which as mentioned earlier is abnormally large at 4k buildings).  Still, we might want to consider [gzipping it](https://docs.djangoproject.com/en/3.2/ref/middleware/#module-django.middleware.gzip) to save bandwidth.

client/src/data/sample-portfolio-graph.json
```diff
@@ -0,0 +1 @@
...:false,\"bbl\":\"1020340048\"},{\"from\":45060,\"to\":156,\"reg_contacts\":1,\"is_bridge\":false,\"bbl\":\"2030220001\"},{\"from\":45060,\"to\":200,\"reg_contacts\":1,\"is_bridge\":false,\"bbl\":\"2032370097\"},{\"from\":156,\"to\":199,\"reg_contacts\":46,\"is_bridge\":false,\"bbl\":\"1020250041\"},{\"from\":156,\"to\":3961,\"reg_contacts\":9,\"is_bridge\":false,\"bbl\":\"2022670011\"},{\"from\":156,\"to\":1933,\"reg_contacts\":57,\"is_bridge\":false,\"bbl\":\"5012430020\"},{\"from\":14783,\"to\":1933,\"reg_contacts\":2,\"is_bridge\":false,\"bbl\":\"2034990007\"},{\"from\":804,\"to\":45275,\"reg_contacts\":1,\"is_bridge\":false,\"bbl\":\"3031720019\"},{\"from\":200,\"to\":1933,\"reg_contacts\":7,\"is_bridge\":false,\"bbl\":\"2030680093\"},{\"from\":200,\"to\":199,\"reg_contacts\":6,\"is_bridge\":false,\"bbl\":\"2032600064\"},{\"from\":5137,\"to\":199,\"reg_contacts\":10,\"is_bridge\":false,\"bbl\":\"1018020040\"},{\"from\":22166,\"to\":816,\"reg_contacts\":6,\"is_bridge\":false,\"bbl\":\"1015220037\"},{\"from\":103770,\"to\":816,\"reg_contacts\":1,\"is_bridge\":false,\"bbl\":\"1013620037\"},{\"from\":816,\"to\":20187,\"reg_contacts\":3,\"is_bridge\":false,\"bbl\":\"1014400030\"}]}"
```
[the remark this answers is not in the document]
sraby: yes! good catch

samaratrilling:
After GZIPping, it's only 787 KB (down from 8 MB)
`[19/Aug/2021 12:46:18] "GET /api/address/wowza?block=01238&lot=0016&borough=3 HTTP/1.1" 200 787994`

client/src/components/AddressToolbar.tsx
```diff
@@ -1,81 +1,89 @@
...
+  const toggleUrlChange = () => {
+    if (isWowzaRoute) {
+      history.push(pathname.replace("/wowza", ""));
+    } else {
+      const locale = parseLocaleFromPath(pathname);
+      const newUrl = `/${locale}/wowza${removeLocalePrefix(pathname)}`;
+      history.push(newUrl);
+    }
+    history.go(0);
+  };
```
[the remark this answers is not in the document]
sraby: good point! just did that

client/src/components/APIDataTypes.ts
sraby: noice!

client/src/containers/AddressPage.tsx
sraby: yes! I like `useNewPortfolioMethod`

PR #524, opened 2021-08-17
Generate a table of portfolios using networkx

toolness:
This adds a new `wow_portfolios` table which contains disjoint portfolios that can be indexed by BBL.  It's generated whenever the `dbtool.py builddb` command is run.  Additionally, for development and debugging purposes, a new `dbtool.py exportgraph` command has been added, which outputs the rows of the table as a JSON file.

The new table's schema is defined in `sql/create_portfolios_table.sql`.  Each row is a separate portfolio; one column contains a list of BBLs in the portfolio, another contains a list of landlord names in it, and the last contains the JSON graph structure of the portfolio itself.

## What is the portfolio graph structure?

The portfolio graph is undirected and has the following properties:

1. Each individual name is a node.

2. Each business address is a node.

3. A name node and address node are connected via an edge if at least one HPD registration contact contains both (i.e., if the name is associated with the address).

Portfolios are then inferred by finding strongly connected components within the graph.

This algorithm was created by Sam Rabiyah, Samara Trilling, and Atul Varma during the week of July 26, 2021.

In practical terms, the JSON is structured similarly to the JSON graph portfolio structure of [hpd-graph-fun][], albeit without quite as much edge information--specifically, there's no `is_bridge` or `bbl` properties on edges at the moment.  Also, nodes are a tagged union rather than the weird thing Rust serialized them as: each node has a `kind` attribute that can be `name` or `bizaddr`, and a `value` attribute that is either a person's name or a business address.

[hpd-graph-fun]: https://github.com/JustFixNYC/hpd-graph-fun/blob/main/typescript/portfolio.d.ts

## Why networkx?

We wanted to use something Python-based, since the rest of our data pipeline is in Python.

We looked at two primary graphing libraries for Python: networkit and networkx.  We decided to go with networkx for the following reasons:

1. While networkit is fast (it's written in C++), networkx (which is pure Python) loads data at the scale we're using _fast enough_,

2. networkx is very flexible in some ways compared to networkit (for instance, networkit only supports floating-point numbers for edge weights while networkx lets you use whatever you want),

3. The documentation for networkx appears to be better (or at least, I had an easier time finding my way around it).  In particular, because it's all Python, it's very easy to jump to from an IDE (e.g. using VSCode's "go to definition" feature), read, and understand (it also has great docstrings).  In contrast, finding documentation and implementation for networkit was more challenging, partly due to the fact that it's implemented in C++ via Cython.

4. networkx does what we need (detection of connected components) without any dependencies. In contrast, networkit brings in numpy, scipy, the whole kitchen sink of python scientific computing, and also requires Cython and CMake.

While networkit's implementation of cutting-edge research algorithms is awesome (especially for finding central vertices and community detection), we'll likely use it for offline analysis of our graph structures rather than leveraging it in our data pipeline.  And if we ever really need to, we can always switch over to it--it shouldn't be too hard.


portfoliograph/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+        print("Building graph.")
+        cur.execute("""
+            SELECT * FROM hpd_contacts
+            WHERE
+              type = any('{HeadOfficer,IndividualOwner,CorporateOwner}') AND
+              businesshousenumber != '' AND
+              businessstreetname != '' AND
+              firstname != '' AND
+              lastname != ''
```
sraby, line 46: Just FYI @toolness, the SQL filtering we use for Worst Evictors (and that we used to generate test data at the last retreat) is a bit more inclusive than this logic. See below:
```sql
where 
  type = any('{HeadOfficer,IndividualOwner,CorporateOwner}')
  and (businesshousenumber is not null OR businessstreetname is not null)
  and length(concat(businesshousenumber,businessstreetname)) > 2
  and (firstname is not null OR lastname is not null)
```

toolness: Thanks! I've added a TODO in the code for this for now

dbtool.py
```diff
@@ -180,6 +185,10 @@ def run_sql_file(self, sqlpath: Path) -> None:
             with self.conn.cursor() as cursor:
                 cursor.execute(sql)
 
+    def populate_portfolios_table(self) -> None:
+        with self.conn:
+            populate_portfolios_table(self.conn)
```
toolness, line 190: This probably doesn't need to be its own separate method.

portfoliograph/__init__.py
toolness: Ok addressed this in 276015e!

portfoliograph/graph.py
```diff
@@ -0,0 +1,133 @@
...
+    kind: NodeKind
+    name: str
+
+    def to_json(self) -> Dict[str, Any]:
+        # This format is an artifact from the way hpd-graph-fun's
+        # nodes were serialized. For more details, see:
+        #
+        #   https://github.com/JustFixNYC/hpd-graph-fun
+        if self.kind == NodeKind.NAME:
+            return {"Name": self.name}
```
sraby, line 22: @toolness is there any reason you didn't add a "Type" property here so we can differentiate between names and bizaddr node types for easier visualization? And also make it easier to add new types

sraby (approved):
Approved by Sam + Samara! This looks awesome and is very clear and well documented.

portfoliograph/graph.py
toolness: Oh that's just because it's the default way a Rust library I used for hpd-graph-fun serialized enums.  I agree that it's not great--fixed in ad5ee4fd0cd6b5ec0b49abef301dfc26196924aa!

PR #526, opened 2021-08-23
Calculate aggregation of portfolio data without external API call

sraby:
This PR replaces the front-end use of our `/aggregate` API endpoint with a js function that calculates data for the Summary tab purely in the front-end and based on portfolio data already saved in client. This effectively removes the need for any external network request to render the Summary tab and all of it's aggregate calculations.

Notably, this change also involves updating our "state machine"— now that we are not making a network request on Summary tab page loading, there is no longer the concept of a network "error" state or network request "pending" state. Those states were removed.

Also, I went ahead and tracked how much time it took to run the new `calculateAggDataFromAddressList()` function. It was running in 1.5 milliseconds on my device, wahoo! 

TODO:

- [x] Add tests
- [x] Handle rounding and formatting on the front-end
- [x] Remove front-end elements of old API endpoint
- [x] Performance analysis (takes 1.5 milliseconds to run the function)


sraby:
[ch7807]

client/src/components/APIDataTypes.ts
```diff
@@ -107,11 +107,11 @@ export type SearchResults = {
 
 // TYPES ASSOCIATED WITH SUMMARY AGGREGATE QUERY:
```
samaratrilling, line 108: Does this comment still hold? Maybe `types associated with summary aggregation`, since it's not a query anymore?

client/src/components/PropertiesSummary.tsx
```diff
@@ -125,14 +122,16 @@ export default class PropertiesSummary extends Component<Props, {}> {
...
+                  <Trans>
+                    The <Plural value={agg.bldgs} one="" other="average" /> age of{" "}
+                    <Plural value={agg.bldgs} one="this building" other="these buildings" /> is{" "}
+                    <b>{Math.round(agg.age)}</b> years old.
+                  </Trans>
+                )}
               </p>
               <aside>
-                {agg.violationsaddr && (
+                {agg.violationsaddr.lat && agg.violationsaddr.lng && (
```
samaratrilling, line 134: Ooh, will this fix any of our errors?

client/src/components/SummaryCalculation.test.ts
```diff
@@ -0,0 +1,36 @@
+import { SAMPLE_ADDRESS_RECORDS } from "state-machine-sample-data";
+import { calculateAggDataFromAddressList, extractLocationDataFromAddr } from "./SummaryCalculation";
+
+describe("extractLocationDataFromAddr()", () => {
+  it("works", () => {
```
samaratrilling, line 5: lol mayyyybe a little more specificity here would be cool - like `it('gets all 5 fields from address')` or something

client/src/components/SummaryCalculation.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  it("calculates sums with filtering", () => {
+    expect(agg.totalrsloss).toBe(-8);
+  });
+  it("calculates proportions", () => {
+    expect(agg.openviolationsperresunit).toBeCloseTo(0.05);
+  });
+  it("is able to find top addresses", () => {
+    expect(agg.violationsaddr.streetname).toBe("LEWIS AVENUE");
+  });
+  it("is able to find most common names, in order", () => {
```
samaratrilling, line 33: Nice, loving these tests

client/src/components/SummaryCalculation.ts
```diff
@@ -0,0 +1,83 @@
...
+  }))({
+    ...addr,
+  });
+};
+
+export const calculateAggDataFromAddressList = (addrs: AddressRecord[]): SummaryStatsRecord => {
+  const bldgs = addrs.length;
+  const units = _.sumBy(addrs, (a) => a.unitsres || 0);
+
+  const totalopenviolations = _.sumBy(addrs, (a) => a.openviolations);
```
samaratrilling, line 29: should these be in camelCase?

client/src/components/SummaryCalculation.ts
```diff
@@ -0,0 +1,83 @@
...
+  const totalviolations = _.sumBy(addrs, (a) => a.totalviolations);
+
+  const totalevictions = _.sumBy(addrs, (a) => a.evictions || 0);
+
+  const totalrsgain = _.sumBy(addrs, (a) => (a.rsdiff && a.rsdiff > 0 ? a.rsdiff : 0));
```
samaratrilling, line 34: do you need the `&& a.rsdiff > 0`? i.e. are there any cases where this number is negative? The cases are:
- rs is null (in which case you want to reset to 0)
- rs is 0 (in which case who cares if it gets summed, since it's just 0)
- rs > 0 (in which you want to sum it)
- rs < 0 (in which case I can imagine you wanting to explicitly reset it to 0, but does that happen?)

client/src/components/SummaryCalculation.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      streetname: "PARK PLACE",
+      boro: "BROOKLYN",
+      lat: 40.6737974139504,
+      lng: -73.9562781322538,
+    });
+  });
+});
+
+describe("calculateAggDataFromAddressList()", () => {
+  const agg = calculateAggDataFromAddressList(SAMPLE_ADDRESS_RECORDS);
```
samaratrilling, line 17: Are there any other tests to add here for when certain other values are null and you want to make sure they get summarized appropriately? Like rsloss?

client/src/util/helpers.ts
```diff
@@ -168,6 +169,14 @@ export default {
     return max;
   },
 
+  getMostCommonElementsInArray(array: string[], numberOfResults: number): string[] {
+    const elementsByFrequency = _.countBy(array);
+    const sortedElementsByFrequency = Object.entries(elementsByFrequency).sort(
+      (a, b) => b[1] - a[1]
+    );
+    return sortedElementsByFrequency.slice(0, numberOfResults).map((a) => a[0]);
```
samaratrilling, line 177: Can you explain to me what the map() is doing here?

client/src/components/SummaryCalculation.ts
sraby: hey @samaratrilling, so there are many cases where rsdiff is negative, specifically when there is a drop in rs units from 2007 to 2019 for a given building. The idea behind my function is that we want to filter out the negative values so we can see just a tally of how many units were gained across the portfolio. I can add a comment here to make it more clear-does that make sense?

client/src/components/PropertiesSummary.tsx
sraby: I hope so! I just noticed it and was like ooof

client/src/components/SummaryCalculation.ts
sraby: hmmm so they normally would but since our API responses give their fields in lower case like this (bc I think the SQL query result has them like this), I think it would be good to keep things consistent with the other big objects of data we house in APIDataTypes. Does that make sense?

client/src/util/helpers.ts
sraby: Yes, so the `sortedElementsByFrequency` const is an array of string + number pairs with the string being the element value (like landlord name) and the number being the frequency. Since the overall helper function just wants to return a simple array, this last map is essentially deleting the record of the frequency and just keeping the element values, in order. Does that make sense?

sraby: Gonna add a comment.

client/src/components/SummaryCalculation.test.ts
sraby: Hmmmm I can't think of any right now. Do you have one in mind?

sraby:
@samaratrilling I'm gonna go ahead and merge this as it seems like I was able to resolve all of your comments. Let me know if this is not the case and I can create a new PR with changes!

samaratrilling:
Ok with me!

PR #527, opened 2021-08-23
Fix Node Version error on CircleCI

sraby:


sraby:
This was a cluster*$#% of a time trying to upgrade Node which ultimately left us in dependency hell. Other much more elegant solution found at #528 !

PR #530, opened 2021-08-24
Fix incorrect lingui import on WowzaToggle

sraby:
This PR fixes a bug where the "Switch to New/Old Version" button on the Address Toolbar wasn't showing any text within it's <Trans> tags. Turns out it was because we were importing the Trans component from lingui/react and NOT lingui/macro which is where it should come from.


sraby:
@samaratrilling this was a super quick bug fix so I merged this on my own, but just wanted to flag this issue that importing the `<Trans>` component from  `@lingui/react` instead of `@lingui/macro` creates weird issues like this and should be avoided at all costs! Might even be worth making a VSCode custom linter rule to check for this...

samaratrilling:
Ooh, good point! Thanks for this!

PR #536, opened 2021-09-20
Add 311 complaints section to summary page

sraby:
This PR adds a new section to the Summary tab called "Complaints called in to 311" on Who Owns What that shows users a sum of 311 complaints across the portfolio, as well as a list of the most common complaint types.

### DIFF
--- a/client/src/components/APIDataTypes.ts
+++ b/client/src/components/APIDataTypes.ts
@@ -139,6 +139,9 @@ export type SummaryStatsRecord = {
   topbusinessaddr: string | null;
   topcorp: string | null;
   topowners: string[];
+  totalcomplaints: number;
+  totalrecentcomplaints: number;
+  recentcomplaintsbytype: HpdComplaintCount[] | null;
   totalevictions: number | null;
   totalopenviolations: number;
   totalrsdiff: number | null;
--- a/client/src/components/APIDataTypes.ts
+++ b/client/src/components/APIDataTypes.ts
@@ -141,7 +141,7 @@ export type SummaryStatsRecord = {
   topowners: string[];
   totalcomplaints: number;
   totalrecentcomplaints: number;
-  recentcomplaintsbytype: HpdComplaintCount[] | null;
+  recentcomplaintsbytype: HpdComplaintCount[];
   totalevictions: number | null;
   totalopenviolations: number;
   totalrsdiff: number | null;
--- a/client/src/components/APIDataTypes.ts
+++ b/client/src/components/APIDataTypes.ts
@@ -139,9 +139,9 @@ export type SummaryStatsRecord = {
   topbusinessaddr: string | null;
   topcorp: string | null;
   topowners: string[];
-  totalcomplaints: number;
-  totalrecentcomplaints: number;
-  recentcomplaintsbytype: HpdComplaintCount[];
+  totalhpdcomplaints: number;
+  totalrecenthpdcomplaints: number;
+  recenthpdcomplaintsbytype: HpdComplaintCount[];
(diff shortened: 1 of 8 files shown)


client/src/components/SummaryCalculation.ts
```diff
@@ -10,6 +15,34 @@ export const getTopFiveContactsInPortfolio = (addrs: AddressRecord[]) => {
   return helpers.getMostCommonElementsInArray(allContactNames, 5);
 };
 
+export const NUM_COMPLAINT_TYPES_TO_SHOW = 3;
+
+export const getTopComplaintTypesInPortfolio = (addrs: AddressRecord[]) => {
+  // Generate array of alll HpdComplaintCount objects across entire portfolio
```
samaratrilling, line 21: nit: extra L

client/src/components/APIDataTypes.ts
```diff
@@ -139,6 +139,9 @@ export type SummaryStatsRecord = {
   topbusinessaddr: string | null;
   topcorp: string | null;
   topowners: string[];
+  totalcomplaints: number;
```
samaratrilling, line 142: any chance you want to call this total311complaints, like we did for hpd violations? if there are any other types of complaints it might be harder to refactor later

samaratrilling: Or will this number continue to be the absolute total even if we added other types of complaints?

client/src/locales/en/messages.po
```diff
@@ -858,11 +866,15 @@ msgstr "The city borough where your search address is located"
...
 msgstr "The city’s fifth <0>worst evictor</0> in 2018, A&E is a prime example of a landlord who engages in <1>aggressive eviction strategies</1> to displace low-income tenants. Besides lack of repairs and frivolous evictions in housing court, A&E has also been known to use <2>MCIs</2> to <3>double rents</3> in rent-stabilized buildings."
 
-#: src/components/PropertiesSummary.tsx:108
+#: src/components/PropertiesSummary.tsx:109
 msgid "The most common corporate entity is <0>{0}</0> and the most common business address is <1>{1}</1>."
 msgstr "The most common corporate entity is <0>{0}</0> and the most common business address is <1>{1}</1>."
 
-#: src/components/PropertiesSummary.tsx:102
+#: src/components/ComplaintsSummary.tsx:18
+msgid "The most common reported in this portfolio over the last three years <0/>."
```
samaratrilling, line 874: does this just exclude the "issue" or "issues" because of the plural thing?

sraby:
[ch7315]

client/src/components/APIDataTypes.ts
sraby: good point! you're right that I should name this more specifically. I'll use `totalhpdcomplaints` bc that's what we call it internally

client/src/locales/en/messages.po
sraby: I believe so! it's a little odd tho, lemme look in to how other Plural tags show up in these files.

sraby: omg!! @samaratrilling this is so annoying but the reason these messages werent being extracted properly was because of a very bizarre error where if you import the Plural component _before_ you import the Trans component lingui cannot recognize the Plural tags in strings.... I fixed it in this PR but gonna flag it for review tomorrow as somethign we should brainstorm about documenting bc it's super annoying and not intuitive...

samaratrilling: We can definitely write a custom prettier rule for this!

samaratrilling (approved):

PR #542, opened 2021-09-23
Add black to format WOW Python code.

samaratrilling:
[ch7808]


sraby (approved):
Awesome, thank you @samaratrilling!

PR #543, opened 2021-09-28
Allow "wowza" prefix within bbl-based URLs

sraby:
This PR builds out the "bbl-based" url entrypoint into WOW by allowing users to specify in this URL scheme whether they want to use the new 'wowza' portfolio mapping algorithm or not. 

Basically, we have an existing feature where users can access WOW via a url like `/en/bbl/4015640058` and get redirected to the stable address page on WOW, which in this case is `/en/address/QUEENS/41-25/CASE%20STREET`. This is particularly helpful when you have a whole list of bbls (say, in a spreadsheet), and you want to quickly generate a list of direct links to WOW from it. 

So, this PR is essentially just adding an extra feature to these bbl-based urls where you can now ALSO define whether you want to access a new wowza page or a regular page.

NOTE: since we are introducing another route path, I changed some of our naming conventions around for these bbl-based urls just for clarity.

### DIFF
--- a/client/src/containers/App.tsx
+++ b/client/src/containers/App.tsx
@@ -102,7 +102,7 @@ const WhoOwnsWhatRoutes: React.FC<{}> = () => {
           <AddressPage currentTab={3} {...machineProps} {...props} useNewPortfolioMethod />
         )}
       />
-      <Route path={paths.oldBbl} component={BBLPage} />
+      <Route path={paths.bblSeparatedIntoParts} component={BBLPage} />
       <Route path={paths.bbl} component={BBLPage} />
       <Route
         path={paths.wowzaBbl}
--- a/client/src/containers/App.tsx
+++ b/client/src/containers/App.tsx
@@ -102,8 +102,12 @@ const WhoOwnsWhatRoutes: React.FC<{}> = () => {
           <AddressPage currentTab={3} {...machineProps} {...props} useNewPortfolioMethod />
         )}
       />
+      <Route path={paths.oldBbl} component={BBLPage} />
       <Route path={paths.bbl} component={BBLPage} />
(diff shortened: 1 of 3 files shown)


client/src/containers/App.tsx
```diff
@@ -102,8 +102,12 @@ const WhoOwnsWhatRoutes: React.FC<{}> = () => {
           <AddressPage currentTab={3} {...machineProps} {...props} useNewPortfolioMethod />
         )}
       />
+      <Route path={paths.oldBbl} component={BBLPage} />
```
samaratrilling, line 105: See my note below about changing 'oldbbl' to something else

sraby:
[ch7815]

client/src/routes.tsx
```diff
@@ -56,18 +56,20 @@ export const createWhoOwnsWhatRoutePaths = (prefix?: string) => {
...
-    bblWithFullBblInUrl: `${pathPrefix}/bbl/:bbl`,
+    bbl: `${pathPrefix}/bbl/:bbl`,
+    /** This route path corresponds to a page identical to the `bbl` path above, just specifying that the user
+     * requests to use the new "WOWZA" portfolio method.
+     */
+    wowzaBbl: `${pathPrefix}/wowza/bbl/:bbl`,
+    /** This route path corresponds to a page identical to the `bbl` route above, but with an older url
+     * pattern that we want to support so as not to break any old links that exist out in the web.
+     */
+    oldBbl: `${pathPrefix}/bbl/:boro/:block/:lot`,
```
samaratrilling, line 72: can you call this something like "separatedBBL" to be more descriptive than "oldBBL"

samaratrilling (approved):
Approved, with the note attached about changing the `oldbbl` name

PR #544, opened 2021-09-28
Remove deprecated fields from indicatorhistory endpoint

sraby:
It has been more than two months since we updated the schema of our `address_indicatorhistory` endpoint on the WOW backend (https://github.com/JustFixNYC/who-owns-what/pull/500). This PR removes some extra code we wrote that had made the old names of certain fields still available to cached front ends. It's time... I think?


sraby:
Decision: we are going to merge this in and see If we get any rollbar errors reporting an incompatability.

PR #545, opened 2021-09-29
Add `getLandlordNameFromAddress` helper function

sraby:
This PR adds a new helper function called `getLandlordNameFromAddress` which takes in an AddressRecord object and finds the name of the landlord (or names, if there is a rare edge case where there is more than one) and returns a string array with the name(s) inside.

### DIFF
--- a/client/src/util/helpers.ts
+++ b/client/src/util/helpers.ts
@@ -195,7 +195,7 @@ export default {
     const { ownernames } = addr;
     if (!ownernames) return [];
     const landlords = ownernames.filter((owner) =>
-      ["HeadOfficer", "IndividualOwner", "CorporateOwnerowner"].includes(owner.title)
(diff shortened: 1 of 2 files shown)


client/src/util/helpers.ts
```diff
@@ -186,6 +186,22 @@ export default {
...
+  /**
+   * Note: while almost all address records will just have one landlord listed, there is a rare
+   * edge case where more than one distinct landlord name might be listed, so we return an array
+   * of names here to accommodate that edge case.
+   */
+  getLandlordNameFromAddress(addr: AddressRecord): string[] {
+    const { ownernames } = addr;
+    if (!ownernames) return [];
+    const landlords = ownernames.filter((owner) =>
+      ["HeadOfficer", "IndividualOwner", "CorporateOwnerowner"].includes(owner.title)
```
samaratrilling: Small typo: should be "CorporateOwner" not "CorporateOwnerowner"

PR #547, opened 2021-09-29
Add search address and detail address nodes to portfolio graph visualization

sraby:
This PR refactors our `<PortfolioGraph>` component to dynamically add on nodes to the graph visualization for the user's Search Address, as well as the Detail Address if it differs from the Search Address. The idea behind these additional nodes was to create some feature parity with the "How is this building associated..." feature on the old WOW:

image

Now, when users click on this link, they are redirected to the Portfolio Graph (currently on the Summary tab) to see how their search address is related to the address in focus on the Overview tab:

image

This is particularly important because in order to build trust with our users regarding this new portfolio method, we need to have some feature that illustrates how each building in the portfolio connects back to the original address they searched.

Potential further improvements/considerations:
- adding better labels to the two additional nodes on the graph (like `Search address: ...` and `Focus address: ...`) maybe?
- updating the colors/style to pop out a bit more
- making sure that the graph is fully "unfurled" with the 2 new nodes added (little to no overlapping edges)

### DIFF
--- a/client/src/components/DetailView.tsx
+++ b/client/src/components/DetailView.tsx
@@ -129,6 +129,10 @@ const LearnMoreAccordion = () => (
   </I18n>
 );
 
+const BldgAssociationHeader = () => (
+  <Trans>How is this building associated to this portfolio?</Trans>
+);
+
 class DetailViewWithoutI18n extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
@@ -211,17 +215,17 @@ class DetailViewWithoutI18n extends Component<Props, State> {
                       {!Helpers.addrsAreEqual(detailAddr, searchAddr) &&
                         (useNewPortfolioMethod ? (
                           <Link to={this.props.addressPageRoutes.summary}>
-                            <Trans render="i">
-                              How is this building associated to this portfolio?
-                            </Trans>
+                            <i>
+                              <BldgAssociationHeader />
+                            </i>
                           </Link>
                         ) : (
                           <a // eslint-disable-line jsx-a11y/anchor-is-valid
                             onClick={() => this.setState({ showCompareModal: true })}
                           >
-                            <Trans render="i">
-                              How is this building associated to this portfolio?
-                            </Trans>
+                            <i>
+                              <BldgAssociationHeader />
+                            </i>
                           </a>
                         ))}
                     </div>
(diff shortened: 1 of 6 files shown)


sraby:
[ch7985]

client/src/components/DetailView.tsx
```diff
@@ -207,15 +212,22 @@ class DetailViewWithoutI18n extends Component<Props, State> {
...
+                            <i>
+                              <BldgAssociationHeader />
+                            </i>
+                          </Link>
+                        ) : (
+                          <a // eslint-disable-line jsx-a11y/anchor-is-valid
+                            onClick={() => this.setState({ showCompareModal: true })}
+                          >
+                            <i>
+                              <BldgAssociationHeader />
```
samaratrilling, line 227: This sounds like a tenant association within a building! lol i got confused. Maybe PortfolioAssociationHeader?

client/src/components/PortfolioGraph.tsx
```diff
@@ -3,55 +3,163 @@ import Cytoscape from "cytoscape";
...
+  }
+  return additionalNodes;
+}
+
+/**
+ * Given an existing array of graph nodes and an Address Record, this function
+ * finds the `id` property of a node of type `name` that corresponds to the owner
+ * (or owners) listed with the address.
+ */
+export function getLandlordNodeIDFromAddress(
```
samaratrilling, line 75: any chance you got to do some profiling of this on larger sets? how does the performance seem?

samaratrilling (approved):

client/src/components/PortfolioGraph.tsx
sraby: lemme check!

client/src/components/DetailView.tsx
sraby: Ok did a quick profile and it seems like the computation is pretty instant even for our largest portfolio! Check out this cute pic too and see if u can spot our new bbls: 
image

client/src/components/PortfolioGraph.tsx
sraby: Ok did a quick profile and it seems like the computation is pretty instant even for our largest portfolio! Check out this cute pic too and see if u can spot our new bbls: 
image

samaratrilling: 🤩 bb's first bbls 🤩

PR #551, opened 2021-10-06
Update `wow_bldgs` in place without downtime

sraby:


samaratrilling:
WOw so glad this is getting fixed! can you add that sample rollbar error to the description?

samaratrilling (approved):

PR #555, opened 2021-10-13
Only load Portfolio table when visible to user

sraby:
This PR seeks to reduce the amount of load time for users accessing really really large (> 1000 building) portfolios on the new WOWZA version of WOW. Specifically, this PR targets the Portfolio tab which is the #1 culprit in load time for these large portfolios. Specifically, this PR:

- Does not render the portfolio table at all until it is visible to the user
- Adds an additional warning message on the loading screen when the table contains more than 500 rows
- Adds pagination for when the table contains more than 500 rows 

I spent quite a bit of time researching better solutions and unfortunately it looks like implementing some kind of lazy loading scheme requires react-table version 7, which we do not use and would require significant refactoring (if not building the table from scratch) to upgrade.

### DIFF
--- a/client/src/components/PropertiesList.tsx
+++ b/client/src/components/PropertiesList.tsx
@@ -61,7 +61,7 @@ const getWidthFromLabel = (label: string, customDefaultWidth?: number) => {
 const FIRST_COLUMN_WIDTH = 130;
 const HEADER_HEIGHT = 30;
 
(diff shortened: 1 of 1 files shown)


samaratrilling:
Before:

Screen Shot 2021-10-13 at 2 17 36 PM

After:
Screen Shot 2021-10-13 at 2 17 27 PM

PR #559, opened 2021-10-14
Add warning for when a building's hpd registration expired before its last sale

sraby:
This PR adds a special warning to the Detail View on the overview tab for any building that:
1. has an HPD registration that is expired 
2. was sold to a new owner _after_ the registration expired

Members of our program team made it clear that is important to include expired registrations as it often is just a symptoms of the current landlord being negligent. That being said, there are roughly 6000 buildings with an expired registration that were sold _after_ the registration expired, which would imply there may be a new owner that is not reflected in the hpd contact details. 

So, this PR adds a small flag underneath the "Landlord Contact Details" section for this particular case. I think of this as an "MVP" of a warning system like this that we can build out later.


sraby:
[ch7987]

PR #576, opened 2021-11-19
Upgrade Pluto version to 20v8

sraby:
This PR upgrades the version of the PLUTO dataset that we utilize in the `create_bldgs_table` SQL query as well as the `buildinginfo` sql query. 

This is something we should've probably done already, but it's particularly important now as this commit to nycdb (https://github.com/nycdb/nycdb/pull/183) requires that we only refer to lat/long fields as `latitude` and `longitude` for any PLUTO dataset we are querying. Luckily, `pluto_20v8` is already loaded into nycdb and has only ever included `latitude` and `longitude` as the field names, so we can merge this PR in without having to update nycdb at all, and then update nycdb after the fact without having to worry about breaking anything with WOW.


sraby:
@samaratrilling this is a rather pressing PR, so I'm gonna go ahead and merge it once the tests pass but please give it a quick review once you return and have a free minute!

samaratrilling (commented):
LGTM!

PR #577, opened 2021-12-13
Make Address and BBL clickable on Portfolio Tab

sraby:
This PR makes both the  BBLs & addresses clickable on the Portfolio tab, linking them back to the detail view just like the "View Detail" arrow button already does.


sraby:
[ch7804]

samaratrilling (approved):

samaratrilling:
Looks good!

PR #578, opened 2022-01-18
Add Facebook Pixel whitelisting code snippet

sraby:
This PR addresses a request from WholeWhale: 

> would it be possible for you to whitelist the Facebook pixel on WoW domain? we identified this as the source of the issue for why we couldn't use the conversion we set up in our test run.


sraby:
[ch8411]

PR #580, opened 2022-01-19
Make new WOWZA demo site easier to navigate for beta testers

sraby:
This PR makes some minor adjustments to the WOWZA view of who owns what in order to make it easier for beta testers to navigate the site:
- the portfolio graph on the summary tab no longer takes up the full page, and has a legend and new styling for legibility
- we now save whether the user is searching normally or searching using the new WOWZA method in the actual url of the homepage itself (users at the path `/` are searching normally but when they are at the path `/wowza` they are searching using the new method)
- toggling between the two methods on the site is now more comprehensive with a new button on the homepage and dynamic page links in the header menu


samaratrilling (approved):
Looks good!

PR #588, opened 2022-03-30
add Algolia objectID as hash of contents + tests

austensen:
this PR adds the unique `objectID` for each record that Aloglia requires. It is created as a hash of the rest of the contents of the record (`{"portfolio_bbl": 1234567890, "landlord_names": "name1, name2"}`), so that later we can use the hash to determine if we need to update the record, and hopefully save us some Algolia operations/credits.

Also add tests of this new hash, and adjust the existing tests now that we can ensure consistent ordering of landlord names in the concatenated string.

### DIFF
--- a/portfoliograph/landlord_index.py
+++ b/portfoliograph/landlord_index.py
@@ -1,10 +1,39 @@
 import numpy
 import math
+from typing import Dict, Any
+import hashlib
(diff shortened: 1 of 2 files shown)


sraby (approved):
Looks good!

PR #591, opened 2022-03-31
Add corporation names to the Algolia landlord search index

sraby:
This PR adds a new helper function called `get_corpname_data_for_algolia` which grabs the names of all corporations listed in the `wow_bldgs` as well as one of their associated bbls and creates a new individual algolia search record for each unique corporation name. We then just append these records to the list we send our Algolia app when updating the landlord search index.

### DIFF
--- a/portfoliograph/landlord_index.py
+++ b/portfoliograph/landlord_index.py
@@ -97,10 +97,10 @@ def get_corpname_data_for_algolia(conn):
     with conn.cursor(cursor_factory=DictCursor) as cursor:
         cursor.execute(
             f"""
-            SELECT 
+            SELECT
                 corpname,
                 array_agg(distinct bbl) as bbls
-            FROM    
+            FROM
                 (SELECT
                     unnest(corpnames) as corpname,
                     bbl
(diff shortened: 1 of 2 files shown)


austensen (approved):
The code all looks perfect to me! 

However, I think it'd be good to look more into this issue I found while looking into the corp/portfolio question [sc-8955] before merging this in. Somehow there are a lot of corporations from `wow_bldgs` that are associated with bbls that don't appear in `wow_portfolios` (and some where none of the corp's bbls are in portfolios).

For example, corpname `1016 PROPERTIES LLC` is associated with bbls `1017290035,1017290136` and `1017290035` isn't in any portfolios, and shows up as unregistered on wowza (but shows up normally on wow). So if we added the corpnames and you searched for that corp it would land you at that unregistered bbl (given sort order).

austensen:
[sc-8955]

sraby:
gah! good catch @austensen! I think this may have to do with omitting `Joint Owner` as a valid type of hpd contact when building the portfolio groupings!! I had logged this earlier on this [shortcut card](https://app.shortcut.com/justfixnyc/story/8951/consider-adding-jointowner-as-a-valid-owner-contact-type-for-the-new-wowza-portfolio) but didn't realize the scale of the issue. At the very least, I suspect this must be the problem with the example you mentioned above [1016 PROPERTIES LLC](https://demo-whoownswhat.justfix.nyc/en/address/MANHATTAN/2144/5%20AVENUE), curious if it is for others too

sraby:
Now that #592 has been deployed, there are only about ~600 cases where a corpname is not part of a wow portfolio. We'll keep working on this edge case but we feel comfortable moving forward with adding corpnames to the search bar.

PR #592, opened 2022-04-01
add JointOwner to wow_portfolios table

austensen:
When building the `wow_portfolios` table we previously included only `HeadOfficer, IndividualOwner, CorporateOwner`, this PR adds `JointOwner` as well. Sam identified that we should add JointOwner already, since some only show up in that category. But we now also might need to include it to ensure that all `wow_bldgs` records appear in `wow_portfolios`, since we found that not all of them do currently (see second SC card)

[sc-8951]


austensen:
[sc-8955]

sraby (approved):
Looks good!

PR #593, opened 2022-04-01
fix list concat method for algolia upload

austensen:
We had a problem with the way we were trying to concatenate the two lists of data for the agloglia upload. This PR fixes that bug and adds some type hints for clarity.


sraby (approved):
Cool!

PR #594, opened 2022-04-04
First clear our Algolia index, then replace with new data

sraby:


sraby:
[sc-8969]

sraby:
Algolia upped our quota for index records, so we don't need to implement this for now!

PR #596, opened 2022-04-05
add warning label overlay on graph for large portfolios

austensen:
With the new portfolio algorithm some portfolio networks are really large and we want to have a warning appear on the summary page and map to warn users that it may represent a network of related portfolios

[sc-8965]


austensen:
Made a bit more progress adding the icon, getting it to float against the text (so it doesn't spill over when wrapping). 

image
image


I can't get the scss styling for the text (span) to work though. 

Also the little gap below the warning div is still there. 
image

austensen:
A few outstanding questions:

* We need to decide on what "Learn More" will link to.
* The background color used in the figma file (`#FFBA33`) is very close to the existing variable `$warning` (`#ffb700`), should we change to use the existing one for consistency?
* There was a weird issue where the banner appear off by 1px at some zoom levels, so that the graph behind shows through. After some searching it sounds like this is a weird browser float rounding precision issue and I couldn't find a clear best practice solution. I just adjusted from `bottom: 0` to `bottom: -0.4px` and that hack seemed to solve the rounding issue.

client/src/components/PortfolioGraph.tsx
```diff
@@ -238,6 +239,15 @@ const PortfolioGraphWithoutI18: React.FC<PortfolioGraphProps> = ({ graphJSON, st
           },
         ]}
       />
+      <div className="warning-banner">
+        <div className="float-left">
+          <img src={warning} className="icon" alt="Warning" />
+          <span className="warning">
+            {/* eslint-disable-next-line */}
+            This diagram may show a network of portfolios. <a href="">Learn more.</a>
```
sraby, line 247: Mind wrapping this in a `<Trans>` tag and running `yarn lingui extract --clean` in the `/client` directory to internationalize this snippet of text?

client/src/styles/PropertiesSummary.scss
```diff
@@ -51,6 +50,39 @@
           margin-left: 0;
         }
       }
+      .warning-banner {
+        position: absolute;
+        z-index: 1000;
+        // rounding error causes bottom=0 to be off by 1px at some zoom levels,
+        // this hack seems to fix it https://stackoverflow.com/a/66623955/7051239
+        bottom: -0.4px;
+        background-color: #ffba33;
```
sraby, line 59: There should be a new `$wowza-warning` color variable you can use here, instead!

sraby (changes requested):
Nice work so far!! Really great attention to detail. I left a few comments here and there. Also, are you planning on adding the warning overlay on the map as well in this PR, or in a separate one? No preference either way.

client/src/components/PortfolioGraph.tsx
austensen: Oh right, thanks for the reminder!

austensen:
Okay, I think this is now done! 
* created a BigPortfolioWarning component to use in the map and summary page
* the in the file for the component it has a constant to define the threshold of portfolio size (300) to display the warning
* moved the summary warning below the graph based on sam/tahnee's comment (let me know if this looks ok now, or if there should be changes to the spacing or anything). I think it was working ok now over the graph, so if you prefer to go back to that rather than doing more changes to this new layout that works for me)
image
image

client/src/components/BigPortfolioWarning.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import { t, Trans } from "@lingui/macro";
+import { withI18n, withI18nProps } from "@lingui/react";
+import warning from "../assets/img/icon-warning.svg";
+
+export const PORTFOLIO_SIZE_THRESHOLD = 300;
+
+type BigPortfolioWarningProps = withI18nProps & {
+  size: number;
```
sraby, line 9: `size` sounds a little ambiguous. I'd make this type a bit more descriptive, like `sizeOfPortfolio` or something along those lines.

client/src/styles/PropertiesMap.scss
```diff
@@ -39,6 +39,43 @@
         }
       }
     }
+
+    .warning-banner {
+      position: absolute;
+      z-index: 1000;
+      top: 0;
+      margin: 1.1rem 1.1rem 0 4.5rem;
```
sraby, line 47: I'd make the top margin piece here 1rem instead of 1.1rem — seems to make the banner align better with the zoom control on the map.

sraby (approved):
Amazing work @austensen! I have two knit-picky comments left but otherwise this looks great.

PR #598, opened 2022-04-06
Add "Joint Owner" as a valid owner contact type across Who Owns What

sraby:
This PR makes sure that `JointOwner` names are shown on the "Owners" column of the Portfolio tab, and that `JointOwner` names also can define the `name` property of a node on the portfolio network graph visualization.


sraby:
[sc-8962]

sraby:
Good Example to try: 2615 Grand Concourse, Bronx

PR #599, opened 2022-04-06
Adjust placement and context of portfolio network graph

sraby:
This PR makes some minor styling and wording adjustments to the Summary tab to make the presence of the portfolio network graph on that page make sense. 

image


sraby:
[sc-8982]

austensen (approved):
This all looks great! Love this addition of the text summary, and it works nicely with woza/legacy. I tested it out locally, and tried out woza/legacy version, some without biz multiple addresses, etc. and all the logic for changes to the text look good.

PR #600, opened 2022-04-06
Migrate WOWZA routes across the site to the default root pathname

sraby:
This PR makes sure that all of our `/wowza` pages are now found at the root, and all of the existing root pages, that utilize the old version of WOW, have been moved to the `/legacy` directory. This PR also simplifies our route structure to make sure that all pages, including our about, methodology, etc., have a "legacy" and wowza version so users stay within their selected version of the site upon navigation.

### DIFF
--- a/client/src/components/WowzaToggle.tsx
+++ b/client/src/components/WowzaToggle.tsx
@@ -4,17 +4,17 @@ import React from "react";
 import { useHistory, useLocation } from "react-router-dom";
 
 /**
- * Determines whether a url corresponds to a new WOWZA portfolio mapping page vs an original.
+ * Determines whether a url corresponds to a new WOWZA portfolio mapping page vs an old legacy page.
  */
-export const isWowzaPath = (pathname: string) =>
-  !removeLocalePrefix(pathname).startsWith("/legacy");
+export const isLegacyPath = (pathname: string) =>
+  removeLocalePrefix(pathname).startsWith("/legacy");
 
 /**
  * If the user is on a normal WOW address page, this function generates a pathname for the corresponding
  * WOWZA page with updated portfolio mapping, and vice versa.
  */
 export const getPathForOtherPortfolioMethod = (pathname: string) => {
-  if (!isWowzaPath(pathname)) {
+  if (isLegacyPath(pathname)) {
     return pathname.replace("/legacy", "");
   } else {
     const locale = parseLocaleFromPath(pathname);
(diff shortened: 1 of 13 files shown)


sraby:
[sc-8966]

austensen:
The `createWhoOwnsWhatRoutePaths` function and the concise ternary statements is such a nice way to deal with all this! I got it running locally, and the routes look right to me, but I think there are just a couple places that still need the switching logic:

* [here in DetailView.tsx](https://github.com/JustFixNYC/who-owns-what/blob/b955fd60445c28ce7ccde146ec3e1a115bfef2d6/client/src/components/DetailView.tsx#L117)
* [here in DetailView.tsx](https://github.com/JustFixNYC/who-owns-what/blob/b955fd60445c28ce7ccde146ec3e1a115bfef2d6/client/src/components/DetailView.tsx#L388)

sraby:
great find @austensen thank you!

client/src/components/WowzaToggle.tsx
```diff
@@ -6,18 +6,19 @@ import { useHistory, useLocation } from "react-router-dom";
 /**
  * Determines whether a url corresponds to a new WOWZA portfolio mapping page vs an original.
  */
-export const isWowzaPath = (pathname: string) => removeLocalePrefix(pathname).startsWith("/wowza");
+export const isWowzaPath = (pathname: string) =>
```
samaratrilling, line 9: Let's change this to `isLegacyPath`, and then we can have a positive test here! Since we're not really using `wowza` in the path anymore, it's a tad confusing what this means

netlify.toml
```diff
@@ -10,6 +10,12 @@
   status = 301
   force = true
 
+[[redirects]]
```
samaratrilling, line 13: Ooh excellent

client/src/containers/BBLPage.tsx
```diff
@@ -64,7 +64,7 @@ const BBLPage: React.FC<BBLPageProps> = (props) => {
               ...results.result[0],
               locale,
             },
-            props.useNewPortfolioMethod
+            !props.useNewPortfolioMethod
```
samaratrilling, line 67: so ideally we'd refactor this too to be useOldPortfolioMethod but i get it if you wanna just get this in

samaratrilling (approved):
one small comment but optional to implement. awesome work!

austensen:
Now when I run this locally the home page is the old style. When I add `/legacy` it resets to home and when I add `/woza` it fails. I checked out all the commits to track it down and it changed with [afd7e41](https://github.com/JustFixNYC/who-owns-what/pull/600/commits/afd7e41e2a61f36e98aa5ffba56fe196b2174cc6)

austensen (approved):
oops, sorry, looked more closely at your changes and realized I didn't update my local env to set `REACT_APP_ENABLE_NEW_WOWZA_PORTFOLIO_MAPPING=1`. Everything is working perfectly!!

PR #603, opened 2022-04-11
Add custom zoom widget to portfolio graph viz

sraby:
This pr adds a very simple set of zoom buttons onto our Cytoscape network visualization on the summary tab. 

image


sraby:
[sc-8983]

austensen (approved):
This all looks perfect, and I tested it out locally and it looks and feels good.

PR #604, opened 2022-04-11
When changing portfolio methods, only refresh the page if it's an address page

sraby:
This PR fixes an issue where toggling between old and new WOW portfolio methods on the homepage made the page elements and assets flicker awkwardly. Turns out this was due to the fact that any switch between portfolio methods caused a page refresh. I changed our toggle logic to _only_ trigger a page reload when switching portfolio methods on an internal address page—which we need to do to load in new data for the page.


sraby:
[sc-8998]

austensen (approved):
All the code looks good and tested it out locally and the flicker is gone!

PR #607, opened 2022-04-12
Ensure that only one landlord corresponds to a given `registrationid` from HPD

sraby:
This PR creates an interim fix for the issue of [bbls belonging to multiple landlord portfolios](https://app.shortcut.com/justfixnyc/story/8989/multiple-wow-portfolios-records-cover-same-bbl-set) on WOWZA. So, we changed our SQL query in generating the raw data for the networkx graph to make sure that only one landlord contact corresponds to each registrationid from HPD. In the rare cases where there are multiple owners on a given registration, we have a ranking for prioritizing which owner type we define as the best one to use for portfolio generation. 

Given that the vast majority of these edge casey buildings belong to a portfolio of 1, these changes likely will not affect the scale of our portfolio generation very much, even though we are trimming down the level of possible links a building can have with others. See the [comments section of this Shortcut card](https://app.shortcut.com/justfixnyc/story/8989/multiple-wow-portfolios-records-cover-same-bbl-set) for more analysis.

### DIFF
--- a/portfoliograph/graph.py
+++ b/portfoliograph/graph.py
@@ -51,24 +51,54 @@ def build_graph(dict_cursor) -> nx.Graph:
     # TODO: process synonyms (e.g. folks in pinnacle)
     dict_cursor.execute(
         f"""
-        SELECT DISTINCT
-            firstname,
-            lastname,
-            businesshousenumber,
-            businessstreetname,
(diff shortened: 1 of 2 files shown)


sraby:
[sc-8989]

PR #608, opened 2022-04-12
Add a Loading Page when toggling between portfolio methods

sraby:
This PR makes the transition between legacy and new Address pages more graceful by adding a Loader icon during the transition. While I was at it, I refactored our Loader components to make things more centralized and DRY.


austensen (approved):
This all looks good! Just tested it out locally and it works just as expected.

PR #613, opened 2022-04-13
Fix zoom scaling and add reset button to network visualization

sraby:
Minor structural and design changes to the network visualization on the summary tab.


austensen (approved):
This all looks great! Tested it out locally in the browser, and it all works perfectly. Only minor thing I noticed is that the reset view button is different from how the graph shows on page load (slightly more zoomed in), but that's not necessarily a problem I don't think.

PR #615, opened 2022-04-13
clarify situation for addresses without portfolio (un/incomplete registration)

austensen:
This PR changes the page shown when a building is not in a portfolio because the HPD registration is missing or incomplete. Previously it always said that a property is not registered, but actually in some cases they have a registration but it's missing important necessary to build a portfolio (ie. owner info). Now, the `buildingInfo` data collected for this page includes residential units and hpd registration dates, so we can check whether there is any HPD registration and will say "not registered" or "incomplete registration". This also changes the logic for what to display about the need to register - previously it was based on building class, but now it uses the number of residential units to more closely reflect HPD's guidance from their website (more details in PR comments below)

Longer term, we are planning to do a more substantial change to these non-portfolio properties to provide additional useful information on the property and context around what it means to not be registered (eg. risk that it's an illegal unit and calling HPD for complaints could lead to a vacate order). 

[sc-8815]


austensen:
I think I now have all the logic necessary, but need to check all the possible configurations of registration and property types, improve the layout and info hierarchy, and refine the wording of messages. 

There is also a bug for missing dates being formatted into `1969-12-31` that I need to fix. 

Here's what it looks like now, and here are some good testing bbls:
(incomplete reg: 3011520049, no reg: 4006760090)

image

austensen:
There are a lot of pieces of information to potentially show, so I've tried to show only what's needed to limit possible confusion. These examples should cover all the possibilities.

registration incomplete but active: `4017160001`
image
registration incomplete and expired: `1000530039`
image
registration missing and (maybe) required: `4006760090`
image
registration missing and not required: `1000790003`
image

In addition to the registration info, I also changed the messages about requirements to register. [HPD's site](https://www1.nyc.gov/site/hpd/services-and-information/register-your-property.page) says it's all about the number of units rather than the building class. Before is was using building class, and saying `B` might need to register (if owner lives there), `C` does have to register, and everything else doesn't. However, there are a lot more cases where it's definitely or maybe required:
image

austensen:
[sc-8963]

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -70,103 +76,97 @@ export default class NotRegisteredPage extends Component<Props, State> {
...
 
+    const usersInputAddressFragment = usersInputAddress ? (
+      <>
+        {usersInputAddress.housenumber === " " ? "" : usersInputAddress.housenumber + " "}
+        {usersInputAddress.streetname !== " " && usersInputAddress.streetname}
+      </>
+    ) : null;
+
+    const registrationMissingOrExpired =
+      buildingInfo.registrationenddate === "" ||
```
sraby, line 87: it might be better to check whether `buildingInfo.registrationenddate` is _falsey_ using a NOT `!` operator instead of this equivalency, just in case registrationenddate is ever null or undefined. Totally up to you though.

See here: https://stackoverflow.com/questions/29312123/how-does-the-double-exclamation-work-in-javascript

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -70,103 +76,97 @@ export default class NotRegisteredPage extends Component<Props, State> {
...
+      </div>
+    ) : (
+      <></>
+    );
+
+    let registrationMessage;
+
+    if (!usersInputAddress) {
+      registrationMessage = <Trans>No registration found!</Trans>;
+    } else if (buildingInfo.lastregistrationdate) {
```
sraby, line 123: Hmmmm... so I just checked and it looks like there are 2200 hpd registrations where `lastregistrationdate` is NULL. Not totally sure what that means in the context of HPD's systems but I did also notice that `registrationenddate` is never NULL so maybe using that as our check here could be better? Up to you, as that may affect what we can assume about the truthiness of `lastregistrationdate` elsewhere in this component.

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -175,13 +175,32 @@ export default class NotRegisteredPage extends Component<Props, State> {
...
+                  {failedToRegisterLink}
+                </>
+              ) : (
+                <></>
+              )}
+              {buildingInfo.unitsres === 0 && registrationMissingOrExpired ? (
+                <>{buildingTypeMessage}</>
+              ) : (
+                <></>
+              )}
```
sraby, line 203: I wonder if this logic can be condensed a little bit. Since both cases show the `buildingTypeMessage`, I'm wondering if you keep these as the same case and just have `{lastRegisteredCard}` and `{failedToRegisterLink}` render depending on the number of units? This is another style question so it's your call and I could see arguments either way.

sraby:
Hmmm so something funky is happening with this address: where are these equal signs coming from? 
Screen Shot 2022-04-18 at 10 22 13 AM

client/src/locales/en/messages.po
```diff
@@ -999,11 +995,23 @@ msgstr "This portfolio also had an estimated <0>net {changeType, select, gain {g
...
+msgid "This property doesn't have any residential units, so it is not required to register with HPD."
+msgstr "This property doesn't have any residential units, so it is not required to register with HPD."
+
+#: src/containers/NotRegisteredPage.tsx:99
+msgid "This property has fewer than 3 residential units. If the landlord doesn't reside there, it should be registered with HPD."
+msgstr "This property has fewer than 3 residential units. If the landlord doesn't reside there, it should be registered with HPD."
+
+#: src/containers/NotRegisteredPage.tsx:111
+msgid "This property has more than 2 residential units, so it should be registered with HPD!"
+msgstr "This property has more than 2 residential units, so it should be registered with HPD!======="
```
sraby, line 1014: Ooops! Looks like this code conflict syntax `=======` somehow got committed to the messages file here...

client/src/locales/en/messages.po
```diff
@@ -999,11 +995,23 @@ msgstr "This portfolio also had an estimated <0>net {changeType, select, gain {g
 
 #: src/components/ViolationsSummary.tsx:12
 msgid "This portfolio has an average of <0>{0}</0> open HPD violations per residential unit."
-msgstr "This portfolio has an average of <0>{0}</0> open HPD violations per residential unit."
+msgstr "This portfolio has an average of <0>{0}</0> open HPD violations per residential unit.<<<<<<< HEAD"
```
sraby, line 998: Also here!

sraby (approved):
Amazing work, Maxwell! Your PR handles incomplete hpd registrations really well, and I'm so glad you overhauled our page logic to use residential units instead of building class to determine whether registration is required—the old way is very misleading! 

There are a few code style comments I made, as well as some issue with the messages.po files getting messed up a bit. Other than that, this looks great.

client/src/containers/NotRegisteredPage.tsx
austensen: I totally missed that, thanks for flagging that. I also updated things so this won't cause a weird null -> "1969-01-01" date issue when listing last registered dates.

client/src/containers/NotRegisteredPage.tsx
austensen: Yeah, I wasn't really happy with this either. I looked through some other places and realized you can use `boolean && (<jsx>)` and not everything needs to be a ternary, so I made some more changes to hopefully make this whole section more clear and readable, and also reduce some repetition elsewhere.

client/src/locales/en/messages.po
austensen: so weird, I just looked through everywhere to make sure there are all removed.

client/src/containers/NotRegisteredPage.tsx
sraby: yes! I love `!!` too it's really helpful sometimes

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -56,132 +62,139 @@ export default class NotRegisteredPage extends Component<Props, State> {
...
+            </span>
+          </p>
+        ) : (
+          <p>
+            <b>
+              <Trans>Registration expired:</Trans>
+            </b>{" "}
+            <span className="text-danger">
+              {" "}
+              <Trans>{formattedRegEndDate}</Trans>
```
sraby, line 152: @austensen just realized, this date doesn't need to be internationalized since there is no text beyond just the date!

austensen: oh good catch, thanks, just fixed this

PR #618, opened 2022-04-15
Lazy load the network visualization, and the photo of it within the Big Portfolio modal

sraby:
This PR reduces jank by accounting for the lengthy load times of two new elements:
- The network visualization on the Summary tab takes a longggg time to laod for big portfolios, particularly [the huge one with 2500 bldgs](https://demo-whoownswhat.justfix.nyc/en/address/BROOKLYN/196/RALPH%20AVENUE). I made use of the `LazyLoadWhenVisible` component to allow the user to navigate to the Summary page and see all other content while the network viz loads with a loading icon.
- The image on the BigPortfolioWarning modal flickers on after you open the modal. I added a step to pre-load the image source when the actual warning _banner_ loads so that by the time the user opens the modal it is already loaded.

### DIFF
--- a/client/src/components/BigPortfolioWarning.tsx
+++ b/client/src/components/BigPortfolioWarning.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { t, Trans } from "@lingui/macro";
 import { withI18n, withI18nProps } from "@lingui/react";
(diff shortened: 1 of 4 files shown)


austensen (approved):
This all looks great! I just tested it out for that giant portfolio and the loader appeared without anything weird happening and the image for the modal showed up right away too.

PR #619, opened 2022-04-18
Content and styling tweaks to the network chart and big portfolio warning on WOWZA

sraby:
Simple content and styling tweaks.


sraby:
[sc-9096]

sraby:
[sc-9092]

sraby:
[sc-9093]

austensen (approved):
This all looks great! I tested it out and it all works as expected. I tried clicking another property to add the new annotation, then clicking back to make sure it went away and didn't double up when search and detail properties are the same. I think this does help clarify how your searched properties fit into the portfolio.

PR #620, opened 2022-04-18
Make all outbound links open in a new tab

sraby:
Drawing from our recent QA session, this PR makes sure that all outbound links from our Contentful-driven pages on WOW open in a new tab. This PR also makes some internal links to the methodology page open a new tab as well.


austensen (approved):
This all looks good!

PR #622, opened 2022-04-19
Make sure network viz redistributes nodes after adding detail address

sraby:
When a user selects a new address from the overview tab or portfolio tab, that triggers the addition of a new node to our network viz to show which landlord name owns that new selected property. Before, this new node would by default render in the center of the graph, with a long overlapping edge connecting it with wherever it's related landlord name node is. This would create a very awkward and busy looking graph, so this PR makes sure that we re-render the graph layout whenever the user's selected detail address changes.

This PR also reduces edge overlap overall and makes sure our node labels are visible from more zoom extents.

### DIFF
--- a/client/src/components/PortfolioGraph.tsx
+++ b/client/src/components/PortfolioGraph.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import Cytoscape from "cytoscape";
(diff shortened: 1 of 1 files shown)


sraby:
[sc-9185]

austensen (approved):
This looks great. I went back to check some larger portfolios and see what you mean about the placement, this looks so much better

PR #623, opened 2022-04-19
Analytics

sraby:


sraby:
[sc-9091]

client/src/components/BigPortfolioWarning.tsx
```diff
@@ -29,7 +30,13 @@ export const BigPortfolioWarning = withI18n()(
           <img src={warning} className="icon" alt={i18n._(t`Warning`)} />
           <span className="warning">
             <Trans>Why am I seeing such a big portfolio?</Trans>{" "}
-            <button onClick={() => setModalVisibility(true)}>
+            <button
+              onClick={() => {
+                logAmplitudeEvent("learnWhyPortfolioSoBig");
+                window.gtag("event", "learn-why-portfolio-so-big");
```
samaratrilling, line 36: So we're logging everything to both Google Analytics and Amplitude? Fine if so, just wanted to make sure folks know where the source of truth is.

samaratrilling (approved):
Looks great! Maybe add something to our WOW documentation or engineering handbook about how on WOW, everything is logged to both Google Analytics *and* to Amplitude

client/src/components/BigPortfolioWarning.tsx
sraby: yeah... it's bc google analytics is the legacy system, but we also love Amplitude and want to try that out more and more. I'll def make a not of that in our engineering handbook.

austensen (approved):
All looks good!

PR #624, opened 2022-04-19
fix bugs from mobile browsers QA

austensen:
This PR addresses bugs found in QA testing on mobile browsers. 
So far it's just one issue of the graph network viz zoom controls overlapping the big portfolio modal


austensen:
[sc-9192]

sraby (approved):
Nice!! I tested this out on the BrowserStack emulator and it works great. I also tested the demo link out on all the other devices and browsers on our QA checklist and this didn't change the behavior on those so I think we are good to go.

PR #626, opened 2022-04-19
Make sure the user's locale stays intact on address search

sraby:
Oops! This PR fixes a weird bug where if a user searched and address from the homepage in Spanish, it would redirect them to an address page in English! Not sure how we missed this, but this PR ensures that the user's locale gets passed on correctly when they search an address.


austensen (approved):
This all looks good! Glad you caught that

samaratrilling:
We should add it to the QA list!

PR #629, opened 2022-04-20
Optimize the run time of the new WOWZA search query

sraby:


austensen (approved):
This is great, it's so much faster! I just tested it out and double checked that all the results are identical.

PR #631, opened 2022-04-25
link from number of buildings to portfolio tab

austensen:
On an address page there is a line at the top saying "your search address is associated with __X__ other buildings" with the number underlined, and it feels like you should be able to click it but it wasn't a link. This PR makes it a link to open the Portfolio tab. I didn't make this focusable because it felt like this would only be confusing and redundant since you can toggle through all the available tabs already, but let me know if it should be or there is any other accessibility change I should make. 

image

[sc-9182]

### DIFF
--- a/client/src/containers/AddressPage.tsx
+++ b/client/src/containers/AddressPage.tsx
@@ -165,7 +165,7 @@ export default class AddressPage extends Component<AddressPageProps, State> {
                       to={routes.portfolio}
                       tabIndex={this.props.currentTab === 2 ? -1 : 0}
                       onClick={() => {
-                        window.gtag("event", "portfolio-tab");
+                        window.gtag("event", "num-addrs-click");
                       }}
                     >
                       {assocAddrs.length}
--- a/client/src/containers/AddressPage.tsx
+++ b/client/src/containers/AddressPage.tsx
@@ -160,7 +160,16 @@ export default class AddressPage extends Component<AddressPageProps, State> {
               <div className="float-left">
                 <h1 className="primary">
                   <Trans>
-                    PORTFOLIO: Your search address is associated with <u>{assocAddrs.length}</u>{" "}
+                    PORTFOLIO: Your search address is associated with{" "}
+                    <Link
+                      to={routes.portfolio}
+                      tabIndex={this.props.currentTab === 2 ? -1 : 0}
+                      onClick={() => {
+                        window.gtag("event", "portfolio-tab");
+                      }}
+                    >
+                      {assocAddrs.length}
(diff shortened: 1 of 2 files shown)


client/src/containers/AddressPage.tsx
```diff
@@ -160,7 +160,16 @@ export default class AddressPage extends Component<AddressPageProps, State> {
               <div className="float-left">
                 <h1 className="primary">
                   <Trans>
-                    PORTFOLIO: Your search address is associated with <u>{assocAddrs.length}</u>{" "}
+                    PORTFOLIO: Your search address is associated with{" "}
+                    <Link
+                      to={routes.portfolio}
+                      tabIndex={this.props.currentTab === 2 ? -1 : 0}
```
samaratrilling, line 166: The way I'm reading this, it says if you're already on the portfolio tab, go left one tab; otherwise stay on the same page. That seems backwards to me - am I reading it wrong?

client/src/containers/AddressPage.tsx
```diff
@@ -160,7 +160,16 @@ export default class AddressPage extends Component<AddressPageProps, State> {
...
+                      onClick={() => {
+                        window.gtag("event", "portfolio-tab");
```
samaratrilling, line 168: maybe we should name this event something different, like num_addrs_click or something, to differentiate it from someone clicking the portfolio tab. would be interesting to know how many people are clicking that vs the tab

samaratrilling (commented):
Couple questions, but great job getting this done so quickly!!

client/src/containers/AddressPage.tsx
austensen: good idea, just changed that. that'll be nice to know if it's being used. Originally I was just thinking that I didn't want to mess up the counts of portfolio visits, but we can just remember to add these two tags to get all portfolio tab visits.

client/src/containers/AddressPage.tsx
austensen: So this was all just copied from the link for clicking directly on the portfolio tab. My understanding of `tabIndex` is that it's only about tab-navigation, and not what shown visually - that part is handled below by changing css class `"active"` on and off based on `currentTab` values. For a link to be tab focusable it needs to have `tabIndex=0`, but when a tab is already active we want to exclude so set `tabIndex=-1`. 

https://user-images.githubusercontent.com/16906516/165582510-9da15f14-09d8-4f73-a79c-774124fd0f5e.mov

samaratrilling (approved):
Excellent, makes sense! LGTM!

PR #633, opened 2022-05-02
Consolidate Landlord Search subcomponents

sraby:
This PR improves on #630 making sure that we only send 1 network request to Algolia for each keystroke inputted in to our landlord name search bar. By consolidating some of our component logic, I was able to achieve a state where we only send one network request on rendering the searchbar, one network request on every input character typed into the searchbar, and don't send any requests on deleting characters or even typing already cached searches.

This PR should leave the accessibility of the search bar unchanged.


sraby:
@austensen @samaratrilling this was a small, time-sensitive change so I merged it myself but if you have time today and can give it a quick review that would be super helpful! It's only 1 file to look at

austensen (commented):
This all looks good to me, and testing it out it all works as expected.

PR #634, opened 2022-05-02
Don't request Landlord Search results until user types 2 characters

sraby:
This PR builds on #633 and #630 to even further reduce the number of network request we make to Algolio when a user searches their landlord name. Specifically, we now do not trigger a network request for a search until the user has typed in 2 characters into the search bar.

The basic idea behind this PR is that Algolia provides their own state and state-updater functions for the `<SearchBox>` component, called `currentRefinement` and `refine` respectively, that keep track of the current search query and allow you to update it on changes to the form input, which by default triggers a network request. I created our own internal copy of this state with `searchQuery` and `setSearchQuery` so we can keep track of user inputs without triggering any Algolia network request. Then, in our useEffect hook, I am updating the `currentRefinement` state dynamically, only when our `searchQuery` is 2 or more characters in length.

### DIFF
--- a/client/src/components/LandlordSearch.tsx
+++ b/client/src/components/LandlordSearch.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import algoliasearch from "algoliasearch/lite";
 import {
   InstantSearch,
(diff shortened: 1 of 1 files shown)


austensen (approved):
This all looks good, and testing it out locally works as expected

PR #635, opened 2022-05-03
add amplitude logging on more events

austensen:
After talking with Tahnee and Risha last week we identified a bunch of analyses they'd like to see in amplitude, and so this PR adds more event logging to make sure we have all the data necessary. 

Mostly I just was simply adding logs, but I also noticed that an existing one on the timeline period switcher was logging twice with each click, so I moved that and fixed it. I also added an optional `onClick` to the `Accordion` components props so you can more easily log the click to expand event.

### DIFF
--- a/client/src/components/Amplitude.ts
+++ b/client/src/components/Amplitude.ts
@@ -10,7 +10,7 @@ if (!API_KEY) throw new Error("No Amplitude API key defined!");
 
 amplitude.getInstance().init(API_KEY);
 
-type AmplitudeEvent =
+export type AmplitudeEvent =
   | "closeFeatureCalloutWidget"
   | "openFeatureCalloutWidget"
   | "viewPreviousEntryOnFeatureCalloutWidget"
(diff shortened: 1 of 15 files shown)


austensen:
[sc-9369]

client/src/components/Indicators.tsx
```diff
@@ -279,9 +283,6 @@ class IndicatorsWithoutI18n extends Component<IndicatorsProps, IndicatorsState>
                               "form-radio" +
                               (this.state.activeTimeSpan === timespan ? " active" : "")
                             }
-                            onClick={() => {
-                              window.gtag("event", "month-timeline-tab");
-                            }}
```
sraby, line 284: eeek! so we were tagging every time span change as "month" huh? glad we are fixing this now

client/src/components/Amplitude.ts
```diff
@@ -27,7 +27,44 @@ type AmplitudeEvent =
...
+  | "hpd-not-registered-page"
+  | "dob-not-registered-page"
+  | "dof-not-registered-page"
+  | "dap-not-registered-page"
+  | "numAddrsClick"
+  | "timelineTab"
+  | "portfolioTab"
+  | "summaryTab"
+  | "whoIsLandlordAccordian"
+  | "detailsOpenContactCard"
```
sraby, line 56: Hmmm so some of these are in kebab-case and some are in camelCase. Maybe we should consider defaulting just to camelCase since that's the case that previous event labels have used? Not a huge dealbreaker tho.

sraby (approved):
Looks great @austensen! I left one comment about case syntax but other than that this is good to go.

client/src/components/Amplitude.ts
austensen: Yeah, those non camelCase ones are because of the template strings we need to use in some places, and not wanting to complicate the code in each log step by doing a bunch of string cleaning (especially since the google ones use kebab-case (love that name btw!)

client/src/components/Indicators.tsx
austensen: yeah, and also this is the one that was getting triggered twice for each click. Not sure why the duplication was happening, but moving the logging to `onChange` instead of `onClick` solved that.

PR #637, opened 2022-05-31
Add API endpoint to get portfolio size for DAP Portal

sraby:
This PR addresses an issue where the count of associated properties for a building on ANHD's [DAP Portal](https://portal.displacementalert.org/property/3012380016) did not match up with the new count of associated properties on Who Owns What given our new portfolio mapping method.

image

Turns out that ANHD's [DAP Portal](https://portal.displacementalert.org/property/3012380016) actually relies on a special API endpoint (`/api/address/dap-aggregate`) we created just for them that sends their tool a bunch of aggregate information on a building's portfolio if they send a request with a bbl in it. However, this API endpoint still references the old portfolio method, so in order to fix this, I created a new API endpoint at `/api/address/dap-portfoliosize` that retrieves portfolio sizes based on our new version of WOW.

I also simplified the response of this API endpoint, and it works like this:
- The only request parameter needed is bbl. If the request includes an hpd-registered bbl, the response will include the portfolio size on WOW:

`/api/address/dap-portfoliosize?bbl=4015640058` responds with `{"result": {"portfolio_size": 154}}`

- If the bbl in the request is not registered or not a real bbl, the response will be null:

`/api/address/dap-portfoliosize?bbl=4015640000` responds with `{"result": null}`

- If the bbl in the request is not 10 characters, the response will include an error:

`/api/address/dap-portfoliosize?bbl=40156400000` responds with 

```
{"error": "Bad request", "validationErrors": {"bbl": [{"message": "This should be a 10-digit padded BBL.", "code": "invalid"}]}}
```

### DIFF
--- a/tests/exported_test_data.sql
+++ b/tests/exported_test_data.sql
@@ -256,4 +256,12 @@ INSERT INTO public.wow_bldgs (housenumber, streetname, zip, boro, registrationid
 -- PostgreSQL database dump complete
 --
 
+--
+-- Manual creation of sample wow_portfolios table for testing:
+--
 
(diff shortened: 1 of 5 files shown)


sraby:
[sc-9665]

austensen (approved):
Tested it out locally, and this all looks good to me!

PR #642, opened 2022-07-06
add backend logic for oca evictions data

austensen:
This is the first PR to begin adding OCA eviction filing data to WOW, by creating a module to pull the OCA data into our DB and update the relevant tables. Once this is merged, a second PR will be made on nycdb-k8s-loader to use the new module for the auto-updating. Then a third PR on this repo will incorporate all the new data into the wow frontend. 

This PR adds a new python module, in the same pattern as with portfoliograph, to be used in `nycdb-k8s-loader` to automate the updating of that data source. The process involves connecting via ssh to the private "level-1" address-level OCA database, extracts the summarized data we need for WOW. This includes two tables - one for building level counts of total filings since 2017 (matching the marshals data), and another for the building-level monthly filing counts for the timeline tab. All of these are limited to only buildings with 11+ residential units, per our agreement with OCA. These two queries are run against the OCA database and saved as local CSV files (like nycdb) then uploaded to our main nycdb/wow database. The wow_bldgs and address_indicators tables are also edited to incorporate the new variables. Because we can't share the private data, if you don't provide those credentials the tables are still created but never populated, to the final variables on the main tables are all null but nothing breaks.


austensen:
We need the package `sshtunnel` to connect to the OCA data, and it has `cryptography` as a dependency. At first CircleCI builds were failing because it was trying to install that package from source, which requires rust. I was able to just add a step to that section of the CI upgrading pip and then it found a version with a wheel so we don't need rust.

sql/create_bldgs_table.sql
```diff
@@ -86,7 +86,11 @@ select distinct on (registrations.bbl)
   exemptions.yearstarted421a::smallint,
   firstdeeds.documentid as lastsaleacrisid,
   firstdeeds.docdate as lastsaledate,
-  firstdeeds.docamount as lastsaleamount
+  firstdeeds.docamount as lastsaleamount,
+  case
+    when coalesce(pluto.unitsres, 0) < 11 then NULL
```
austensen, line 91: should be <= 11

austensen:
(Sam reviewed/approved IRL)

PR #643, opened 2022-07-12
Remove duplicate `.csv` file suffix from data export

sraby:
I noticed that data exports from WOW were being named `who_owns_what_export.csv.csv`. This PR removes the duplicate filename extension.


sraby:
@austensen @shakao this is a one-liner so I'm gonna go ahead and merge this but just flagging for u

PR #646, opened 2022-07-12
fix oca units cutoff, address indicators filtering

austensen:
A couple fixes to the SQL files for OCA indicators: correcting the unit size cutoff for reporting and the bbl filtering for oca address indicators for the timeline.


sraby (approved):
Looks awesome! Can't wait to have this live 🥳

PR #654, opened 2022-09-02
use pluto_latest nycdb table

austensen:
They recently published an Open Data dataset for PLUTO that keeps a consistent URL and is updated with each new version. This PR switches everything over to use this instead of manually updating for set pluto versions each year. 

This will help keep the data up-to-date with the now very frequent updates DCP makes to PLUTO. There may still be occasional changes to the schema that would require updates, but that's true of all the datasets and should be rare. If that happens and it breaks NYCDB, we'll be alerted and the most recent good version will stay in the database, so it won't be a problem. 

The only location where there is a set version of PLUTO now is in the SQL files that are run on the OCA database that BetaNYC maintains. This isn't much of a problem because the only use for that is to suppress eviction filings data for < 11 unit properties per our data agreement.  

[sc-9580]


austensen:
Yeah, the specific version files come as zip from city planning, but this updating version on open data is just a regular csv.

PR #655, opened 2022-09-06
Add OCA eviction filings to frontend

austensen:
This PR finishes up the project of adding OCA (office of court administration) eviction filings data to WOW. First the backend work was done in #642 to extract the necessary data from the private DB maintained by BetaNYC for the address-level version of the [housing court data ](https://github.com/housing-data-coalition/oca). Then those changes were incorporated into the auto-updating NYCDB (see https://github.com/JustFixNYC/nycdb-k8s-loader/pull/115 & https://github.com/JustFixNYC/nycdb-k8s-loader/pull/117). This PR deals with getting the data from those new database tables and updating the UI to incorporate the new data. It follows the same process as for executed evictions on the overview page, and other indicators on the timeline page, with the one difference being that we can't report filings data on buildings with fewer than 11 units so there is a difference between 0 filings and `N/A` and this is noted in the timeline graph. 

Overview tab:
image

Portfolio tab:
image -> image

Timeline tab:
image

[sc-9478]


client/src/components/PropertiesList.tsx
```diff
@@ -411,18 +411,29 @@ const TableOfData = React.memo(
...
           },
           {
-            Header: i18n._(t`Evictions`),
+            Header: i18n._(t`Evictions Since 2017`),
             columns: [
               {
                 Header: (
                   <>
-                    <Trans>Since 2017</Trans>
+                    <Trans>Filied</Trans>
```
shakao, line 419: small: "filed"

client/src/components/PropertiesList.tsx
```diff
@@ -411,18 +411,29 @@ const TableOfData = React.memo(
...
+                    <ArrowIcon />
+                  </>
+                ),
+                accessor: (d) => d.evictionfilings || null,
+                id: "evictionsfilings",
+                width: getWidthFromLabel(i18n._(t`Filied`)),
```
shakao, line 425: small: "filed"

sql/create_bldgs_table.sql
```diff
@@ -89,7 +89,7 @@ select distinct on (registrations.bbl)
   firstdeeds.docamount as lastsaleamount,
   case
     when coalesce(pluto.unitsres, 0) < 11 then NULL
-    else coalesce(oca.eviction_filings_since_2017, 0)
```
shakao, line 92: can you explain this change?

austensen: properties will only show up in the OCA tables if there are eviction filings, so if the value is missing we can call that zero, except in cases of buildings with < 11 units where we are suppressing data.

shakao: oh sorry i meant why it moved from being `oca.eviction_filings_since_2017` to just `eviction_filings_since_2017` ?

shakao (approved):
looks awesome!! 🎉

sql/create_bldgs_table.sql
shakao: ah okay, that makes sense thank you!! that happened here, yes? https://github.com/JustFixNYC/nycdb-k8s-loader/pull/117

austensen: yes that's right!

PR #661, opened 2022-10-13
Floating 'modal' component, button styling for survey CTA

shakao:
[sc-11089]

- adds a `StickyModal` component (very open to better names for this) that is fixed & dismissable, can be aligned to the top/bottom and left/right of the page
- styling for the modal (black bg, white text, spacing)
- styling for the custom survey button (pink, black text)
- ports variables and styles from orgsite -> we should look into making these variable files an npm package or something porting them manually to WOW and LATAC feels a bit messy

notes:
- the base font size for WOW is 10px (as opposed to the orgsite's 16), so sizing will look a little different for now. i'm keping the rems sizing to make unification easier later on
- i'm not including the actual survey embed since the linked survey is a prototype still (code can be found here https://github.com/JustFixNYC/who-owns-what/commit/357c6d255f1d5a4d1ff8dcab0d7f0f71a57949bc)

Screen Shot 2022-10-13 at 1 55 30 PM

Screen Shot 2022-10-13 at 1 55 43 PM


client/src/styles/_button.scss
```diff
@@ -124,3 +124,18 @@
 .icon-mail:before {
   border-radius: 0.1rem;
 }
+
+// WHO ARE OUR USERS SURVEY BUTTON
+.woau-survey-button {
```
austensen, line 129: Think this is a typo - should be `waou`, right?

austensen (approved):
This looks great! I think the name `StickyModal` works well. Exciting to start getting parts of the new brand in here. Good idea about the npm package for all that

client/src/styles/_button.scss
shakao: oops yes, thank you!

PR #662, opened 2022-10-20
use new portfolio mapping for bbl parts route

austensen:
We have an old version of the BBL route that is still being used by ANHD's DAP Map and these are going to the `/legacy` version of WOW portfolio mapping method. This PR updates the route so they all go to the new WOW version.

image

[sc-11155]


shakao (approved):

shakao:
ah! sorry for the slow review, lgtm!

PR #663, opened 2022-10-24
Move survey ID to environment variable

shakao:
[sc-11165]

it looks like all our client-side environment variables are directly committed to github, unfortunately, so we won't be able to change the survey ID without doing a code change then pushing to production. still moving the ID to the environment since it cleans up the App.tsx code a bit, but we should make sure to remove the temporary ID before we release WOW


austensen (approved):
This all looks great! (I'll make a card in the backlog for moving the client variables out of the code)

PR #669, opened 2022-11-09
Read survey cookie into state, hide button permanently on submit

shakao:
moved all the cookie stuff to state so that we could force a rerender when the user submits the survey (in order hide the button without a page reload)! [sc-11209]


austensen (approved):
Looks good!

PR #670, opened 2022-11-14
Remove default survey value, set in Netlify

shakao:


austensen (approved):
Looks good! Thanks for doing this!

PR #689, opened 2022-12-21
Add close button to big portfolio warning banner on map

austensen:
This PR adds a close button for the big portfolio warning banner on the map and summary tab. For now it reappears on reload, but once JustFixNYC/justfix-ts#41 adds the `storageFactory` to our `util` package we can use sessionStorage to keep it closed for the duration of the user's session. 

This also refactors the banner component to allow for other content, so we can use them for giving info about filter selections once we add that feature in addition to the portfolio size warning. The filter banners only shown in the images below as a preview for how it'll look later with multiple.

Desktop:
image
image

Mobile:
image
image

[sc-11308]
[sc-11309]


client/src/components/MapBanner.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </div>
+      {children}
+    </div>
+  ) : (
+    <></>
+  );
+};
+
+type BigPortfolioBannerProps = Omit<
+  MapBannerProps & {
```
shakao, line 45: hmm, wondering if maybe instead of constructing from MapBannerProps, the BigPortfolioBannerProps could just be `sizeOfPortfolio` and `className`, since those appear to be the only ones used?

client/src/components/PropertiesMap.tsx
```diff
@@ -266,11 +266,13 @@ export default class PropertiesMap extends Component<Props, State> {
                 left: "10px",
               }}
             />
-            {useNewPortfolioMethod ? (
-              <BigPortfolioWarning sizeOfPortfolio={this.state.addrsPoints.length} />
-            ) : (
-              <></>
-            )}
+            <div className="MapBanner__container">
```
shakao, line 269: do we want this div to be always visible, even when useNewPortfolioMethod is false?

client/src/components/PropertiesMap.tsx
```diff
@@ -266,11 +266,13 @@ export default class PropertiesMap extends Component<Props, State> {
...
+              {useNewPortfolioMethod ? (
```
shakao, line 270: small: could be `{useNewPortfolioMethod && <BigPortfolioBanner ... />}` to avoid the empty fragment

client/src/styles/PropertiesMap.scss
```diff
@@ -40,35 +40,62 @@
...
+      .MapBanner {
         display: flex;
-        align-items: center;
-        padding: 0;
+        position: relative;
+        min-height: 5.4rem;
+        padding: 0.5rem 1.7rem 0.5rem 4.2rem;
+        margin-bottom: 0.5rem;
+        border-radius: 0.2rem;
+        font-family: "Inconsolata", monospace;
```
shakao, line 60: small: might be nice to eventually move this font to a sass variable as well

client/src/styles/PropertiesSummary.scss
```diff
@@ -73,41 +73,42 @@
...
-          display: flex;
-          align-items: center;
-          padding: 0;
-        }
-
-        .icon {
-          width: 2.536rem;
-          height: 2.277rem;
-          margin-right: 1.012rem;
+        .MapBanner__message {
```
shakao, line 95: it looks like a chunk of the PropertiesSummary css is duplicated from PropertiesMap, wondering if that can be factored into a shared file or something? might be useful if it's common to the MapBanner and we end up reusing the MapBanner component elsewhere (def could also be done later)

austensen: oh yeah for sure, good idea - just did that.

client/src/styles/PropertiesMap.scss
austensen: oh yeah for sure good idea, I'll check with Tahnee about how this fits into the design system and do this later.

client/src/components/PropertiesMap.tsx
austensen: I tried this but then the `false` comes through and it causes and error in the containing Map component

client/src/components/PropertiesMap.tsx
austensen: oh yeah good point. We will eventually need the container for the filter-related banners, but for now that's not necessary so will put that inside and adjust later when needed.

client/src/components/MapBanner.tsx
austensen: right of course, that didn't make any sense. thanks!

client/src/components/PropertiesMap.tsx
shakao: ohh i see, the Map is doing something with the children. no worries then!

shakao (approved):
lgtm!

PR #690, opened 2022-12-22
fix git in devcontainer

austensen:
Git hasn't been working inside the devcontainer. Turns out it's a known issue when `git-lfs` in enabled. The solution is explained here: https://github.com/devcontainers/features/tree/main/src/git-lfs


shakao (approved):
very cool! it looks like they're migrating to https://github.com/devcontainers/features/tree/main/src/git-lfs where there's a slightly different version of the bash script -- not urgent, but we might want to note in case we later need to pull updates

austensen:
Oh amazing, thanks so much for catching that! Just updated and it's working perfectly and now we don't have to copy that script

PR #693, opened 2023-01-05
update latestdeed docdate for unlocknyc api

austensen:
Minor change to the query for the API endpoint we added for UnlockNYC in #691. Now fill in missing `docdate` values with `recordedfiled`


shakao:
not deeply familiar with the API use case but i'm curious about the tradeoffs of having our API fill in the missing values vs returning both as separate fields and having the end user process them?

austensen:
Yeah, I was thinking the same thing initially, but then realized that since they are also still using our main wow building-info API and it does this coalesce and calls it docdate, it might be confusing to be different on this one. Also I'm trying to upgrade our heroku stack for the wow server so this is a nice simple commit to trigger the redeploy.

shakao:
oh yeah totally makes sense to match the existing field behavior! thanks for the context, PR looks good to me

PR #696, opened 2023-01-19
update amplitude sdk, track addr change & table sort

austensen:
In anticipation of the release of the new portfolio filter feature, we want to start collecting a baseline of data on how often users are visiting other properties. This includes clicking on an address/bbl/detail from the portfolio table, and clicking on map points from the overview page. Also, this adds tracking for when a user sorts a column in the portfolio table (including event properties with the column name), so we can see if these interactions with the table also increase with the new filter features. 

While we're making these changes I've also upgraded WOW to the new Amplitude SDK, which we are now also using for org-site (JustFixNYC/justfix-website/pull/328)

I also make a minor change to the `logAmplitudeEvent` in WOW to match what we do for org-site, which just allows us to more flexibly add properties for events.


client/src/containers/AddressPage.tsx
```diff
@@ -260,7 +265,11 @@ export default class AddressPage extends Component<AddressPageProps, State> {
               <PropertiesList
                 state={state}
                 send={send}
-                onOpenDetail={this.handleAddrChange}
+                onOpenDetail={(bbl: string) => {
+                  this.handleAddrChange(bbl);
+                  logAmplitudeEvent("addressChangePortfolio", analyticsEventData);
```
shakao, line 270: do we want the granularity of knowing that the address changed from a click on the address vs the bbl vs the view details button?

austensen: oh right, thanks, yeah that'd be good to know! Just updated to have the column name as a property on the event.

austensen:
[sc-11352]

PR #700, opened 2023-01-20
portfolio table filters

austensen:
Initial implementation of the portfolio filters for table on desktop. This sets up the react Context to share filter options and selections between the filters bar component and the table (and later the map). It upgrades from the older react-table v6 to TanStack Table v8. Adds a multiselect component adapted from [multiselect-react-dropdown](https://github.com/srigar/multiselect-react-dropdown), as well as a binary toggle and min/max numeric input. It also addres a details/summary dropdown component to contain the filter components. Along with the new v8 table, this adds the filter logic to update the columns. The filters we have are for presence of any rent stabilized (toggle), landlord name (multiselect), units in the building (min/max), and zipcode (multiple select). The "landlord" column/filter now includes all the different types of contacts (owner, corporation, head officer, etc.) in both the filter, and in a new expanding row for the column/table. 

There are a number of follow up PRs that will address all of the review comments, and later mobile styles and map view will be added separately. 

[sc-11336]


client/src/components/Alert.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </div>
+    ) : (
+      <></>
+    );
+  }
+);
+
+Alert.displayName = "Alert";
+
+Alert.propTypes = {
```
shakao, line 68: is default typechecking not working with the AlertProps?

client/src/components/Alert.tsx
```diff
@@ -0,0 +1,80 @@
...
+  closeType?: "none" | "state" | "session" | "local";
+  storageId?: string;
+  className?: string;
+}
+
+export const Alert = forwardRef<HTMLDivElement, AlertProps>(
+  ({ type, variant, closeType, storageId, className, children, ...props }, ref) => {
+    const usingStorage = closeType && ["session", "local"].includes(closeType);
+
+    // TODO: see if there's a way to do this that works with intellisense.
```
shakao, line 23: i think one way to do this would be to have the base `Alert` only do the "none" closetype, then subclass/extend for state, and separately subclass for local/session storage

client/src/components/Multiselect.tsx
```diff
@@ -0,0 +1,735 @@
+// This component is slightly adapted from https://github.com/srigar/multiselect-react-dropdown
```
shakao, line 1: hmm wondering if it would be worth importing and extending from their component instead of forking, then overwriting the functions we need (render, etc)? would be able to quickly pull future changes/fixes to this, but might be more effort than it's worth

client/src/components/Multiselect.tsx
```diff
@@ -0,0 +1,735 @@
...
+  disablePreSelectedValues: false,
+  selectedValues: [],
+  isObject: true,
+  displayValue: "model",
+  showCheckbox: false,
+  selectionLimit: -1,
+  placeholder: "Select",
+  groupBy: "",
+  style: {},
+  emptyRecordMsg: "No Options Available",
```
shakao, line 718: does this need to be localized?

client/src/styles/PortfolioFilters.scss
```diff
@@ -0,0 +1,286 @@
...
+  font-family: "Inconsolata";
+  font-style: normal;
+  font-weight: 600;
+  font-size: 1.4rem;
+  line-height: 100%;
+  letter-spacing: 0.02em;
+
+  .filter-for {
+    text-align: right;
+    min-width: 102px;
```
shakao, line 22: rem instead of px?

client/src/styles/PortfolioFilters.scss
```diff
@@ -0,0 +1,286 @@
...
+
+            .multiselect-container,
+            .minmax-container {
+              display: flex;
+              flex-direction: column;
+              margin-top: 0.8rem;
+              .optionListContainer {
+                position: static;
+              }
+              .jf-alert {
```
shakao, line 167: could consider moving this to the alerts css file

client/src/styles/PortfolioFilters.scss
```diff
@@ -0,0 +1,286 @@
...
+                  margin-right: auto;
+                }
+              }
+              .inputs-container {
+                display: flex;
+                gap: 1rem;
+                align-items: center;
+                justify-content: center;
+                input {
+                  // background: $justfix-white;
```
shakao, line 188: still needed?

client/src/styles/PortfolioFilters.scss
```diff
@@ -0,0 +1,286 @@
+@import "_vars.scss";
+@import "_scrollbar.scss";
+@import "_button.scss";
+
+.PortfolioFilters {
+  display: flex;
+  gap: 1.8rem;
+  padding: 1.4rem;
+  background-color: #f2f2f2;
```
shakao, line 9: small: color variable (used in a couple places below also)

client/src/styles/PortfolioTable.scss
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    tbody {
+      tr {
+        border-bottom: $table-border-light;
+
+        &.row-odd {
+          background-color: #f7f7f7;
+          &:hover {
+            background-color: #f2f2f2ff;
```
shakao, line 93: typo?

client/src/styles/PortfolioTable.scss
```diff
@@ -0,0 +1,220 @@
...
+    align-items: center;
+    justify-content: space-around;
+
+    .page-info {
+      display: flex;
+      align-items: center;
+      margin: 3px 10px;
+      white-space: nowrap;
+
+      input {
```
shakao, line 166: just making sure i'm reading this right, this is a custom wow input and not design system, right?

client/src/components/PropertiesList.tsx
```diff
@@ -1,78 +1,68 @@
...
 
-const ArrowIcon = () => (
-  <span className="arrow-icon">
-    ↑<span>↓</span>
-  </span>
-);
+export type IFilterContext = {
+  totalBuildings?: number | undefined;
+  filteredBuildings?: number | undefined;
+  filterSelections: FilterValues & { rsunitslatest?: boolean };
```
shakao, line 29: wondering why this one is outside of the FilterValues type?

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+              cell: (info) => info.getValue(),
+              footer: (props) => props.column.id,
+              enableColumnFilter: false,
+              size: "auto",
+            },
+            {
+              accessorKey: "recentcomplaintsbytype",
+              header: i18n._(t`Top Complaint`),
+              cell: ({ row }) => {
+                const mostCommonType = findMostCommonType(row.original.recentcomplaintsbytype);
```
shakao, line 235: i'm not super familiar with React.memo, does that ensure that this function is only being called once per row vs on render?

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+                      style={{ minWidth: header.getSize() || undefined }}
+                    >
+                      {header.isPlaceholder ? null : (
+                        <>
+                          <div
+                            {...{
+                              className: header.column.getCanSort()
+                                ? "cursor-pointer select-none"
+                                : "",
+                              onClick: header.column.getToggleSortingHandler(),
```
shakao, line 513: log `portfolioColumnSort` event here?

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+                </Fragment>
+              );
+            })}
+          </tbody>
+        </table>
+        <div className="pagination">
+          <div className="prev">
+            <button
+              className="page-btn"
+              onClick={() => table.previousPage()}
```
shakao, line 571: log event?

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+                <option key={pageSize} value={pageSize}>
+                  Show {pageSize}
+                </option>
+              ))}
+            </select>
+          </div>
+          <div className="next">
+            <button
+              className="page-btn"
+              onClick={() => table.nextPage()}
```
shakao, line 610: log event?

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+        .map((contact, i) => (
+          <li key={i}>
+            {contact.title}: {contact.value}
+          </li>
+        ))}
+    </ul>
+  );
+};
+
+type ArrowIconProps = {
```
shakao, line 637: could be consolidated with `Icons.tsx` maybe

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+        ownernames: Array.from(new Set(Array.from(ownernamesOptionValues.keys()).flat())).sort(
+          compareAlphaNumLast
+        ),
+        unitsres: unitsresOptionValues,
+        zip: Array.from(zipOptionValues.keys())
+          .filter((zip) => zip != null)
+          .sort(),
+      },
+    });
+    console.log("filter options updated");
```
shakao, line 700: leftover console log (couple more below as well)

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+  const unitsresOptionValues =
+    table.getColumn("unitsres").getFacetedMinMaxValues() || MINMAX_DEFAULT;
+  const zipOptionValues = table.getColumn("zip").getFacetedUniqueValues();
+
+  React.useEffect(() => {
+    setFilterContext({
+      ...filterContext,
+      filterOptions: {
+        ownernames: Array.from(new Set(Array.from(ownernamesOptionValues.keys()).flat())).sort(
+          compareAlphaNumLast
```
shakao, line 692: just for my own clarity, what are the cases with numeric names?

client/src/components/PortfolioTable.tsx
```diff
@@ -0,0 +1,741 @@
...
+
+    const columns = React.useMemo<ColumnDef<AddressRecord, any>[]>(
+      () => [
+        {
+          accessorFn: (row) => `${row.housenumber} ${row.streetname}`,
+          id: "address",
+          header: () => i18n._(t`Address`),
+          cell: ({ row }) => {
+            return (
+              <Link to={createRouteForFullBbl(row.original.bbl, locale)}>
```
shakao, line 98: it looks like with the links here/in the bbl column, we are no longer logging the `addressChangePortfolio` event--intentional?

client/src/components/PortfolioFilters.tsx
```diff
@@ -0,0 +1,394 @@
...
+          ...filterContext.filterSelections,
+          ownernames: selectedList,
+        },
+      });
+    };
+
+    const [unitsresActive, setUnitsresActive] = React.useState(false);
+    const [unitsresIsOpen, setUnitsresIsOpen] = React.useState(false);
+    const onUnitsresApply = (selectedList: any) => {
+      console.log({ onUnitsresApply: selectedList });
```
shakao, line 63: leftover console log

client/src/components/PortfolioFilters.tsx
```diff
@@ -0,0 +1,394 @@
...
+    const { about, methodology, legacy } = createWhoOwnsWhatRoutePaths();
+
+    const { filterContext, setFilterContext } = React.useContext(FilterContext);
+    const { filteredBuildings } = filterContext;
+    const { ownernames: ownernamesOptions, zip: zipOptions } = filterContext.filterOptions;
+    const { ownernames: ownernamesSelections, zip: zipSelections } = filterContext.filterSelections;
+
+    const [rsunitslatestActive, setRsunitslatestActive] = React.useState(false);
+    const updateRsunitslatest = () => {
+      setRsunitslatestActive(!rsunitslatestActive);
```
shakao, line 36: noting that we need telemetry/logging for all of these, but no worries if that's already scheduled for a future PR

client/src/components/PortfolioFilters.tsx
```diff
@@ -0,0 +1,394 @@
...
+          <Trans>MIN</Trans>
+        </label>
+        <label htmlFor="max-input">
+          <Trans>MAX</Trans>
+        </label>
+      </div>
+      {hasError ? (
+        <div className="alerts-container">
+          <Alert type="error" variant="primary" closeType="none">
+            Error
```
shakao, line 330: translation

client/src/components/PortfolioFilters.tsx
```diff
@@ -0,0 +1,394 @@
...
+          <Trans>Filters</Trans>:
+        </div>
+        <div className="filters-container">
+          <div className="filters">
+            <button
+              aria-pressed={rsunitslatestActive}
+              onClick={updateRsunitslatest}
+              className="filter-toggle"
+            >
+              <div className="checkbox">{rsunitslatestActive && <CheckIcon />}</div>
```
shakao, line 120: noting for future: we might want to look into making this a checkbox element for accessibility

shakao (approved):
scattered small comments, nothing big. looks awesome!!! 🎉

client/src/styles/_button.scss
```diff
@@ -140,3 +140,112 @@
...
+
+@mixin button-variant($color: $justfix-black) {
+  background-color: $color;
+
+  // mobile-eyebrow:
+  font-weight: 400;
+  font-style: normal;
+  letter-spacing: 0.01rem;
+  font-family: "Suisse Int'l Mono", "Courier New", Courier, monospace;
+  font-size: 1.4rem; // to 10rem
```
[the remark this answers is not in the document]
austensen: I think a left over reminder about the different root size in wow from when I was copying over styles from org site. will remove

client/src/styles/_button.scss
```diff
@@ -140,3 +140,112 @@
 .waou-survey-button:focus {
   opacity: 0.7;
 }
+
+// Copied from justfix-website
+
+@mixin button-variant($color: $justfix-black) {
+  background-color: $color;
+
+  // mobile-eyebrow:
```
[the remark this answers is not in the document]
austensen: yes, definitely - sorry forgot to mention, but I was planning to do that afterwards since styles have still been changing in figma. Some things are following the new system, but not completely. I still think it's worth adding them in like you suggested, and can just adjust as needed after.

client/src/components/Alert.tsx
austensen: I was writing this alert component in the style I had been doing for the component library so it can easily be copied over there and it seemed to need this for some of the storybook stuff.

client/src/components/Alert.tsx
austensen: oh cool, that sounds like a better option. I'll look into this more and try that out. thanks!

client/src/components/Multiselect.tsx
austensen: yes, thanks for catching that!

client/src/components/Multiselect.tsx
austensen: yeah, I think you're right, that would be better. I ended up needed to change a fair bit across multiple methods, but still better to do it that way. I'll plan to do come back and do this soon.

client/src/components/PortfolioTable.tsx
austensen: oh yeah, and that's probably worth adding a comment. Most of the corporations start with numbers (`123 fake st, LLC`), and after talking to Joel we thought it made sense to just put them last so there are some real person names in the list at first so it's less confusing.

client/src/components/PortfolioFilters.tsx
austensen: I was wondering that at first, but checking with tahnee my understanding is that it's purely aesthetic - the whole surrounding box is a button and clicking anywhere on it is supposed to trigger the toggle, so I was treating it like a toggle button (like from that [a11y toggle article](https://joshcollinsworth.com/blog/accessible-toggle-buttons) tahnee shared a while back)

client/src/components/PortfolioFilters.tsx
austensen: Thanks for catching this! Yeah, I'll add all of the telemetry in a separate PR, and listing all the ones you've pointed out.

client/src/components/PortfolioTable.tsx
austensen: oh yeah good idea!

client/src/components/PortfolioTable.tsx
austensen: I'm also not super familiar, but that was my understanding as well. I saw that this was always used in the tanstack table examples

client/src/components/PropertiesList.tsx
austensen: Since the RS units is just a toggle, we don't have any filter options (unlike landlord where we need the list of possible names for the multiselect), so since the selections and options are otherwise the same structure I just made the FilterValues type to reduce the copy/paste.

client/src/styles/PortfolioTable.scss
austensen: Yeah, this is specific to the min/max filter.

austensen:
I'm going to merge this in as is, and all of the issues raised in review will be addressed in future PRs coming soon.

PR #702, opened 2023-01-27
Add 2021 rent stab data

austensen:
We have scraped the 2022 DOF tax bills via [nyc-doffer](https://github.com/JustFixNYC/nyc-doffer), and generated the 2021 rent stab unit counts data. 
This has been [added to nycdb](https://github.com/nycdb/nycdb/pull/243).
Now we're updating wow tables to use that latest year of data. 

The updated version of the `rentstab_v2` nycdb table has already been updated in the demo database. Once these changes to WOW are pushed to master we can then merge in [this nycdb-k8s-loader PR](https://github.com/JustFixNYC/nycdb-k8s-loader/pull/125), and update the prod database with the same change. Then once that's done we'll be free to merge this wow PR from master to prod.


austensen:
[sc-11369]

dbtool.py
```diff
@@ -480,13 +491,9 @@ def exporttestdata(db: DbContext):
         oca_ssh_host=args.oca_ssh_host,
         oca_ssh_user=args.oca_ssh_user,
         oca_ssh_pkey=args.oca_ssh_pkey,
-        is_testing=args.use_test_data,
+        is_testing=False if cmd == "exporttestdata" else args.use_test_data,
```
shakao, line 494: just for my own clarification, why this exception?

shakao (approved):

dbtool.py
austensen: Existing documentation says that the exporttestdata requires a fully loaded version of the database (I think because we have some specific BBLs hardcoded into that function and our tests, and so it needs to have them all available). 

And so there isn't an argument for using test data on that command within dbtool.py, so I just put this exception in so that doesn't fail when it's missing. 

Maybe it'd be better to break up the if/else block for commands - putting exporttestdata case first, then set up this oca_config, then run through the other command cases where that argument will be present.

shakao: does args.use_test_data default to true? maybe we could have something where if it's none/undefined, we set it to false here?

austensen: After looking more closely at this I realized that `use_test_data` argument wasn't actually being used, since the `builddb` command is always the full data, and `loadtestdata` is always test data. So I now only create the `oca_config` when it's one of these commands, and it just assigns `use_test_data` appropriately and the argument is removed from the CLI.

PR #705, opened 2023-02-13
update legacy banner language

austensen:
In anticipation of portfolio filter release and deprecation of the legacy portfolio method, we are updating the site-wide banner to let folks know legacy version won't be available starting March 2023. 

[sc-11493]

### DIFF
--- a/client/src/containers/App.tsx
+++ b/client/src/containers/App.tsx
@@ -296,11 +296,19 @@ const WowzaBanner = withI18n()((props: withI18nProps) => {
     <div className={"App__banner " + (!isBannerOpen ? "d-hide" : "")}>
       <div className="content">
         {isLegacyPath(pathname) ? (
-          <Trans>You are viewing the old version of Who Owns What.</Trans>
+          <Trans>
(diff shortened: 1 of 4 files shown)


client/src/containers/App.tsx
```diff
@@ -296,11 +296,19 @@ const WowzaBanner = withI18n()((props: withI18nProps) => {
...
+            </ToggleLinkBetweenPortfolioMethods>
+          </Trans>
         ) : (
-          <Trans>You are viewing the new version of Who Owns What.</Trans>
-        )}{" "}
-        <ToggleLinkBetweenPortfolioMethods />
+          <Trans>
+            Starting March 2023{" "}
+            <ToggleLinkBetweenPortfolioMethods>the old version</ToggleLinkBetweenPortfolioMethods>{" "}
+            of Who Owns What will no longer be available.
```
shakao, line 309: small: remove period at the end or add period to "Switch to the new version" so they match?

austensen: great catch - thanks!

PR #707, opened 2023-02-15
Add placeholder legacy deprecation modal hooked up to environment variable and cookie

shakao:
getting the component and plumbing set up so we just need to fill in content and localize when we launch

[sc-11497]


austensen (approved):
Looks perfect, thanks!

PR #717, opened 2023-03-02
reorg design system typography, refactor filter styles

austensen:
Moved the new design system typography out of _vars, and used those mixins/variables where possible (we're still not using much of the new system)

This PR also includes a refactor of the portfolio filter styles, simplifying things now that the markup is mostly finalized and to make things easier to adapt for mobile.

### DIFF
--- a/client/src/styles/_button.scss
+++ b/client/src/styles/_button.scss
@@ -189,7 +189,6 @@
     transform: translateX(0.7rem);
   }
 
-  // TODO: review in design system - current figma focus state is too subtle
   &:focus-visible {
     box-shadow: none;
     outline: 2px solid #727e96;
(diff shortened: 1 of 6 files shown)


client/src/styles/_button.scss
```diff
@@ -195,7 +189,6 @@
     transform: translateX(0.7rem);
   }
 
-  // TODO: review in design system - current figma focus state is too subtle
   &:focus-visible {
     box-shadow: none;
     outline: 2px solid #727e96;
```
shakao, line 194:
```suggestion
    outline: 2px solid $focus-outline-color;
```

shakao (approved):

PR #718, opened 2023-03-08
Filters mobile

austensen:
This PR adds support for mobile to the filter bar. Makes the basic changes to the styles of filter components and the bar, plus adds the fullscreen wrapper of all filters - and the scroll snapping and shadow effects. Plus a small fix to the handling of min-max errors, and its default values I caught while in here.

Also included is an update to the filter input styling on desktop and mobile so they read more clearly as inputs (not full width, white background)

Not included is the mobile-specific behavior of applying in-progress filters when the user clicks outside of that filter's dropdown. It seems like this will require some more substantial reorganization of how/where that state is managed to allow for this. I'll make a separate PR for this once I figure things out.

[sc-11706]
[sc-11738]

### DIFF
--- a/client/src/components/Multiselect.tsx
+++ b/client/src/components/Multiselect.tsx
@@ -612,19 +612,23 @@ export class Multiselect extends React.Component<IMultiselectProps, any> {
             <Trans>Error</Trans>
           </Alert>
         )}
-        {!hideSelectedList && this.renderSelectedList()}
-        {selectedValues.length > 0 && (
-          <button
-            className="clear-all button is-text"
-            key={2}
-            onClick={() => {
-              this.resetSelectedValues();
-              onApply([]);
-            }}
-          >
-            <Trans>Clear</Trans>
-          </button>
-        )}
+        <div className="selectedListContainer">
+          {!hideSelectedList && this.renderSelectedList()}
+        </div>
+        <div className="clear-all-container">
+          {selectedValues.length > 0 && (
+            <button
+              className="clear-all button is-text"
+              key={2}
+              onClick={() => {
+                this.resetSelectedValues();
+                onApply([]);
+              }}
+            >
+              <Trans>Clear</Trans>
+            </button>
+          )}
+        </div>
         <div
           className={classnames("search-wrapper searchWrapper", { hasError: hasError })}
           ref={this.searchWrapper}
--- a/client/src/components/Multiselect.tsx
+++ b/client/src/components/Multiselect.tsx
@@ -30,6 +30,7 @@ export interface IMultiselectProps {
   onSearch?: (value: string) => void;
   onKeyPressFn?: (event: any, value: string) => void;
   onApply: (selectedList: any) => void;
+  onFocusInput: () => void;
   infoAlert?: React.ReactNode;
   caseSensitiveSearch?: boolean;
   id?: string;
@@ -563,6 +564,7 @@ export class Multiselect extends React.Component<IMultiselectProps, any> {
     } else {
       this.toggelOptionList();
     }
+    this.props.onFocusInput();
   }
 
   onBlur() {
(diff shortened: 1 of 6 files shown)


client/src/components/PortfolioFilters.tsx
```diff
@@ -101,92 +104,111 @@ export const PortfolioFilters = React.memo(
...
+              selectedValues={zipSelections}
+              displayValue="name"
+              placeholder={i18n._(t`Search`) + `... (${zipOptions.length})`}
+              onApply={onZipApply}
+              onFocusInput={() => helpers.scrollToBottom(".mobile-wrapper-dropdown")}
+              avoidHighlightFirstOption={true}
+            />
+          </FilterAccordion>
+        </FiltersWrapper>
+        {[rsunitslatestActive, ownernamesActive, unitsresActive, zipActive].includes(true) && (
```
shakao, line 196: could this just be a big "or" statement? `(rsunitslatesActive || ownernamesActive || ...)` probably not a huge diff but might be marginally more performant than the includes

client/src/components/PortfolioFilters.tsx
```diff
@@ -234,6 +256,60 @@ export const PortfolioFilters = React.memo(
...
+              <span className="active-filter-count">{activeFilters}</span>
+            )}
+            {isOpen ? <CloseIcon className="closeIcon" /> : <ChevronIcon className="chevronIcon" />}
+          </summary>
+          <div className="dropdown-container scroll-gradient mobile-wrapper-dropdown">
+            {children}
+            {!!activeFilters && (
+              <button onClick={() => setIsOpen(!isOpen)} className="button is-primary">
+                <Trans>View Results</Trans>
+                {resultsCount && <span className="view-results-count">{resultsCount}</span>}
```
shakao, line 303: this will evalutate to false when resultsCount is zero, do we want to show the "0" ?

shakao (approved):

client/src/components/PortfolioFilters.tsx
austensen: oh right, thanks for catching!

client/src/components/PortfolioFilters.tsx
austensen: ah good to know, yeah that is just simplier all around not sure what I was doing here.

client/src/components/PortfolioFilters.tsx
```diff
@@ -101,92 +104,111 @@ export const PortfolioFilters = React.memo(
...
-              onClick={updateRsunitslatest}
-              className="filter filter-toggle"
-            >
-              <div className="checkbox">{rsunitslatestActive && <CheckIcon />}</div>
+        <FiltersWrapper
+          isMobile={isMobile}
+          activeFilters={activeFilters}
+          resultsCount={filteredBuildings}
+        >
+          {" "}
```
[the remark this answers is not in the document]
austensen: thanks yeah this shouldn't be there!

client/src/components/PortfolioFilters.tsx
```diff
@@ -101,92 +104,111 @@ export const PortfolioFilters = React.memo(
...
     };
 
+    const activeFilters = [rsunitslatestActive, ownernamesActive, unitsresActive, zipActive].filter(
+      Boolean
+    ).length;
+
     return (
       <div className="PortfolioFilters" ref={ref}>
-        <div className="filter-for">
+        <div className="new-container">
```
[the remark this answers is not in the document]
austensen: ah yeah good point, I changed this one for now, but will later do a pass over them all

client/src/components/Multiselect.tsx
```diff
@@ -30,6 +30,7 @@ export interface IMultiselectProps {
   onSearch?: (value: string) => void;
   onKeyPressFn?: (event: any, value: string) => void;
   onApply: (selectedList: any) => void;
+  onFocusInput: () => void;
```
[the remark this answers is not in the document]
austensen: right yeah, should be optional

PR #720, opened 2023-03-11
add "toast" notifications on results

austensen:
This PR adds the "toast" notifications that pop up when results are updated. 

[sc-11711]

They slide up from the bottom, stay for 5 seconds, then slide out. If you hover on them it pauses the animation. They have [`role="status"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role) so they're read out on screen readers (after learning about this I also went through and added this role to other alerts within filters that made sense). 

image

In addition to the toasts, there is an in-line alert for when there are 0 results. 

image

On mobile, they only appear after you close the fullscreen filters menu, and so if you apply multiple filters within one trip into the filter menu then multiple toasts pop up at once. There is a desired priority given by Design and this is given just by the ordering of the html elements, and following Shannons comment these are now in a container and via css all but the first are not displayed.

### DIFF
--- a/client/src/components/PortfolioFilters.tsx
+++ b/client/src/components/PortfolioFilters.tsx
@@ -190,9 +190,10 @@ export const PortfolioFilters = React.memo(
             />
           </FilterAccordion>
         </FiltersWrapper>
+
         {(rsunitslatestActive || ownernamesActive || unitsresActive || zipActive) && (
-          <>
-            <div className="filter-status">
+          <div className="filter-status">
+            <div className="filter-status-info">
               <span className="results-count" role="status">
                 <Trans>
                   Showing {filteredBuildings || 0}{" "}
@@ -207,7 +208,7 @@ export const PortfolioFilters = React.memo(
               </button>
             </div>
             {filteredBuildings === 0 ? ZeroResultsAlert : <></>}
-          </>
+          </div>
         )}
 
         <Modal key={1} showModal={showInfoModal} width={20} onClose={() => setShowInfoModal(false)}>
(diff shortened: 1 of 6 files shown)


client/src/components/PortfolioFilters.tsx
```diff
@@ -255,67 +257,139 @@ export const PortfolioFilters = React.memo(
...
-            {!!activeFilters && (
-              <button onClick={() => setIsOpen(!isOpen)} className="button is-primary">
-                <Trans>View Results</Trans>
-                {resultsCount != null && <span className="view-results-count">{resultsCount}</span>}
-              </button>
-            )}
+          <div className="filters">
+            <details
+              className={classnames("filter filter-accordion filters-mobile-wrapper", {
+                active: !!numActiveFilters,
```
shakao, line 293: small: maybe `numActiveFilters > 0` for clarity?

client/src/components/PortfolioFilters.tsx
```diff
@@ -255,67 +257,139 @@ export const PortfolioFilters = React.memo(
...
+    <Trans>
+      Expand the the Owner/Manager column in Table view to see all contacts associated with that
+      building.
+    </Trans>
+  </Alert>
+);
+
+const RsUnitsToastAlert = (
+  <Alert
+    className="ownernames-toast-alert filter-toast-alert"
```
shakao, line 360: should the class for the RsUnits still be `ownernames-toast-alert` ? maybe this class could be removed altogether (here & in the above component) if it's not used in the CSS?

shakao (approved):
hmm you could try having toasts be `display: block` `position: relative` inside a container that was `position: fixed` to the bottom of the screen. i think if the parent container is fixed but the toasts are not, they would stack on top of each other without having to do the layout computation

austensen:
Thanks so much for that suggestion - that makes much more sense. After trying that out I also realized I would also be able to use `:not(:first-child) {display:none}` so that only the first (higher priority) alert pops up in those situations.

PR #721, opened 2023-03-16
focus inputs on opening filter accordions

austensen:
When filter accordions are opened the focus is initially set to the filter input (search bar for multiselect and min input for min-max select)

The multiselect component already uses the component ID to create an id for the search bar input, so I followed the same pattern for the min-max component.

[sc-11770]


shakao:
does the focus trap element take a css selector? wondering if you did something more general like "input", if it would still select the right element since there's usually only one child input (and the "min" input is the first of it's kind)

shakao (approved):
lgtm overall though!

austensen:
> does the focus trap element take a css selector? wondering if you did something more general like "input", if it would still select the right element since there's usually only one child input (and the "min" input is the first of it's kind)

It does take a string css selector, but unfortunately it doesn't scope it within the bounds of the focus-trap, so just putting "input" just goes to the first one on the page.

PR #725, opened 2023-03-21
Add table edge blur when horizontal scroll available

austensen:
We have had a subtle white blur effect on the side of the portfolio table when there is space available for horizontal scroll. When rewriting all the table code for the new react-table version this broke and now needs to be added back. I've copied over the css that was used previously, but it's not working quite right yet - it's getting stuck where the edge of the screen is on load and doesn't stay with that edge as it scrolls. oh and we also need to change the white color to our `$justfix-white` color. 

image

[sc-11604]


client/src/styles/PortfolioTable.scss
```diff
@@ -6,226 +6,270 @@ $table-border-light: 1px solid $gray-light;
...
+    position: relative;
+    overflow: auto;
 
-    tr {
-      padding: 0px 5px;
+    // Add right-side fade out when list is scrollable:
+    &::after {
+      content: "";
+      display: block;
+      position: fixed;
```
kiwansim, line 22: this was the fix!

austensen:
This looks perfect! Thanks so much for figuring it out! (I can't approve since I opened it, but you should be able to)

kiwansim (approved):

PR #726, opened 2023-03-27
multiselect dropdown - add checkboxes, keep searchbar text, keep selected options in list

austensen:
A few changes related to the multiselect component:

* The dropdown options are changed so that selected options aren't removed from the list. Instead they have a checkbox to show if they are selected. And for keyboard navigation/hover they are still underlined. 
* The component is improved for tab navigation/screen readers, but it's still not working on all platforms, and a follow up PR will address that. For now this makes sure that escape closes the option list returning to the searchbar (so you don't get cught in the often super long lists). The screen reader experience is working mostly in Chrome/OSX, but not on Windows. 
* The Multiselect component styles are cleaned up so that it's all nested under the component class
* The font size for the searchbar input is increased to 16px to prevent auto-zoom on focus for mobile
* Search bar height increased for mobile
* Search bar inputs are no longer cleared when an option is selected
* I've also cut the "grouby" option for multiselect, which we aren't using just to simplify that component a bit.

image
image

[sc-11701]
[sc-11808]
[sc-11864]
[sc-11946]
[sc-11687]

### DIFF
--- a/client/src/components/Multiselect.tsx
+++ b/client/src/components/Multiselect.tsx
@@ -343,13 +343,7 @@ export class Multiselect extends React.Component<IMultiselectProps, any> {
   }
 
   renderOptionList() {
-    const {
-      id,
-      style,
-      emptyRecordMsg,
-      loading,
-      loadingMessage = "loading...",
-    } = this.props;
+    const { id, style, emptyRecordMsg, loading, loadingMessage = "loading..." } = this.props;
     const { options, highlightOption } = this.state;
     if (loading) {
       return (
--- a/client/src/components/Multiselect.tsx
+++ b/client/src/components/Multiselect.tsx
@@ -20,7 +20,6 @@ export interface IMultiselectProps {
   showCheckbox?: boolean;
   selectionLimit?: any;
   placeholder?: string;
-  groupBy?: string;
   loading?: boolean;
   style?: object;
   emptyRecordMsg?: string;
@@ -66,7 +65,6 @@ export class Multiselect extends React.Component<IMultiselectProps, any> {
       highlightOption: props.avoidHighlightFirstOption ? -1 : 0,
       showCheckbox: props.showCheckbox,
       keepSearchTerm: props.keepSearchTerm,
-      groupedObject: [],
       hasError: false,
     };
     // @ts-ignore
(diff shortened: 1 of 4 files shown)


client/src/styles/Multiselect.scss
```diff
@@ -1,128 +1,169 @@
...
+              width: 0.5rem;
+              height: 1rem;
+              border: solid $justfix-white;
+              border-width: 0 0.2rem 0.2rem 0;
+              transform: rotate(45deg);
+            }
+          }
+          span {
+            align-self: center;
+          }
```
austensen, line 131: This is the only section that was actually changed, everything else was just nested under the `multiselectContainer` class.

client/src/components/Multiselect.tsx
```diff
@@ -380,85 +358,70 @@ export class Multiselect extends React.Component<IMultiselectProps, any> {
...
-            isObject ? option[displayValue] : (option || "").toString(),
-            option
+            <input
+              id={`option-check-${i}`}
+              type="checkbox"
+              readOnly
+              className="checkbox"
+              checked={isSelected}
+              onKeyDown={this.onEscKeyCloseOptionList}
+            />
```
kiwansim, line 407:
```suggestion
              aria-labelledby={LABEL_ID}
```

client/src/components/Multiselect.tsx
```diff
@@ -380,85 +358,70 @@ export class Multiselect extends React.Component<IMultiselectProps, any> {
...
           )}
+          <span>
+            {this.props.optionValueDecorator(
+              isObject ? option[displayValue] : (option || "").toString(),
+              option
+            )}
+          </span>
```
kiwansim, line 414:
```suggestion
          <label id={LABEL_ID} for={`option-check-${i}`}>
            {this.props.optionValueDecorator(
              isObject ? option[displayValue] : (option || "").toString(),
              option
            )}
          </label>
```

kiwansim: i think any interactive button with labels should have an explicit label tag! 

I also thought perhaps this would fix the screenreader reading "BLANK" for each checkbox option, but changing to semantic html alone didn't fix our issue.

austensen: ah ok, thanks! I wasn't sure it needed it because it was reading the text of the `li` and I was worried it would then read that twice. 
yeah I think the "blank" is actually it just announcing that the search box is empty every time you arrowkey navigate through the option list because that textbox is supposed to stay active (focused, targeted? not sure the word) so you can type to narrow it more at any time. 
Really am regretting picking this multiselect library initially, and not testing it out for accessibility at the start - won't do that again!

kiwansim (approved):
looks good to merge! noting that we will address screen reader accessibility in a different PR

client/src/components/Multiselect.tsx
austensen: no I think you're right, it probably should be there, since it wasn't reading it properly on browserstack/windows, so maybe that was just a quirk of OSC screen reader.

PR #727, opened 2023-03-27
fix RS units toggle height and hover on mobile

austensen:
Fixes a couple small bugs for the RS units toggle on mobile:
* It was getting squished when the other filter accordions were expanded
* It was staying underlined after toggling it off (mobile hover state)

[sc-11939]
[sc-11926]


kiwansim (approved):
looks good to me!

PR #728, opened 2023-03-29
add council district to legacy SQL search function

austensen:
When we added city council district for the portfolio table we added it to the `wow_bldgs` table (#687) and API data types (#688), but I forgot to also add it to the legacy method search function.


austensen:
Yeah, I noticed this because the portfolio table was breaking when I switched to legacy view because this column was missing. We only added council district recently as part of adding the portfolio filters, and since they are only going to be available on the new version I wasn't thinking about it but we'll want the portfolio table columns to be consistent even if the filters aren't available.

PR #730, opened 2023-03-30
Show portfolio filters on new WOW version only

austensen:
The new portfolio filters should only be available on the new version of wow - so this hides the filter bar on legacy.

Also, we changed the way links to other properties in the portfolio table works recently for filters, but they all went to the new version. This fixes that so they match the current version users are on. 

[sc-11670]


kiwansim (approved):
neat ✨

PR #732, opened 2023-03-30
sticky header shadow, slide in, scrolling, hide background page

austensen:
Various QA fixes for mobile filter menu

* Slide up to fill screen
* Sticky header shadow - smaller/darker, only on top
* Hide page from showing through around rounded corners
  * was using `outline` for this, but it doesn't follow border-radius on safari. plus now it needs to be animated with slide
* remove `max-height` to prevent internal scrolling for nested filer accordions, so whole menu scrolls instead
* this should also fix the buggy scrolling when there are active selections

https://user-images.githubusercontent.com/16906516/228920984-c226451d-89eb-499f-a080-ea2cbcf449b1.mov

* Also, reduce the width of the landlord search bar/dropdown to leave room for scrolling.  
image

[sc-11993]
[sc-11875]
[sc-11879]
[sc-12095]

### DIFF
--- a/client/src/styles/PortfolioFilters.scss
+++ b/client/src/styles/PortfolioFilters.scss
@@ -370,17 +370,63 @@
       &[open] {
         --filter-bar-top-height: 4.6rem;
         --filter-bar-height: 6.6rem;
-        position: absolute;
+        position: fixed;
         top: 0;
         left: 0;
         right: 0;
+        z-index: 101;
         height: var(--filter-bar-top-height);
         background-color: $justfix-table-grey;
         border-radius: 0.8rem 0.8rem 0 0;
         box-sizing: border-box;
         border: none;
         border-bottom: 0.2rem solid $justfix-grey-700;
-        outline: 1rem solid $justfix-grey-700;
+
+        @keyframes slide-up {
+          0% {
+            transform: translateY(100vh);
+          }
+          100% {
+            transform: translateY(0);
+          }
+        }
+
+        animation-name: slide-up;
+        animation-duration: 0.5s;
+        animation-timing-function: ease-in-out;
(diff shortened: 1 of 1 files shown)


kiwansim:
really appreciated your concise explanation of what the fixes for each of these buggy behaviors in the pr description! will be a good for future reference and it also helped with guiding the code review mentally. i'll try to do this as well if you think it would be helpful to you!

kiwansim (approved):

client/src/styles/PortfolioFilters.scss
```diff
@@ -370,17 +370,63 @@
       &[open] {
         --filter-bar-top-height: 4.6rem;
         --filter-bar-height: 6.6rem;
-        position: absolute;
+        position: fixed;
         top: 0;
         left: 0;
         right: 0;
+        z-index: 101;
```
kiwansim, line 377: lol is this z-index 101 because the vertical gradient at the right side of the table is z-index 100???

austensen: lol yea, not sure exactly how things got to 100, but yeah otherwise it blurs through. I'll make a ticket to reset all of these z-indexes to something more logical

kiwansim: from my quick reading into z-indices it seems like anything that isn't 0 or -1 is discouraged because of how it explicitly sets the tab order, so 100 / 101 aren't bad values to me personally! was just wondering for the giggles

austensen: oh interesting, I hadn't thought about the effect on tab order, that's good to know. yeah I was about to go deep into z-index and "stacking context" for something in here but found a better solution so have saved that for later.

PR #734, opened 2023-04-03
only auto-focus filter inputs on desktop

austensen:
Previously we made all the multiselect filter accordions automatically focus in the search bar input on opening (#721). This PR removes that for mobile only. 

[sc-12092]


kiwansim (approved):
👍🏽

PR #740, opened 2023-04-05
Font weight and spacing cross browser consistency

austensen:
There has been a very noticeable difference in font-weight and letter-spacing across browsers. 

**Chrome**: already pretty much as intended.
**Safari** and **Firefox**: both the font-weight appeared much higher, and the letter spacing was larger. 

I made sure that we have the specific weights specified when loading the font files. I also added `-webkit-font-smoothing: antialiased`, and this solves the weight issue in Safari, but if applied in Chrome it makes it too bold, and does nothing for Firefox. This also didn't help the letter spacing. 

I kept trying to solve this with these sorts of browser-independent fixes, but have just not been able to find something that solves all of them. 

So I've tried to instead use JS to figure out the browser for the user-agent on start up and store that in a data attribute in the html. Then in SASS we can use a mixin to give browser-specific styles, like we do for screen sizes. 

I know it's not best practice to try to infer browser from user agent to change styles, and but since all the global fixes I could find didn't work, and it's not a situation where we could specifically test for availability of some css property, I can't think of how else to deal with his. Also, since the changes are not critical, the risk seems low. 

I use these browser-specific mixins to adjust the font weight and letter spacing for the remaining issues. 

Before:

* chrome: 
image
* firefox: 
image
* safari: 
image

After:

* chrome: 
image
* firefox: (looks a bit different since this one I was using browserstack)
image
* safari: 
image

### DIFF
--- a/client/src/styles/PortfolioTable.scss
+++ b/client/src/styles/PortfolioTable.scss
@@ -105,6 +105,7 @@ $table-border-dark: 1px solid $dark;
         position: sticky;
         top: 0;
         z-index: 50;
+        white-space: nowrap;
 
         tr {
           height: 30px;
@@ -113,7 +114,7 @@ $table-border-dark: 1px solid $dark;
         tr:first-child th {
           border-bottom: 1px solid $table-border-light;
           border-top: 1px solid $gray-dark;
-          font-weight: bold;
+          font-weight: 600;
           color: #000;
           background-color: $justfix-table-white;
         }
--- a/client/src/styles/PortfolioTable.scss
+++ b/client/src/styles/PortfolioTable.scss
@@ -5,7 +5,7 @@
 $table-border-light: 1px solid $gray-light;
 $table-border-dark: 1px solid $dark;
 
-.PortfolioTable {
+#PortfolioTable {
   overflow: auto;
   position: relative;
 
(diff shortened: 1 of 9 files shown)


client/src/styles/PortfolioTable.scss
```diff
@@ -105,15 +111,21 @@ $table-border-dark: 1px solid $dark;
         position: sticky;
         top: 0;
         z-index: 50;
+        white-space: nowrap;
 
         tr {
           height: 30px;
+          div {
+            font-weight: 600 !important;
```
kiwansim, line 119: is this just taking the place of `font-weight: bold` that was removed from `$gray-dark`?

client/src/styles/_typography.scss
```diff
@@ -5,7 +5,7 @@
 
 @font-face {
   font-family: "Degular";
-  font-weight: normal;
```
kiwansim, line 8: by any chance, did explicitly defining the weight here fix the `NEW` pill font being weird?

kiwansim (approved):
looks good to me 🏅

client/src/styles/PortfolioTable.scss
austensen: Yeah, it was suggested to use the number rather than names, but then I also realized that there was some default styling (I think coming from the spectre.css we use, or maybe just general defaults) that was making it bold, and adding the specificity by included the nested div that are in each cell helped override that)

kiwansim: oh cool, that's really good to know for future reference!

client/src/styles/_typography.scss
austensen: no sadly it doesn't seem to help

kiwansim: Screen Shot 2023-04-06 at 2 45 51 PM

PR #743, opened 2023-04-09
Multiselect rewrite

austensen:
The original multiselect library I started with for filters has turned out to be not that great - it was not originally written in typescript, it's difficult to customize, and it's not accessible for screen readers. So after struggling to add screen-reader accessibility to the existing component, I've just decided to use the much better [`react-select`](https://react-select.com/) library and rewrite our custom multiselect filter. 

`react-select` is written in typescript, has great accessibility features (nicely handles keyboard navigation for search-bar/option-list/selected-list, and has great default screen reader setting plus customization options), and it has a nice system for customizing by overwriting and recomposing its subcomponents. (The entire `MultiSelect.tsx` file is new, so you can just review that file alone and ignore the diff)

Since I then needed to update a lot of the filter styles for the new component, I ended up doing a much needed refactor of all the PortfolioFilters styles. I pulled out the minmax-select and multi-select styles into their one contained component styles, and also tried to simplify the organization of desktop/mobile styles to keep things close together (rather than the mostly separate mobile-specific styles nested in it's own section. I also tried to pull other changes out to component styles (Alerts, Buttons), and address some other issues at the same time.

While I was separating out the minmax-select component styles I decided to pull the component into it's own file as well (the contents haven't changed).

This all addresses a few outstanding tickets:
* multiselect now works with screen-readers [sc-12099]
* outline replaced with border for elements with border-radius [sc-11773]
* fixes multiselect searchbar placeholder text alignment [sc-12186]
* center align apply buttons on mobile [sc-12060]


client/src/components/PortfolioFilters.tsx
```diff
@@ -142,22 +152,18 @@ export const PortfolioFilters = React.memo(
             isActive={ownernamesActive}
             isOpen={ownernamesIsOpen}
             setIsOpen={setOwnernamesIsOpen}
-            initialFocus={isMobile ? undefined : "#filter-ownernames-multiselect_input"}
+            initialFocus={isMobile ? undefined : "#filter-ownernames-multiselect input"}
             selectionsCount={filterContext.filterSelections.ownernames.length}
             className="ownernames-accordion"
+            i18n={i18n}
```
kiwansim, line 158: could the i18n just be imported in the `FilterAccordion` file rather than being passed as a prop? same with `MultiSelect` component below

client/src/styles/_alert.scss
```diff
@@ -15,27 +16,34 @@
...
+    padding-top: 1rem;
+    &:last-child {
+      padding-right: 1rem;
+    }
   }
 
-  button.jf-alert__close {
+  .jf-alert__close {
     align-self: flex-start;
+    // TODO: target area of button is still less than 44x44 px for a11y
```
kiwansim, line 37: is there a ticket for this? if so i'd err on the side of not leaving in `TODO` comments

client/src/components/Multiselect.tsx
```diff
@@ -1,712 +1,348 @@
...
-  matchValues(value, search) {
-    if (this.props.caseSensitiveSearch) {
-      return value.indexOf(search) > -1;
-    }
-    if (value.toLowerCase) {
-      return value.toLowerCase().indexOf(search.toLowerCase()) > -1;
-    }
-    return value.toString().indexOf(search) > -1;
-  }
+const ariaLiveGuidanceGeneral = `Type to refine list of options. Use Up and Down to choose options,
```
kiwansim, line 42: didn't know this was a thing!! 😮

client/src/components/Multiselect.tsx
```diff
@@ -1,712 +1,348 @@
...
-    const { onRemove, showCheckbox, displayValue, isObject } = this.props;
-    if (isObject) {
-      index = selectedValues.findIndex((i) => i[displayValue] === item[displayValue]);
+    },
+    [onInputChange]
+  );
+
+  // Passing custom prop to make accessible to child compnents via selectProps
+  const removeValue = useCallback(
+    // @ts-ignore (TODO: says property 'value' doesn't exist on type Option)
```
kiwansim, line 95: hm, can we extend the type Option to include the property `value`? kind of confused about this error

client/src/components/Multiselect.tsx
```diff
@@ -1,712 +1,348 @@
...
-        aria-activedescendant={`${id || "option"}-item-${highlightOption}`}
+      <components.MultiValue
+        getValue={getValue}
+        {...props}
+        components={{
+          Container: components.MultiValueContainer,
+          Label: components.MultiValueLabel,
+          Remove: CustomMultiValueRemove,
+        }}
+        // TODO: Can't know when it's "focused" via arrow key navigation, it's not using ::focus,
```
kiwansim, line 198: thoughts on making github issues for the things that aren't breaking errors but would be nice to fix? unsure if that's a good use of our time but i dont know if these non-breaking errors are being kept track of anywhere. they don't feel urgent enough to be addressed right now in VQA, but I think we do need a visible log of them

kiwansim (approved):
this looks so good! tested with browserstack and looks good to me. i just have some comments that aren't going to affect functionality regardless of what you do with my thoughts! going to approve so that this merge isn't blocking the push to demo site and further VQA

client/src/styles/_alert.scss
austensen: yeah good point - added a shortcut card to the design system collection and will remove this and others.

client/src/components/Multiselect.tsx
austensen: Yeah I don't understand it either, because the `Option` type has value, and it knows it has this type, but for some reason it's doesn't know what the properties are. 

image

At the top of this files it's defined as:

```
export type Option = {
  value: string;
  label: string;
  [key: string]: string;
};
```

PR #744, opened 2023-04-10
add portfolio filter info alert on overview map

austensen:
Adds a new info alert on the Overview page map prompting users to use the portfolio filters when the portfolio has >= 15 properties. This also restyles the "big portfolio" warning alert to look the same. Both alerts stay closed for the session.

image
image

[sc-11997]


client/src/components/PortfolioAlerts.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+type PortfolioAlertProps = Omit<AlertProps, "children"> & {
+  portfolioSize: number;
+};
+
+type FilterPortfolioAlertProps = PortfolioAlertProps & {
+  addressPageRoutes: AddressPageRoutes;
+};
+
+export const BigPortfolioAlert = ({ portfolioSize, className, ...props }: PortfolioAlertProps) => {
```
kiwansim, line 23: alerts would be an excellent candidate for making into a reusable class component in storybook later! esp since these exact patterns also exist in LOC rebrand

kiwansim (approved):
👍🏽

client/src/components/PortfolioAlerts.tsx
austensen: Yeah for sure! That was the next one I was hoping to add, it should be mostly ready to copy over from here and just the styles need to be updated.

austensen:
[sc-12226]

PR #745, opened 2023-04-10
Filter toast alerts

austensen:
This PR changes the placement of the alerts that appear once particular filters are applied. This were first added in #720 as "toast" notifications, that pop up for a few seconds at the bottom of the screen. Design decided to change the placement of these to show up in the place of the first table row. For accessibility the alerts still have `role="status"` so they are read out ("polite") once they appear so you get the info read out right after the number of results without having to navigate around in the table. And when one of the alerts is closed it won't reappear for the session. Multiple alerts stack. The copy for each of these is also updated for the final content decisions. 

image
image

[sc-11895]


kiwansim (approved):

austensen:
[sc-12229]

PR #746, opened 2023-04-11
owner filter info modal pauses accordion focus trap

austensen:
This PR fixes an issue with the landlord filter accordion that when the info modal open focus stays trapped in the accordion. In a newer version of `react-focus-trap` (I had to update the package) there is a `pause` prop to deal with this sort of situation. However this required that I have access to the open/closed state of the modal, so instead of passing the callback from the parent, we pass the contents of the modal and create it within the accordion component. There are also some copy changes. Also, seeing that the text for the info button is now custom for the filter topic (not the generic "what's this") I added that as a prop for the filter accordion.  

Later we'll want to add an analytics event for when people open this info modal, but I'm planning to do a separate PR with all the new analytics stuff.

https://user-images.githubusercontent.com/16906516/231035921-9309d2c8-2f6a-4782-a490-3fb8c71ab418.mov

[sc-11823]


client/src/components/PortfolioFilters.tsx
```diff
@@ -142,8 +169,10 @@ export const PortfolioFilters = React.memo(
           </button>
           <FilterAccordion
             title={i18n._(t`Landlord`)}
-            subtitle={i18n._(t`Officer/Owner`)}
-            infoOnClick={() => setShowOwnerModal(true)}
+            subtitle={i18n._(t`Person/Entity`)}
+            infoLabel={i18n._(t`Who are they?`)}
+            // TODO: add analytics event for onInfoClick
```
kiwansim, line 174: only request is to move this todo into the outstanding shortcut cards (along w/ other todos)

kiwansim (approved):
🚀

PR #749, opened 2023-04-12
Minmax select refactor and add presets

austensen:
A big set of changes to our building size filter. 
* The data structure of the options/selections ranges is changed from `[number, number]` to a more sensible [`{min: number, max: number}]`, (an array because we can no have multiple ranges - see below). Also instead of using `undefined` as the default value, it uses `-Infinity`/`Infinity` (the undefined were changed to this anyway in the table filter application)
* The error handling for the min/max inputs also uses a similar data structure (plus a msg field) for better readability 
* There are now a set of preset unit ranges to choose from (can be multiple), or you can expand a dropdown to reveal the original min/max inputs to define a custom range. When you enter anything in the custom inputs it disables the presets (but remembers your selections if you then clear the inputs). There's still the one apply button for both options. 
* There are also a lot of a11y fixes in here: input options following [these suggestions](https://www.magentaa11y.com/checklist-web/number-input/), proper labels and descriptions for screen readers (including the new presets), focus states, alert message. 

https://user-images.githubusercontent.com/16906516/231514510-dbbec3f5-d12b-4888-b887-4fdedbb5fa8e.mov

https://user-images.githubusercontent.com/16906516/231514889-fbab4d86-9eda-4266-8526-7138c045b35f.mov

[sc-11742]
sc-12103]
[sc-12197]
[sc-11988]


client/src/components/MinMaxSelect.tsx
```diff
@@ -1,89 +1,193 @@
...
+import React, { InputHTMLAttributes } from "react";
+import { Trans, t } from "@lingui/macro";
+import { i18n } from "@lingui/core";
 import classnames from "classnames";
 import helpers from "util/helpers";
 import { Alert } from "./Alert";
-import { FilterNumberRange, MINMAX_DEFAULT } from "./PropertiesList";
+import { FilterNumberRange, NUMBER_RANGE_DEFAULT } from "./PropertiesList";
+
+type MinMaxErrors = { min: boolean; max: boolean; msg: JSX.Element | undefined };
```
kiwansim, line 9: nit: I think the current variable names get harder to read when we start referencing the values like `MinMaxErrors.min || MinMaxErrors.max`. but this is also personal preference so truly take it or leave it!

```suggestion
type RangeValidation = { isMinValid: boolean; isMaxValid: boolean; errorMsg: JSX.Element | undefined };
```

kiwansim (approved):
yo this looks _so so good_! it's much cleaner and the default ranges are such a good UX improvement addition! no more decision fatigue (that i didn't even know I had) :)

client/src/components/MinMaxSelect.tsx
austensen: oh yeah great point! Just updated for this change - thanks!

austensen:
> yo this looks _so so good_! it's much cleaner and the default ranges are such a good UX improvement addition! no more decision fatigue (that i didn't even know I had) :)

Thanks so much!! Was definitely satisfying to get this all into better shape. And thanks for the original suggestion to change this up - it also made it a lot easier to adjust to the new requirements of the presets

PR #750, opened 2023-04-13
reduce filter accordion info text button to 12px

austensen:
* Reduce the LL Filter tooltip copy from 14px down to be 12px. [sc-12222]


kiwansim (approved):
LGTM!

PR #751, opened 2023-04-26
add extra params for GTM on portfolio events

austensen:
Previously we weren't sending any extra data along with our google analytics events in WOW, we only did for Amplitude. This PR adds the same extra properties along with our events for portfolio table column sorts and address clicks. I'm doing just these two now so they can go to prod and I can make sure everything works correctly on GTM for updating the custom variables/tags for the events. After that there is another filters-specific PR to follow that sets up all the new events for filters.


austensen:
For some reason the build failed when installing node, but [it worked on retry](https://app.netlify.com/sites/wow-django-dev/deploys/64491e6d18b0b478801e5d00).

kiwansim (approved):
would love to hear back on whether the events are being logged as expected in GTM once this is pushed to prod

PR #752, opened 2023-04-26
add analytics events for filters

austensen:
This PR adds a collection of analytics events - some are already in wow but needed to be added back after the table was completely rewritten for filters, and others are brand new to cover the filters interactions. All of the events have parity between Google and Amplitude, including the custom properties we send along with each (eg. portfolioSize, portfolioColumn, etc.). [sc-10784]

The way I'm actually logging the events in kind of awkward, because we have some common extra properties to send with them all, I tried to make them not too clunky when actually doing the logging, and we also have some frustrating inconsistencies between amplitude and google. Once we have everything up for filters, and can spend a bit of time getting settled into the new Google ecosystem of analytics tools, we'll have a better idea of what to change (eg. maybe we can drop amplitude all together?). These things should be addressed in a update of all the analytics event setup for WOW - I've started adding some thoughts on that in shortcut [sc-12296]

There is also a small bug that I noticed while testing - when there are 0 records for the table (from filtering) it showed up as `0` so now it returns a react fragment.

### DIFF
--- a/client/src/components/PortfolioFilters.tsx
+++ b/client/src/components/PortfolioFilters.tsx
@@ -3,7 +3,12 @@ import { t, Trans, Plural } from "@lingui/macro";
 import classnames from "classnames";
 import React from "react";
 import { CheckIcon, ChevronIcon, CloseIcon, InfoIcon } from "./Icons";
-import { FilterContext, FilterNumberRange, NUMBER_RANGE_DEFAULT } from "./PropertiesList";
+import {
+  FilterContext,
+  FilterNumberRange,
+  LogPortfolioAnalytics,
+  NUMBER_RANGE_DEFAULT,
+} from "./PropertiesList";
 import FocusTrap from "focus-trap-react";
 import { FocusTarget } from "focus-trap";
 import { Alert } from "./Alert";
@@ -19,8 +24,12 @@ import MinMaxSelect from "./MinMaxSelect";
 
 import "styles/PortfolioFilters.scss";
 
+type PortfolioFilterProps = {
+  logPortfolioAnalytics: LogPortfolioAnalytics;
+};
+
 export const PortfolioFilters = React.memo(
-  React.forwardRef<HTMLDivElement>((props, ref) => {
+  React.forwardRef<HTMLDivElement, PortfolioFilterProps>((props, ref) => {
     const isMobile = Browser.isMobile();
 
     const [showInfoModal, setShowInfoModal] = React.useState(false);
@@ -32,8 +41,13 @@ export const PortfolioFilters = React.memo(
     const { filteredBuildings } = filterContext;
     const { ownernames: ownernamesOptions, zip: zipOptions } = filterContext.filterOptions;
 
+    const { logPortfolioAnalytics } = props;
+
     const [rsunitslatestActive, setRsunitslatestActive] = React.useState(false);
     const updateRsunitslatest = () => {
+      logPortfolioAnalytics(rsunitslatestActive ? "filterCleared" : "filterApplied", {
+        column: "rsunitslatest",
+      });
       setRsunitslatestActive(!rsunitslatestActive);
       setFilterContext({
         ...filterContext,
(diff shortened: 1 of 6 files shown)


client/src/components/PropertiesList.tsx
```diff
@@ -61,6 +62,16 @@ const FilterContextProvider: React.FC<{}> = (props) => {
   return <FilterContext.Provider value={useValue()}>{props.children}</FilterContext.Provider>;
 };
 
+export type LogPortfolioAnalytics = (
```
kiwansim, line 65: thoughts on renaming the type to `PortfolioAnalyticsEvent`? since the `logPortfolioAnalytics` is also the function that actually logs the event

client/src/components/PropertiesList.tsx
```diff
@@ -103,13 +128,18 @@ const PropertiesListWithoutI18n: React.FC<
     <div className="PropertiesList" ref={tableRef}>
       {isTableVisible ? (
         <FilterContextProvider>
-          {useNewPortfolioMethod ? <PortfolioFilters /> : <></>}
+          {useNewPortfolioMethod ? (
+            <PortfolioFilters logPortfolioAnalytics={logPortfolioAnalytics} />
+          ) : (
+            <></>
+          )}
```
kiwansim, line 135: unsure if there was a reason for the empty div but if it's extraneous: 

```suggestion
          {useNewPortfolioMethod && (
            <PortfolioFilters logPortfolioAnalytics={logPortfolioAnalytics} /> )} 
```

client/src/components/PortfolioTable.tsx
```diff
@@ -587,7 +606,7 @@ export const PortfolioTable = React.memo((props: PortfolioTableProps) => {
             ))}
           </thead>
           <tbody>
-            {table.getRowModel().rows.length &&
+            {table.getRowModel().rows.length ? (
```
kiwansim, line 609: does this get fixed if you `parseInt(table.getRowModel().rows.length)`? 
it's weird that the 0 is being logged as a valid length and i wonder if it's a underlying type issue

client/src/components/PropertiesList.tsx
austensen: oh yeah, good point - much more clear

client/src/components/PortfolioTable.tsx
austensen: yeah this was the issue - when filters are applied that exclude all properties the length is `0` and so that's falsey in the `&&` and so the `0` goes through here. I'm not sure I know what you mean about the type issue - like it's weird that the list of rows can be empty?

client/src/components/PortfolioFilters.tsx
```diff
@@ -449,6 +482,7 @@ function FilterAccordion(props: {
           <summary
             onClick={(e) => {
               e.preventDefault();
+              !isOpen && onOpen && onOpen();
```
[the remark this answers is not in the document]
austensen: yeah I guess right now it's provided in every case so could be required - I was just thinking since it's not necessary to the functioning of the component, just an opportunity to do something extra when it opens, it felt like should be optional. Like maybe we decide at some point we don't need to log the open event, or maybe only for new filters as we add them, then we can just not provide the prop rather than remove it from the component.

client/src/components/PortfolioTable.tsx
```diff
@@ -598,7 +617,10 @@ export const PortfolioTable = React.memo((props: PortfolioTableProps) => {
                     {activeFilters.ownernamesActive && OwnernamesResultAlert}
                   </td>
                 </tr>
-              )}
+              )
+            ) : (
+              <></>
+            )}
```
kiwansim, line 623: if you decide to go with the `!!` above

```suggestion
            )}
```

client/src/components/PortfolioTable.tsx
austensen: ah ok, thanks for clarifying. I don't think that there have been issues with the type from `rows`, so this `!!` should deal with it all and make it more clear.

kiwansim (approved):

PR #753, opened 2023-04-27
I18n fixes

austensen:
Previously (#743) I tried to remove `i18n` as a prop from some of the filters components and instead just import it in those files. However, I misunderstood exactly how this is supposed to work and since we didn't have all the translations in I didn't notice that it broke at first. 

This PR fixes those issues using the proper ways to pass around the correct `i18n` prop without having to manually add it each time, following the examples elsewhere in wow. This makes use of `<I18n> {({i18n}) => ()} </I18n>` pattern for the multiselect and minmaxselect components, and the `withI18nProps` & `withI18n()` pattern for the Filters and Table components.

In looking this up a bit more, I think we should update to the latest version of lingui since it seems like there might be [easier ways to manage](https://js-lingui-git-next.lingui-js.vercel.app/ref/react.html) this now, like with `const { i18n } = useLingui()`.   I started to document this in shortcut [sc-12308]


client/src/components/PortfolioFilters.tsx
```diff
@@ -24,11 +24,11 @@ import MinMaxSelect from "./MinMaxSelect";
 
 import "styles/PortfolioFilters.scss";
 
-type PortfolioFilterProps = {
+type PortfolioFilterProps = withI18nProps & {
   logPortfolioAnalytics: LogPortfolioAnalytics;
 };
 
-export const PortfolioFilters = React.memo(
+const PortfolioFiltersWithoutI18n = React.memo(
```
kiwansim, line 31:
```suggestion
const PortfolioFiltersComponent = React.memo(
```

client/src/components/PortfolioFilters.tsx
```diff
@@ -522,4 +525,8 @@ function FilterAccordion(props: {
       )}
     </>
   );
-}
+});
+
+const PortfolioFilters = withI18n()(PortfolioFiltersWithoutI18n);
```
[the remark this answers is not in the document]
kiwansim: nit, but the `withoutI18n` sounds a lil clunky

client/src/components/PortfolioTable.tsx
```diff
@@ -88,8 +89,8 @@ type PortfolioTableProps = {
  * This component memoizes the portfolio table via React.memo
  * in an attempt to improve performance, particularly on IE11.
  */
-export const PortfolioTable = React.memo((props: PortfolioTableProps) => {
-  const { data, locale, rsunitslatestyear, getRowCanExpand, logPortfolioAnalytics } = props;
+const PortfolioTableWithoutI18n = React.memo((props: PortfolioTableProps) => {
```
kiwansim, line 92: same here

kiwansim (approved):
this looks good to me, but two q's:
1. how did you know i18n was broken when it wasn't initially caught when you were doing dev testing for #743 ? 
2. could you point me to docs you used to understand `<I18n>` react component? a bit confused about the difference between i18n from @lingui/core and I18n from @lingui/react but the docs i'm finding are not particularly detailed.

client/src/components/PortfolioFilters.tsx
kiwansim: another option 

```suggestion
const BasePortfolioFilters = React.memo(
```

austensen: Yeah the naming is a little odd, but I was just copying the convention that used elsewhere in WOW - there are a lot of cases like this and they all just keep the final component name with `WithoutI18n` appended.

kiwansim: didn't realize it's an existing convention! thanks for clarifying

austensen:
The table has all the translations, since that mostly hasn't changed, so I saw that it wasn't translating. I think I had just overlooked that at the time since I was focused on the filter components and wasn't really changing the table at all until this i18n change, which was kind of a last minute addition during the PR review. But then I was testing it out by just swapping strings with ```i18n._(t`Landlord`)``` since that is translated, and making sure that came through. 

Yeah, the docs are pretty space, and we are also pretty behind in the lingui versions, so I was mostly taking a closer look at other examples from within WOW. Like in [DetailView](https://github.com/JustFixNYC/who-owns-what/blob/master/client/src/components/DetailView.tsx#L87) where the `<I18n> {({i18n}) => ()} </I18n>` pattern is used for smaller more standalone components (and I think this has been removed from future versions, since I don't see it in the [current docs](https://js-lingui-git-next.lingui-js.vercel.app/ref/react.html) I was looking at), and other more substantial components, like for pages, that use the `withI18n()` pattern. 

I'm still feeling a little unsure about all this, but as I understand it, everything needs to be accessing the `i18n` object (from `/core`) that's initialized and shared throughout using the `I18nProvider` that wraps the whole app. Then everything else is about different ways to kinda magically access that from child components. So both of the above patterns are was to do that. 

Confusingly, the type of `i18n` from `/core` is `I18n` (also from `/core`), the same as the `<I18n>` component from `/react`. And we are commonly importing the type for when `i18n` is explicitly passed around as props, in the situation where we aren't using the above two special patterns.

PR #756, opened 2023-05-01
fix hidden underline on IOS

austensen:
For the design system we use `text-underline-position: under` for everything so that the underline never crosses descenders in the text. This has caused some issues (on IOS only it seems) because the auto height of the element content doesn't account this lower underline, it just uses the line-height as normal, and there are inconsistencies with whether the underline shows above other content. On IOS if `display` is `block` or `inline-block` it doesn't show through, but it does if it's `flex`. And it just seems to work regardless on other browser/devices. 

You can kinda see what's happening here:
image

I documented this issue for the design system in shortcut as well [sc-12324]

[sc-12267]

### DIFF
--- a/client/src/styles/_button.scss
+++ b/client/src/styles/_button.scss
@@ -222,6 +222,7 @@
 .button.is-text {
   padding: 0;
   color: $justfix-black;
+  display: flex; // needed for IOS issue with hidden underline position under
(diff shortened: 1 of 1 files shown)


austensen:
[sc-12310]

PR #759, opened 2023-05-02
remove auto focus of inputs on opening accordions (desktop)

austensen:
We've gone back and forth on this a few times. Most recently, on desktop only, when you opened the filter accordions it would automatically focus the inputs. This had already been disabled for mobile, and now this behavior has been removed from desktop as well. (Though the optional prop is retained in FilterAccordion in case we change our minds again later)
[sc-12270]


kiwansim (approved):
nice, looks so much cleaner!

PR #761, opened 2023-05-02
Fix mobile multiselect searchbar width 

austensen:
On mobile our multiselect searchbar/options list is only 75% width. There was an invisible div that was missed in setting that width, so it extended out of the intended width and clicking in this area was focusing the search bar and opening the options dropdown. 
image
Now everything is the correct width, and also the "clear" selections button is aligned with the search bar. 
image

### DIFF
--- a/client/src/styles/PortfolioFilters.scss
+++ b/client/src/styles/PortfolioFilters.scss
@@ -284,6 +284,12 @@
             @include for-phone-only {
               padding: 0;
 
+              .multiselect__menu,
+              .multiselect__control,
+              .multiselect__selected-value-control-container {
+                width: 75%;
+              }
+
               .multiselect__input-container,
               .multiselect__placeholder {
                 height: 5rem;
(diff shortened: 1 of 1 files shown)


client/src/styles/PortfolioFilters.scss
```diff
@@ -284,6 +284,12 @@
             @include for-phone-only {
               padding: 0;
 
+              .multiselect__menu,
+              .multiselect__control,
+              .multiselect__selected-value-control-container {
+                width: 75%;
+              }
+
```
kiwansim, line 292: there are some classnames that are redeclared after this block. consolidate like this?
`
```suggestion
          .multiselect__input-container,
          .multiselect__placeholder {
                font-size: 1.8rem;
                line-height: 100%;
                height: 5rem;
                padding-left: 1.2rem;
              }

              .multiselect__menu {
                font-size: 1.6rem;
              }
```

austensen: Thanks! just incorporated that

kiwansim (approved):

PR #764, opened 2023-05-03
Clear unapplied filter selections

austensen:
For our multiselect filters all selections made from the options dropdown are saved in the multiselect component state as "pending" selections, then once "apply" is clicked that are confirmed and passed into the main filters context/state. We want to make sure things match user expectations, so this PR changes so that:

any selections that are not applied are cleared when the filter dropdown is closed
any applied selections persist in the filter when opened/closed
when all applied filter selections are cleared with the "clear all" button the "pending" selections should also be cleared from the multiselect component. 

Also, in doing this I realized there was some inconsistency with the types of the multiselect filter selections that needed to be fixed. No more `any` used in those!

For the MinMax select filter we do the same thing. However, to be able to show the applied selections in the component based on the applied filter selections (since they are cleared from the internal minmax select state on every close) we needed to change the data structure of those applied selections to know whether the selection is from the custom range or the presets (or if it just the default -Infinity/Infinity min max values). 

Here's how that all looks for both kinds of filters:

https://user-images.githubusercontent.com/16906516/236201775-f2250916-b1aa-460d-aef8-2d941f23aeae.mov

[sc-12346]
[sc-11789]
[sc-12329]

### DIFF
--- a/client/src/components/MinMaxSelect.tsx
+++ b/client/src/components/MinMaxSelect.tsx
@@ -62,7 +62,6 @@ function MinMaxSelect({
           return { ...preset, checked: selectionMinValues.includes(preset.min) };
         });
         setPresets(presetSelections);
-        console.log({ defaultSelections, selectionMinValues, presetSelections });
       } else {
         setCustomRange(NUMBER_RANGE_DEFAULT);
         setPresets((prev) => prev.map((preset) => ({ ...preset, checked: false })));
@@ -122,7 +121,6 @@ function MinMaxSelect({
         tabIndex={hasCustomInputs ? 0 : -1}
       >
         {presets.map((preset, i) => {
-          console.log({ presets });
           return (
             <div className="minmaxselect__preset-value" key={i}>
               <input
--- a/client/src/components/MinMaxSelect.tsx
+++ b/client/src/components/MinMaxSelect.tsx
@@ -52,19 +52,23 @@ function MinMaxSelect({
 
   React.useEffect(() => {
     if (!isOpen) {
-      if (defaultSelections.type === "custom") {
-        setCustomRange(defaultSelections.values[0]);
-        setPresets((prev) => prev.map((preset) => ({ ...preset, checked: false })));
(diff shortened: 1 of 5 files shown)


kiwansim:
found this bug where the building size selections don't clear fully but otherwise the PR looks good to me

https://user-images.githubusercontent.com/17574626/236054094-c683449a-8d33-4b48-8c81-75d2605fcb08.mp4

austensen:
Thanks, I totally forgot about the minmax component! Just updated to apply the same changes there. Updated the description and video above as well.

client/src/components/MinMaxSelect.tsx
```diff
@@ -84,25 +121,28 @@ function MinMaxSelect(props: {
...
-              checked={preset.checked}
-            />
-            <label htmlFor={`minmaxselect__preset-${i}`}>
-              {preset.min}
-              {isFinite(preset.max) ? `-${preset.max}` : "+"}
-            </label>
-          </div>
-        ))}
+        {presets.map((preset, i) => {
+          console.log({ presets });
```
kiwansim, line 125: remove console.log

client/src/components/MinMaxSelect.tsx
```diff
@@ -35,6 +50,25 @@ function MinMaxSelect(props: {
 
   const hasCustomInputs = isFinite(customRange.min) || isFinite(customRange.max);
 
+  React.useEffect(() => {
+    if (!isOpen) {
```
kiwansim, line 54: for code clarity, i'd change this to a switch statement.
`switch (!isOpen) {
    case (defaultSelections.type === "custom"):
        ...
        break;
    case (defaultSelections.type === "presets"):
        ...
        break;
    default:
        ...
}
`

kiwansim: that's my only note for this PR before approving!

austensen: good idea! much nicer. just updated

kiwansim (approved):
🚀

PR #767, opened 2023-05-05
widen zip filter for desktop, chips flex direction row

austensen:
We changed the language for the multiselect filters from "clear" to "clear selections", and on desktop we ran out of room. So we are widening the filter dropdown. And now that there is more room, we decided to change the selected chips to be `flex-direction: row` so the spacing looks better. 

image image


kiwansim (approved):
🏅

PR #768, opened 2023-05-05
Mobile filter menu tests

kiwansim:


kiwansim:
this was branching from Maxwell's changing the mobile filters to a modal. we decided to scrap the modal approach since it didn't solve all of our issues.

PR #770, opened 2023-05-05
add flag to enable filters

austensen:
Adds a feature flag to enable/disable the filters (when on new ("wowza") portfolio mapping version). I've set up the variables on netlify as well - on demo it's enabled, on prod it's disabled for now. I followed the example from Shannon #707 

[sc-12386]


kiwansim (approved):
:shipit: yay!! one step closer

PR #786, opened 2023-06-06
add map view for portfolio filters

austensen:
This PR continues our work on WOW for Collective Action adding filters to the portfolio tab - first the filter UI and updated table were added in #709, and this adds the map view.

We adapt the existing PropertiesMap component used on the Overview tab to be used on both tabs, adding a new prop to indicate the version to use. There are different point styles, a new selected building info alert, and logic for filtering the list of addresses based on the selected filters (passed through the same ContextProvider as we used for the portfolio Table view). 

Each MapBox map load is fairly expensive, so we try to only load it when necessary (the tab is in view), and also prevent multiple loads. In the WOW AddressPage component, where all of the contents of the tabs (overview, portfolio, etc) is included, we now keep track of when the overview is active to prevent the MapBox load before the tab is in view. We do the same thing within the portfolio tab's PropertiesList component and also use display:none to toggle map/table view to prevent extra map loads.

For the filters context we add a `viewType` (map/table), and move some of the hooks for updating the filter options/selections react context from the table component to the filters component. In the filters bar we also add a map/table view toggle. In the filters component file we add a function for filtering a list addresses based on filter selections, and this is used in the map component. 

We also decided to make a few changes not directly tied to the portfolio map (sorry these are all bundled in this big PR!):
* the order of the tabs on the address page - to put Portfolio second - so there is some rearranging on the AddressPage component
* one of the sample portfolios on the homepage was removed (kushner/westminster) because it has been absorbed into a large network portfolio, so the info card isn't accurate and it doesn't serve as a good example anymore. 
* a few minor style changes to the filters ui to keep things more in line with design system rules
* the alerts for when RS & owner name filters are applied no longer appear within the body of the portfolio table, and just go in the filters bar

image

### DIFF
--- a/client/src/components/PortfolioFilters.tsx
+++ b/client/src/components/PortfolioFilters.tsx
@@ -8,7 +8,7 @@ import {
   PortfolioAnalyticsEvent,
   NUMBER_RANGE_DEFAULT,
   FilterNumberRangeSelections,
-  filterAddrs,
+  filterAddresses,
 } from "./PropertiesList";
 import FocusTrap from "focus-trap-react";
 import { FocusTarget } from "focus-trap";
@@ -64,23 +64,20 @@ const PortfolioFiltersWithoutI18n = React.memo(
       setFilterContext({
         ...filterContext,
         filterOptions: {
-          // put corporations like "123 Fake St, LLC" at the end so real names are shown first
           ownernames: getOwnernamesOptions(assocAddrs),
           unitsres: getUnitsresOptions(assocAddrs),
           zip: getZipOptions(assocAddrs),
         },
       });
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [assocAddrs]);
+    }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
     React.useEffect(() => {
       setFilterContext((prevContext) => ({
         ...prevContext,
         totalBuildings: assocAddrs.length,
-        filteredBuildings: filterAddrs(assocAddrs, filterSelections).length,
+        filteredBuildings: filterAddresses(assocAddrs, filterSelections).length,
       }));
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [assocAddrs, filterSelections]);
+    }, [filterSelections]); // eslint-disable-line react-hooks/exhaustive-deps
 
     const [rsunitslatestActive, setRsunitslatestActive] = React.useState(false);
     const updateRsunitslatest = () => {
@@ -243,7 +240,7 @@ const PortfolioFiltersWithoutI18n = React.memo(
             <FilterAccordion
               title={i18n._(t`Landlord`)}
               subtitle={i18n._(t`Person/Entity`)}
-              infoIconAria={i18n._(
+              infoIconAriaLabel={i18n._(
                 t`Learn more about what it means for someone to be listed as a landlord`
               )}
               infoModalContents={ownernamesInfoModalContents}
@@ -486,7 +483,7 @@ const OwnernamesInfoAlert = (
 type FilterAccordionProps = withI18nProps & {
   title: string;
   subtitle?: string;
-  infoIconAria?: string;
+  infoIconAriaLabel?: string;
   onInfoClick?: () => void;
   infoModalContents?: JSX.Element;
   children: React.ReactNode;
@@ -510,7 +507,7 @@ const FilterAccordion = withI18n()((props: FilterAccordionProps) => {
   const {
     title,
     subtitle,
-    infoIconAria,
+    infoIconAriaLabel,
     onInfoClick,
     infoModalContents,
     children,
@@ -563,7 +560,7 @@ const FilterAccordion = withI18n()((props: FilterAccordionProps) => {
                 {infoModalContents && (
                   <button
                     className="filter-info"
-                    aria-label={infoIconAria}
+                    aria-label={infoIconAriaLabel}
                     onClick={() => {
                       setShowInfoModal(true);
                       onInfoClick && onInfoClick();
--- a/client/src/components/PortfolioFilters.tsx
+++ b/client/src/components/PortfolioFilters.tsx
@@ -207,27 +207,27 @@ const PortfolioFiltersWithoutI18n = React.memo(
             </span>
           )}
         </div>
+        <div className="view-type-toggle-container">
+          <button
+            aria-pressed={viewType === "table"}
+            className="view-type-toggle"
+            onClick={() => setFilterContext((prev) => ({ ...prev, viewType: "table" }))}
+          >
+            <Trans>Table</Trans>
+          </button>
+          <button
+            aria-pressed={viewType === "map"}
+            className="view-type-toggle"
+            onClick={() => setFilterContext((prev) => ({ ...prev, viewType: "map" }))}
+          >
+            <Trans>Map</Trans>
+          </button>
+        </div>
         <FiltersWrapper
           isMobile={isMobile}
           activeFilters={activeFilters}
           resultsCount={filteredBuildings}
         >
-          <div className="view-type-toggle-container">
-            <button
-              aria-pressed={viewType === "table"}
-              className="view-type-toggle"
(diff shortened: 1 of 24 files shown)


client/src/components/PortfolioTable.tsx
```diff
@@ -558,168 +557,183 @@ const PortfolioTableWithoutI18n = React.memo((props: PortfolioTableProps) => {
...
                         >
-                          {flexRender(cell.column.columnDef.cell, cell.getContext())}
-                        </td>
+                          {header.isPlaceholder ? null : (
+                            <>
+                              <div
+                                {...{
+                                  className: header.column.getCanSort()
+                                    ? "cursor-pointer select-none"
+                                    : "",
```
kiwansim, line 602:
```suggestion
                                  className: ! header.column.getCanSort()
                                    ?? "cursor-pointer select-none"      
```

client/src/components/PropertiesList.tsx
```diff
@@ -62,7 +69,48 @@ const useValue = () => {
...
+export const filterAddrs = (addrs: AddressRecord[], filterSelections: FilterSelections) => {
+  const {
+    rsunitslatest: filterRsunitslatest,
+    ownernames: filterOwnernames,
+    unitsres: filterUnitsres,
+    zip: filterZip,
+  } = filterSelections;
+
+  return addrs.filter((addr) => {
+    let keepAddr = true;
```
kiwansim, line 81: weirdly, i don't mind local variables with shorthand names

client/src/components/PropertiesList.tsx
```diff
@@ -62,7 +69,48 @@ const useValue = () => {
...
+        Object.entries(_groupBy(addr.allcontacts, "value")).map((contact) => contact[0]);
+
+      keepAddr = filterOwnernames.some((filterOwnername) =>
+        addrOwnernames?.includes(filterOwnername)
+      );
+    }
+
+    if (keepAddr && filterUnitsres.type !== "default") {
+      keepAddr = filterUnitsres.values.reduce(
+        (acc, rng) =>
```
kiwansim, line 99: rng?

client/src/components/PropertiesMap.tsx
```diff
@@ -112,13 +193,88 @@ export default class PropertiesMap extends Component<Props, State> {
...
+          },
+        });
+      }
+    );
+  }
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    this.state.mapRef?.resize();
+
+    // this.isOnOverview() && this.zoomToNewDetailAddr(prevProps);
```
kiwansim, line 221: remove comment

client/src/components/PropertiesMap.tsx
```diff
@@ -112,13 +193,88 @@ export default class PropertiesMap extends Component<Props, State> {
...
+    const { newAssocAddrs, newAddrsBounds } = this.getAddrsPointsAndBounds();
+
+    // sets things up, including initial portfolio level map view
+    this.setState(
+      {
+        addrsBounds: newAddrsBounds,
+        addrsPoints: newAssocAddrs,
+      },
+      () => {
+        // yeah, this sucks, but it seems to be more consistent with
```
kiwansim, line 205: lol the dev internal monologue 😆
i do think the reasoning behind this approach is valuable if you want to keep the comment (or some version of it) in

client/src/components/PropertiesMap.tsx
```diff
@@ -112,13 +193,88 @@ export default class PropertiesMap extends Component<Props, State> {
...
+    } else {
+      map.getCanvas().style.cursor = "";
+    }
+  };
+
+  handleAddrSelect = (addr: AddressRecord, e: any) => {
+    // On Overview page, this updates global context/state with new detail address. On Portfolio this just has telemetry.
+    this.props.onAddrChange(addr.bbl);
+
+    // TODO: when on portfolio page, update state for selected address
```
kiwansim, line 250: remove comment

client/src/containers/HomePage.tsx
```diff
@@ -261,20 +205,6 @@ const HomePage: React.FC<HomePageProps> = ({ useNewPortfolioMethod }) => {
                         A&amp;E Real Estate
                       </Link>
                     </h6>
-                    <Link
```
kiwansim, line 264: is there a reason why we're removing the logos? it was a nice visual break on our landing page imo

kiwansim (changes requested):
first pass -- continuing to look at Properties Map with a fine tooth comb and haven't looked at style files.

client/src/components/PortfolioFilters.tsx
```diff
@@ -38,16 +44,44 @@ const PortfolioFiltersWithoutI18n = React.memo(
...
+      setFilterContext({
+        ...filterContext,
+        filterOptions: {
+          // put corporations like "123 Fake St, LLC" at the end so real names are shown first
+          ownernames: getOwnernamesOptions(assocAddrs),
+          unitsres: getUnitsresOptions(assocAddrs),
+          zip: getZipOptions(assocAddrs),
+        },
+      });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
austensen: yeah you're right, `assocAddrs` couldn't update on this page, thanks!

client/src/components/PortfolioTable.tsx
```diff
@@ -558,168 +557,183 @@ const PortfolioTableWithoutI18n = React.memo((props: PortfolioTableProps) => {
...
-                          }}
+                        <th
+                          key={header.id}
+                          colSpan={header.colSpan}
+                          className={`col-${header.column.id}`}
+                          style={{ minWidth: header.getSize() || undefined }}
                         >
-                          {flexRender(cell.column.columnDef.cell, cell.getContext())}
-                        </td>
+                          {header.isPlaceholder ? null : (
```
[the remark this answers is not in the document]
austensen: this is breaking the table for me, it doesn't seem to handle `false` the same as `null`

client/src/components/PortfolioTable.tsx
austensen: this whole little chunk was written in a weird spread notation thing. I cleaned that up, and then used `classnames()` with your suggestion (otherwise the `false` wasn't accepted for that prop)

client/src/components/PropertiesList.tsx
austensen: makes sense, I saw a lot of use of `addrs` as prop names and such, but I agree for function/component names the full word makes sense.

client/src/components/PropertiesList.tsx
austensen: range.. I was just copying some patterns in react-table functions, but yeah this isn't clear - fixing!

client/src/components/PropertiesMap.tsx
austensen: haha yeah this was i think from sam, or maybe atul. i also found it useful context - i'll leave it

client/src/containers/HomePage.tsx
austensen: after cutting one, they looked a bit unbalanced and leslie decided to shorten the description and drop the images. I think after portfolio pruning we should revisit these and add back logos and a third example. https://app.shortcut.com/justfixnyc/story/12695/nice-to-have-make-example-portfolios-more-visually-consistent-by-removing-logos-and-shortening-stellar-description

client/src/components/PropertiesList.tsx
```diff
@@ -77,17 +126,40 @@ export type PortfolioAnalyticsEvent = (
...
+    map: viewType === "map",
+  });
+
+  React.useEffect(() => {
+    setViewsEverVisible((prev) => ({
+      table: prev.table || viewType === "table",
+      map: prev.map || viewType === "map",
+    }));
+  }, [viewType]);
+
```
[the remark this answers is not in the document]
austensen: added a comment in the code here - we don't expect it to be false again, but in future we might be able to land on the portfolio page in map or table view, so this will be helpful then.

kiwansim (approved):
🗺️ 👁️ 👍🏽

PR #793, opened 2023-07-05
change detail page streetview to static image

austensen:
The API for the dynamic google sreetview embed we use on the address page detail view is super expensive, and also the quality of the images has been declining (it's now common to be weird interiors of storefronts etc.). So we've decided to swap it with the way cheaper static streetview image we are already using on summary tab and have a link to google maps. 

[sc-11959]


kiwansim (approved):
lgtm 🚗

PR #795, opened 2023-07-07
User object not clearing on logout, Redirects from restricted pages

kiwansim:
* on logout, the user object was still being populated with undefined values in `email`, `verified` and empty `subscriptions`. turns out it's because of how the `postAuthRequest` was still returning a `response.json` even if the response was not 2xx. 
* redirects to home any time the user logs out from the nav bar while on pages that should require a login 

[sc-12766] [sc-12808]


austensen (approved):
This all looks great! Tested it out locally, and it's all working as expected

PR #787, opened 2023-06-27
Map filter QA revisions

austensen:
Various changes from our second round of QA on portfolio filters map view.

(will change base branch to `portfolio-filters` once `map-filters` is reviewed and merged in)

- [fix wrapping of filter controls for narrow screens](https://github.com/JustFixNYC/who-owns-what/commit/c0d5c420ab4ca439d8e63e8ed63bf85e7df548ce) [sc-12745]
- [increase text contrast for disabled pagination buttons](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/b60a02fe423b1e248a4bfaa093a5d68aec4da515) [sc-12749]
- [resize map when container height changes](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/880633602ba20ebe163806cf372896df77e06842) [sc-12744], [sc-12595]
  - this required adding a new package, [react-resize-detector](https://www.npmjs.com/package/react-resize-detector), since there was no good way to detect changes in the element size within this class component. But the package is very small (~3kb), and seems like the most modern and simple solution, and could be helpful elsewhere as too.
- [bldg profile links on selected bldg alert open in new tab](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/a9b26a19d8e72fb0f73f832a41ad4de06c2b49b8) [sc-12746]
- [use unselected point style for filter legend](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/03e75d95f0b2485eceff490f5852bad60b78ce3a) [sc-12751]
- [fix disappearing chevron in building size accordion](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/34f460a39b9741fa585c24d1ced69722f0ce49e3) [sc-12756]
- [fix info icon cut off slightly on safari](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/83923f19766f266adb3e5629abd15f5aec8ab1e9) [sc-12743]
- [use standard wow link styles in modal (blue)](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/a7a6c59deeed75a7f94cd1a06a5af5eb72ff8bd3) [sc-12752]
- [match bg color of dropdown to bg color of filter bar & filter buttons](https://github.com/JustFixNYC/who-owns-what/pull/787/commits/1b3d594c44fb2656fa8a67c41a69da9e058ed0c5) [sc-12754]
[...]

PR #801, opened 2023-07-20
add resize-observer-polyfill package

austensen:
In #787, when we had problems with the portfolio tab map not adjusting when the container height changed, we introduced a new package to use ResizeObserver. It is apparently supported by ~95% of browsers, but in the few cases it's not supported we are getting errors that it's not defined.

A [github issues thread](https://github.com/recharts/recharts/issues/2491) on the package repo explains that there is an alternative (the library [`resize-observer-polyfill`](https://www.npmjs.com/package/resize-observer-polyfill)) that can be assigned to `window.ResizeObserver` to handle these cases.

I tested this out by just first setting `window.ResizeObserver = undefined`, confirming it breaks, then using the polyfill and confirming it works again, and haven't had any problems. 

Only thing I wasn't totally sure about was if this `index.js` is the best place to put this - does that seem ok?

[sc-12880]

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -72,6 +72,7 @@
     "react-select": "^5.7.2",
     "react-social": "^1.10.0",
     "react-transition-group": "^2.2.1",
+    "resize-observer-polyfill": "^1.5.1",
     "rollbar": "^2.14.4",
     "sass": "^1.49.9",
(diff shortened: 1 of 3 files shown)


kiwansim (approved):
index.js looks like a good place to put this since it's globally affected

PR #803, opened 2023-07-24
Validate email address for routing log in / register flow

kiwansim:
* adds a GET endpoint to check if email is already used. does not require user authentication since this check happens pre-login
* Toggles between log in / register -- flows match regardless of entry point to page (e.g. nav bar vs. building page) 
* some UI parity fixes (e.g. show/hide svg icons)
* general refactoring 

forthcoming:
* error messaging UI parity 
* styling login module on building page (looks jank rn)


austensen (approved):
This all looks good to me! Tested it out locally and everything worked perfectly

PR #805, opened 2023-07-27
fix param name for address change gtm tag

austensen:
[sc-12888]


kiwansim (approved):
when the param was `column`, were these events just not being logged?

austensen:
the event was logged, but just without the extra params. So we know that all of those address changes were happening from the table (since the map param was working) but not which columns. Though if we needed to we could get this from the more general "link click" event, it just would take some extra filtering

PR #806, opened 2023-07-28
update portfolio table pagination bar styles

austensen:
The contrast on the prev/next buttons in the portfolio table pagination bar were not meeting a11y standards, so while fixing that we made some other improvements to the design. The prev/next buttons now use icons, have clear hover and disabled states, and have 48px touch target for a11y.  We also removed the ability to select a specific page (the `page 1 of 2` is just text now).
[sc-12749]

While making these changes I needed to fix how the chevron icon was working. I had previously set a default size in the svg component but that made it difficult to override from the css, now those are there are no defaults and you set it in the css like everything else. The viewbox was also tweaked so it's a square and the icon is centered within that - this makes placement easier and ensures it stays centered when rotated. We should make similar fixes to the other icons eventually too. 
[sc-12916]

image

sidenote: for some reason we were using `position: sticky` for the pagination bar (not necessary anymore) but there is some weird kind of bug in some ios versions when something is position:sticky where the dom updates (the page number in this case) but the browser doesn't do the redraw/paint step and so the numbers wouldn't change or would get all glitchy looking. just removing sticky solved it.

### DIFF
--- a/client/src/components/Icons.tsx
+++ b/client/src/components/Icons.tsx
@@ -32,14 +32,7 @@ export const CheckIcon = (props: SVGProps<SVGSVGElement>) => (
 );
 
 export const ChevronIcon = (props: SVGProps<SVGSVGElement>) => (
-  <svg
-    width="11"
-    height="7"
-    viewBox="0 0 11 7"
-    fill="none"
-    xmlns="http://www.w3.org/2000/svg"
-    {...props}
-  >
+  <svg fill="none" viewBox="0 -2 10 10" xmlns="http://www.w3.org/2000/svg" {...props}>
     <path d="M9.5 1L5.25 5.25L1 1" stroke="currentcolor" />
   </svg>
 );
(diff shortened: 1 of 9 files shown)


client/src/components/Icons.tsx
```diff
@@ -32,14 +32,7 @@ export const CheckIcon = (props: SVGProps<SVGSVGElement>) => (
...
 export const ChevronIcon = (props: SVGProps<SVGSVGElement>) => (
-  <svg
-    width="11"
-    height="7"
-    viewBox="0 0 11 7"
-    fill="none"
-    xmlns="http://www.w3.org/2000/svg"
-    {...props}
-  >
+  <svg fill="none" viewBox="0 -2 10 10" xmlns="http://www.w3.org/2000/svg" {...props}>
```
kiwansim, line 35: did you change the viewBox value manually? if so, how did you know what values to set it to

kiwansim (approved):
looks good!!

PR #808, opened 2023-08-01
avoid extra streetview api call using "loading" placeholder

austensen:
In #793 we switched from dynamic to static streetview images for the address page. Because the images get easily distorted switching between mobile and desktop, we fetch a new version of the image based on window size, but on initial load there is a delay in getting the screen size and so it was defaulting to mobile size then quickly switching to desktop size, causing an extra API call. 

This PR fixes the issue by just using a "loading" placeholder until the window size is known so we only make one request.


kiwansim (approved):
i can't reproduce getting the "loading" text to show on streetview, but the code change looks good to me

austensen:
yeah, I think it just happens so quickly you don't really see it, but it does prevent the extra call

PR #812, opened 2023-09-09
update protected oca data access via s3

austensen:
For our BBL-level counts of eviction filings we need to use the address-level OCA data ("level 2") that needs to be kept in a secure database accessible only by members of the OCA Data Collective. We recently updated our system for managing the OCA data: switching over to a new database, and consolidating the ETL process for the public "level 1" and address-level "level 2". https://github.com/housing-data-coalition/oca/tree/level2

The new protected database now uses an IP allow list rather than the SSH tunnel, and since our jobs run on kubernetes it's tricky to get a stable IP. So instead we now export a csv of the address table with BBLs added and it's stored in BetaNYC's S3 bucket and made accessible to use using AWS access key (we're using our "wow" AWS IAM user, see `tenants2-security`). 

In this PR we remove everything for connecting to the OCA database, add the AWS keys for downloading the table `oca_addresses_with_bbl` from S3, and reorganize the SQL scripts to build the aggregated wow-related OCA tables. This also now requires the full public version of OCA (nycdb dataset) as a dependency for WOW, so we add that test data too.


kiwansim (approved):
looks good to me! just a reminder to update .env [here](https://github.com/JustFixNYC/tenants2-security/blob/main/who-owns-what/.env)  as well

PR #827, opened 2023-10-23
Add API endpoint for email alerts data

austensen:
This PR adds new API endpoint(s) to get the data we need to generate custom wow email alerts. These will be used exclusively by our server that generates and sends the emails.

API structure TBD: it will definitely take a bbl and a date range and return indicator data, but we might have separate endpoints for each indicator, a single one where you specify your indicator, or one were you can specify a set of indicators to get all the data in one. 

Now that we have all the bbl-level OCA data in our database (for 11+ unit bbls), this new endpoint can work the same as our existing ones, just calling our wow/nycdb database. However, because we don't want the full bbl-level oca data to be publicly accessible for any date range, we need to secure these endpoints with some authorization that we haven't needed before.

Initially I [tried](https://github.com/JustFixNYC/who-owns-what/pull/827/commits/a449be8f1598132f5f3fa08fbfec3a06374f7e63) the more typical user auth token process as described in this [guide](https://tafadzwalnyamukapa.medium.com/creating-a-secure-well-tested-django-restful-api-54560c65872c) (section 2a). It worked just fine locally, but once deployed to heroku I remembered that since we haven't had a user system for the wow django app we don't have a persistent db set up to check the token against.

Then I found this helpful [SO response](https://stackoverflow.com/a/65037074/7051239), and went with this much more simple method of just storing a secret token in the wow app that we can also use in the email alerts server, and just use that in the authorization header. 

Before deploying this PR, we'll need to add a env variables `ALERTS_API_TOKEN` to heroku (dev & prod)

[sc-13162]
[sc-12832]
[sc-13307]


kiwansim (approved):
looks good to me! 🥇

PR #830, opened 2023-11-21
Add url params for timeline indicators

austensen:
This PR adds a URL parameter on to the `/timeline` route so that we can link to specific indicators, eg. `.../timeline/hpdcomplaints`. 
All of the others use the names we already have set up in the codebase, which are: `hpdcomplaints`, `hpdviolations`,  `dobpermits`, `dobviolations`, `evictionfilings`.
Going to just `/timeline` will bring up the default one (currently HPD Complaints). 
When using the UI to switch between indicators, the URL will update without reloading the page, using the `history` from the router with the `shallow` option.

### DIFF
--- a/client/src/components/Indicators.tsx
+++ b/client/src/components/Indicators.tsx
@@ -25,6 +25,8 @@ import {
 import { NetworkErrorMessage } from "./NetworkErrorMessage";
 import { Dropdown } from "./Dropdown";
 import { AmplitudeEvent, logAmplitudeEvent } from "./Amplitude";
+import { withRouter } from "react-router-dom";
+import { RouteComponentProps } from "react-router";
 
 type TimeSpanTranslationsMap = {
   [K in IndicatorsTimeSpan]: (i18n: I18n) => string;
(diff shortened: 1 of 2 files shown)


austensen:
One thing I wasn't sure about that I'd love to get your take on: when you navigate to the timeline without linking to a specific indicator, should we then update the url to the default indicator, or (as it is currently) should it just say as `/timeline` and only update once a different indicator is selected via the ui?

client/src/components/Indicators.tsx
```diff
@@ -49,10 +52,26 @@ const getDropdownWidthFromLongestSelection = (selections: string[]) => {
...
+  if (indicatorsDatasetIds.includes(indicator)) {
+    return indicator;
+  }
+};
+
+class IndicatorsWithoutI18n extends Component<IndicatorsWithRouterProps, IndicatorsState> {
+  constructor(props: IndicatorsWithRouterProps) {
     super(props);
-    this.state = indicatorsInitialState;
+    console.log(props.match.params);
```
kiwansim, line 67: remove console

client/src/components/DetailView.tsx
```diff
@@ -285,7 +285,7 @@ class DetailViewWithoutI18n extends Component<Props, State> {
                       <BuildingStatsTable addr={detailAddr} />
                       <div className="card-body-timeline-link">
                         <Link
-                          to={this.props.addressPageRoutes.timeline}
+                          to={removeIndicatorSuffix(this.props.addressPageRoutes.timeline)}
```
kiwansim, line 288: was there a reason for implementing `removeIndicatorSuffix` as opposed to just defaulting the timeline route to `timeline: `${prefix}/timeline/dobviolations`` ? either method works, i am just curious!

kiwansim (approved):

PR #832, opened 2023-11-22
Use new combined hpd complaints and problems table

austensen:
HPD has created a new dataset on open data that combines the previously separate complaints and problems files. This is essentially just the same data joined on complaint id, like we were doing ourselves. 

There are some outstanding questions about a large number of new complaint records from recent years that were not present in the old files. We're continuing to talk with folks at HPD about this and they are investigating this. It seems like the overwhelming majority of the "new" records are related to heat/hot water, and it's possible that this relates to a known issue we've seen with HPD aggregating/de-duplicating multiple complaints about the same issue in a building at the same time (eg. everyone in the building calling about a building wide heat outage) that can be seen in the separate 311 dataset, but we've yet to confirm this. Regardless of what we learn here, they are/have discontinued updates to the old files and there isn't any reason to believe there are errors with the new data, so we should move ahead with updating the source for WOW.

[sc-13240]


austensen:
The python tests keep hanging on [wow/tests/test_dbutil.py](https://github.com/JustFixNYC/who-owns-what/blob/update-hpd-complaints/wow/tests/test_dbutil.py), but nothing was changed that relates to this as far as I can tell. 
I thought there might have been some other changes to nycdb since the last time we updated the nycdb version in this repo, so I tried [just updating the nycdb version](https://github.com/JustFixNYC/who-owns-what/pull/833) to the commit right before the new hpd complaints table was added to nycdb, and all the tests on that passed. 
Then I downgraded the nycdb version to just the one commit ahead of that, which contains [just the addition of this one table](https://github.com/nycdb/nycdb/commit/edda39af109b12fd708169b32abe80cfdfe01b17), along with the changes to wow in this PR, and the tests hang at that step still.

austensen:
ugh I forgot to add the nycdb test data for this new table. The "factory" data is only used for the `/tests` but `wow/tests` build the test database using just the regular csv test data files that need to be copied over from the nycdb repo. So since it wasn't finding it in the folder, it was trying to download the full real file each time and timing out.

austensen:
Yes the csv sample data is just a slice of the file, we do this for every table in nycdb for testing, so I just copied it over from that repo. Here in wow they are used when running the tests in `/wow/sql`, specifically when setting up for those test in [`conftest.py`](https://github.com/JustFixNYC/who-owns-what/blob/master/wow/tests/conftest.py#L48C27-L48C27) where `dbtool.loadtestdata` is run which calls [`NycdbBuilder(..., testing=True)`](https://github.com/JustFixNYC/who-owns-what/blob/master/dbtool.py#L255C85-L256C1) and that runs `nycdb` setting it's normal data folder to be this folde rof test data csvs, so when nycdb runs it's `load` commands it finds these files as though they were created via the `downlaod`.

PR #836, opened 2023-12-01
Portfolio pruning

austensen:
Updates our ownership portfolio generation algorithm. 

A couple years ago we introduced graph network data model but over time we've encountered a problem of the portfolios getting extremely large because of one or two (usually inaccurate) connections linking otherwise reasonable portfolios leading to massive networks (many 100s of properties) that are not accurate or useful to users. 

The first major improvement we make with this update is to add much more extensive address standardization. For this we use the city's official geocoding service, specially GeoSupport - the fast local version of what's on the geosearch API we use elsewhere. This does a great job of standardizing all street spellings and other formatting. It doesn't handle unit/apartment numbers and so we implement a simple manually process for cleaning those up as best we can. We also then keep the addresses in components, and match on addresses using a more nuanced method than simple complete exact match (need exact on zip, fuzzy on number and street, use only numbers for unit).

This new method takes the graph network data further to more closely represent the relationships of the data in nodes/edges, and this allows us to use network analysis algorithms to adjust the portfolios after the initial "connected components" are identified. Specifically, we are using Louvian Communities to find opportunities to split large portfolios into smaller and more densely connected components. 

In the process of implementing this we also reworked the user-facing visualization, and examined a lot of example portfolios to better understand how some of these connections were being made that lead to giant portfolios. Overwhelmingly these were because of matches based on contact names, and usually they were common names with completely different business addresses that likely were not truly related. This lead us to great deemphasis the role of name-based matches to focus more on business address matches, and use a more nuanced method for our conditions of when to make a connection and introduced a new practice of scoring/weighting our confidence in each match. 

We only make a connection if there is either an exact name mach or an exact business address match. And we create a confidence score/weight based on the presence of an exact name match (no partial matching for names) and the similarity of the business addresses. We then can use the weights on the edges (confidence for the connection) in the Louvian communities algorithm for splitting large initial portfolios. 

As it turns out, we are currently not splitting any of the portfolios because the new connection criteria is creating strongly connected smaller portfolios than the old method. At the time of this PR there are only 5 300+ property portfolios and they can't be split using this method (no split would improve the connection of the resulting portfolios), and these have all been manually reviewed and they are all based around single business addresses. (May not be accurate _ownership_ portfolios, probably management, but with the available data quality this is what we want to achieve). However, we are still including this attempt of splitting large portfolios, so as new monthly registration data comes in and things shift we may find the splits kick in and help prevent the creep giant portfolios growing like we experienced last time. 

While we initially reworked the visualization for the new method, after we broke down the clear distinction between name and address based matches, and have better partial match methods the portfolios are so densely connected that the graphs are completely tangled and not useful for users. After some discussion we've decided to remove the graph from the summary page, and will consider revisiting this in the future, especially if users reach out telling us they miss having it. And then we'll explore some new approaches to simplify the information presented. 

We also add (for a limited time) our email to the contact us box that appears on the summary page so we can hopefully get feedback on the changes and the removal of the graph visualization. 

I also had to upgrade our python version from 3.9 to 3.10 because of some issues with the networkx package. 

And a few small frontend changes: 
* add (for a limited time) our email to the contact us box that appears on the summary page so we can hopefully get feedback on the changes and the removal of the graph visualization. [sc-15518]
* add the main contact's business address to the portfolio table's expandable landlords row [sc-15523]
* And we add a note about removing the "legacy" pages a month or two after release of this new method [sc-15496]

Related PR for the k8s-loader: https://github.com/JustFixNYC/nycdb-k8s-loader/pull/164


kiwansim (approved):
📈📈📈

PR #843, opened 2024-01-30
add toast alert for removing bldg subcriptions

austensen:
Add an alert that pops up to confirm that a building is removed from subscriptions. 

https://www.loom.com/share/bc7dcdd5b5b84a57b29b59eb79ed993c?sid=7109102d-2207-4655-9558-b30ab4db2981

image

[sc-12160]


austensen:
Update: for now we've decided this isn't necessary - the building being removed from the list is good enough for now, and we can revisit post-MVP launch if needed.

PR #849, opened 2024-02-20
Multi-step log in/sign up modal, with user type

austensen:
- adds ability for account sign up to happen all within a modal from the building page
- adds a required step for account sign up, asking user type (multiple-choice or write in)
- adds verify email/ re-send email step to the end of log in/sign up flow
- refactors email input into separate component (like `PasswordInput`)
  - email, password, and user type input components now work with custom `useInput` hook so that the parent can keep track of values in state. 
  - This makes it easier for the parent component to hold onto input values from each step to use for error alerts and registration at the end. The same setup also helps with the input-level errors for email and whether they shown. 
  - This still feels a bit overcomplicated, but we'll figure this all our for the component library and hopefully swap those in soon. 

https://github.com/JustFixNYC/auth-provider/pull/14 adds this new field to the user model, and https://github.com/JustFixNYC/auth-provider/pull/18 adds it to the `user/register` auth endpoint

[sc-13535]
[sc-13510]
[sc-13594]
[sc-13532]

### DIFF
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -47,8 +47,8 @@ const clearUser = () => (_user = undefined);
  * Authenticates a user with the given email and password.
  * Creates an account for this user if one does not already exist.
  */
-const register = async (username: string, password: string) => {
-  const json = await postAuthRequest(`${BASE_URL}auth/register`, { username, password });
+const register = async (username: string, password: string, user_type: string) => {
+  const json = await postAuthRequest(`${BASE_URL}auth/register`, { username, password, user_type });
   fetchUser();
   return json;
 };
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -47,8 +47,12 @@ const clearUser = () => (_user = undefined);
  * Authenticates a user with the given email and password.
  * Creates an account for this user if one does not already exist.
  */
-const register = async (username: string, password: string, user_type: string) => {
-  const json = await postAuthRequest(`${BASE_URL}auth/register`, { username, password, user_type });
+const register = async (username: string, password: string, userType: string) => {
+  const json = await postAuthRequest(`${BASE_URL}auth/register`, {
+    username,
+    password,
+    user_type: userType,
+  });
   fetchUser();
   return json;
 };
--- a/client/src/components/EmailAlertSignup.tsx
+++ b/client/src/components/EmailAlertSignup.tsx
@@ -10,6 +10,7 @@ import "styles/EmailAlertSignup.css";
 import { JustfixUser } from "state-machine";
 import AuthClient from "./AuthClient";
 import { AlertIconOutline, SubscribedIcon } from "./Icons";
+import { Alert } from "./Alert";
 
 type BuildingSubscribeProps = withI18nProps & {
   bbl: string;
@@ -42,10 +43,9 @@ const BuildingSubscribeWithoutI18n = (props: BuildingSubscribeProps) => {
   const showEmailVerification = (i18n: any) => {
     return (
       <div className="building-subscribe-status">
-        <div className="status-title">
-          <AlertIconOutline />
-          <Trans>Email verification required</Trans>
-        </div>
+        <Alert type="info">
+          <Trans>Verify your email to start receiving updates.</Trans>
+        </Alert>
         <div className="status-description">
           {i18n._(t`Click the link we sent to ${email}. It may take a few minutes to arrive.`)}
         </div>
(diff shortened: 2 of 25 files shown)


client/src/components/AuthClient.ts
```diff
@@ -47,8 +47,8 @@ const clearUser = () => (_user = undefined);
  * Authenticates a user with the given email and password.
  * Creates an account for this user if one does not already exist.
  */
-const register = async (username: string, password: string) => {
-  const json = await postAuthRequest(`${BASE_URL}auth/register`, { username, password });
+const register = async (username: string, password: string, user_type: string) => {
```
kiwansim, line 50: nit, but I think our convention for JS variables is camel case, so `userType`

client/src/components/EmailAlertSignup.tsx
```diff
@@ -49,29 +49,27 @@ const BuildingSubscribeWithoutI18n = (props: BuildingSubscribeProps) => {
...
-            onClick={() => AuthClient.resendVerifyEmail()}
-          >
-            <Trans>Resend email</Trans>
-          </button>
-        </div>
+        <Alert type="info">
+          <Trans>Verify your email to start receiving updates.</Trans>
+        </Alert>
+        <Trans render="div" className="status-description">
+          Click the link we sent to {email}. It may take a few minutes to arrive.
```
kiwansim, line 56: `{email}` is showing up as plaintext here!
Screenshot 2024-03-05 at 2 27 49 PM

client/src/components/Login.tsx
```diff
@@ -1,120 +1,99 @@
...
+import { InfoIcon } from "./Icons";
 import Modal from "./Modal";
 
-export enum LoginState {
-  Default,
+import "styles/Login.css";
+import "styles/UserTypeInput.css";
+import "styles/_input.scss";
+
+enum Step {
```
kiwansim, line 22: this is a much clearer way of defining the login states, thank u!!

client/src/components/Login.tsx
```diff
@@ -1,120 +1,99 @@
...
-  const [username, setUsername] = useState("");
-  const [password, setPassword] = useState("");
-  const [header, setHeader] = useState("Log in / sign up");
-  const [subheader, setSubheader] = useState("");
-  const [loginState, setLoginState] = useState(LoginState.Default);
-  const isDefaultState = loginState === LoginState.Default;
-  const isRegisterState = loginState === LoginState.Register;
+  const userContext = useContext(UserContext);
 
+  const [showRegisterModal, setShowRegisterModal] = useState(false);
```
kiwansim, line 52: should this be set to the `registerInModal` prop value?

client/src/components/Login.tsx
```diff
@@ -231,79 +192,296 @@ const LoginWithoutI18n = (props: LoginProps) => {
...
+                userType={userType}
+                setUserType={setUserType}
+                error={userTypeError}
+                showError={userShowUserTypeError}
+                setError={setUserTypeError}
+                onChange={onChangeUserType}
+              />
+            )}
+            <div className="submit-button-group">
+              {/* {isRegisterUserTypeStep && (
```
kiwansim, line 439: remove commented out bits. surprised the linter missed this!

jfauthprovider/views.py
```diff
@@ -77,6 +77,7 @@ def register(request):
         "grant_type": "password",
         "username": request.POST.get("username"),
         "password": request.POST.get("password"),
+        "user_type": request.POST.get("user_type"),
```
kiwansim, line 80: referring back to the first comment, the python side of this can be 
`"user_type": request.POST.get("userType")`

client/src/util/helpers.ts
```diff
@@ -132,6 +132,27 @@ export function searchAddrsAreEqual(
   );
 }
 
+// https://www.codevertiser.com/reusable-input-component-react/
```
kiwansim, line 135: this is a really neat new pattern

kiwansim (approved):
wow massive props to you! I'm going to approve so I don't block the merge and my comments are small nits!

client/src/components/Login.tsx
austensen: no, sorry it's a bit confusing. `registerInModal` is for whether it can ever be a modal (ie on the building page, rather than nav login). and `showRegisterModal` controls if it's actually open, and we only want that if they submit their email on the building page and it's not associated with a user so that it opens the modal for registration rather than staying on page for login.

jfauthprovider/views.py
austensen: Thanks for catching this. After looking back in AuthClient I realized all the other methods made the change there so I followed that for consistency

client/src/components/EmailAlertSignup.tsx
austensen: It's working on local so I'm pretty sure this is just that issue with not having done `lingui extract` yet, but I've been holding off until everything is merged into `email-alerts..`

PR #850, opened 2024-02-20
remove legacy/new banner

austensen:
Remove the legacy/new version banner from the top of wow. 

image

[sc-13348]

### DIFF
--- a/client/src/containers/App.tsx
+++ b/client/src/containers/App.tsx
@@ -337,44 +337,8 @@ const AppBody = () => {
   );
 };
 
-const WowzaBanner = withI18n()((props: withI18nProps) => {
-  const [isBannerOpen, setBannerVisibility] = useState(true);
-  const { pathname } = useLocation();
-  const { i18n } = props;
-  const { about } = createWhoOwnsWhatRoutePaths();
-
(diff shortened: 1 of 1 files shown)


kiwansim (commented):
currently the PR only removes `allowChangingPortfolioMethod` from the `App` but leaves it in the `AppBody` and `NavBar`. I think we should be consistent about keeping the variable and disabling it or removing the variable entirely.

Do we want the infrastructure for switching portfolio methods available on dev machines? Especially with portfolio pruning changes, I'm wondering if the toggle capability might be reintroduced sooner than later.

austensen:
For now we are keeping the legacy version, but just removing the banner. So now you'll only be able to switch to it via the link on the about page. Then with the release of portfolio pruning (or shortly after, Joel's still thinking about this) we'll probably remove it entirely. So for now I think those `allowChangingPortfolioMethod` vars to change the styling and the route are still needed. Is that right?

kiwansim (approved):
re: keeping in the`allowChangingPortfolioMethod`: ah yep, you're right. totally missed `"wowza-styling"` being dependent on that variable

PR #854, opened 2024-02-29
Add validation/error handling to account settings update credentials

austensen:
This PR adds input validation and error handling to the forms for updating user email and password on the account settings page to ensure the new email is valid and not already in use, and that the new password satisfies rules. It swaps in the new `<EmailInput>`, and adapts the existing validation/error handling logic from `<Login> to show input- and page-level alerts. It prevents the default form reload and controls the toggling of edit mode after successful submission. Also adds an option to show an input-level error for password when rules aren't shown. Also fixes some spacing/styles that might have gotten messed up with the modal login changes. Also adds an input-level error state for password when the rule validation messages aren't shown (eg. for empty input when logging in).

[sc-13684]

https://www.loom.com/share/82f689b3757e43c098bc03336aee6351?sid=37708fe6-6c1a-4674-a483-c0a4a514c9d3

Update: There is now a "callout" box under email when not editing if you haven't verified your email. And when you successfully change your email we mark it un-verified and send a verify email on the server side (https://github.com/JustFixNYC/auth-provider/pull/17) and then the callout appears. 

image

 [sc-13772]

### DIFF
--- a/client/src/components/PasswordInput.tsx
+++ b/client/src/components/PasswordInput.tsx
@@ -86,7 +86,7 @@ const PasswordInputWithoutI18n = forwardRef<HTMLInputElement, PasswordInputProps
             </LocaleLink>
           )}
         </div>
-        {showPasswordRules && (
+        {showPasswordRules ? (
           <div className="password-input-rules">
             {passwordRules.map((rule, i) => {
               let ruleClass = "";
@@ -108,14 +108,16 @@ const PasswordInputWithoutI18n = forwardRef<HTMLInputElement, PasswordInputProps
               );
             })}
           </div>
-        )}
-        {!showPasswordRules && showError && error && (
-          <div className="password-input-errors">
-            <span id="input-field-error">
-              <AlertIcon />
-              {i18n._(t`Please enter password.`)}
-            </span>
-          </div>
+        ) : (
+          showError &&
+          error && (
+            <div className="password-input-errors">
+              <span id="input-field-error">
+                <AlertIcon />
+                {i18n._(t`Please enter password.`)}
+              </span>
+            </div>
+          )
         )}
         <div className="password-input">
           <input
--- a/project/settings.py
+++ b/project/settings.py
@@ -99,10 +99,6 @@ CORS_ALLOWED_ORIGIN_REGEXES = [
     r"https://deploy-preview-(?:\d{1,4})--wow-django-dev\.netlify\.app",
 ]
 
-CORS_ALLOWED_ORIGIN_REGEXES = [
-    r"https://\w+\.deploy-preview-(?:\d{1,4})--wow-django-dev\.netlify\.app"
-]
-
 # This is based off the default Django logging configuration:
 # https://github.com/django/django/blob/master/django/utils/log.py
 LOGGING = {


client/src/components/PasswordInput.tsx
```diff
@@ -82,26 +82,41 @@ const PasswordInputWithoutI18n = forwardRef<HTMLInputElement, PasswordInputProps
             <LocaleLink
               to={`${account.forgotPassword}?email=${encodeURIComponent(username || "")}`}
             >
-              Forgot your password?
+              {i18n._(t`Forgot your password?`)}
             </LocaleLink>
           )}
         </div>
         {showPasswordRules && (
```
kiwansim, line 89: thoughts on restructuring like this for more clarity?
```
showPasswordRules 
   ? <div className="password-input-rules"> 
   : {showError && error && <div className="password-input-errors">}
```

project/settings.py
```diff
@@ -99,6 +99,10 @@ def get_required_env(key: str) -> str:
     r"https://deploy-preview-(?:\d{1,4})--wow-django-dev\.netlify\.app",
 ]
 
+CORS_ALLOWED_ORIGIN_REGEXES = [
```
kiwansim, line 102: looks like this is a duplicate of one of the regexes above?

austensen: ah thanks, got duplicated in a merge update

client/src/components/PasswordInput.tsx
austensen: oh yeah that's much better - thanks!

client/src/components/UserSettingField.tsx
```diff
@@ -67,30 +116,87 @@ type EmailSettingFieldProps = withI18nProps & {
...
+    setShowEmailError(false);
+
+    if (email === oldEmail) {
+      return;
+    }
+
+    if (!email || emailError) {
+      setEmailError(true);
+      setShowEmailError(true);
+      throw new Error("Email format error");
```
kiwansim, line 142: where would you see these exceptions thrown? in the browser console? just confused about the utility of the exception here and in the `handleSubmit` below since we are also setting the error states accordingly

kiwansim (approved):
just a q about if the exceptions thrown are necessary. otherwise looks good on the demosite & in code!

client/src/components/UserSettingField.tsx
```diff
@@ -101,39 +207,52 @@ export const EmailSettingField = withI18n()(EmailSettingFieldWithoutI18n);
...
+  const { title, preview, onSubmit, children, verifyCallout } = props;
   const [editing, setEditing] = useState(false);
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    onSubmit().then(
+      (response) => setEditing(false),
+      (error) => {}
+    );
+  };
```
austensen, line 225: This is the part that needs those errors thrown

client/src/components/UserSettingField.tsx
austensen: ugh yeah this was an issue with how these email and pw fields are nested within the form wrapper component and the `OnSubmit` gets passed around a lot (from the components props, to the `handleSubmit` to the props of the wrapper, to be called in `handleSubmit` in that wrapper. And so I couldn't just `return` at those error steps without the rest of the wrapper's `handleSubmit` running, which included the closing of the edit dropdown. (see my other comment)
I think this whole setup with the wrapper makes this confusing, but short of rewriting all of that this was the only way I could think of to get that working. Agree that it's not ideal still though. Would be nice to refactor maybe once we have the new form component library elements.

PR #855, opened 2024-03-04
Remove onsite landing page for email verification

kiwansim:
changes from previous implementation: 
* listener hook is called upon the verification link being clicked
* immediately shows success message if verification is successful / user is already verified
* user login is not required for verification to be successful

bugs:
* link expired case is still not working as expected but wanted to get this PR in because it makes the verification process more seamless for QA. will fix this issue after email deliverability QA.


austensen (approved):
This looks great! Good idea to just get this merged now and deal with the already verrified issue later.

PR #857, opened 2024-03-05
Button integration [sc-13707]

steve52:
Preview: https://deploy-preview-857--wow-django-dev.netlify.app/en/

https://app.shortcut.com/justfixnyc/story/13707/use-new-button-components-in-wow


client/src/components/SocialShare.tsx
```diff
@@ -46,6 +45,37 @@ type SocialShareProps = {
...
+    variant="secondary"
+    size="small"
+    labelIcon={IconTwitter}
+    iconOnly
+    {...props}
+  />
+);
+const _EmailButton: React.FC = (props) => (
+  <Button
+    labelText={`${i18n._("Email")} (Opens a new window)`}
```
austensen, line 70: If you could use this `i18n._("Email")` format for all of the label text for these three that will ensure the translations work.

client/src/components/PortfolioTable.tsx
```diff
@@ -696,34 +697,36 @@ const PortfolioTableWithoutI18n = React.memo((props: PortfolioTableProps) => {
               </span>
             </div>
             <div className="prev">
-              <button
-                className="page-btn"
+              <Button
```
kiwansim, line 700: currently the disabled state has a drop shadow on hover. is this expected? here's the [figma](https://www.figma.com/file/WZ5AOU0jwtV9PH2d314Sf0/JustFix-Design-System?type=design&node-id=1296%3A1228&mode=dev) i'm looking at for reference

client/src/components/SocialShare.tsx
```diff
@@ -46,6 +45,37 @@ type SocialShareProps = {
   customContent?: SocialShareContent;
 };
 
+const _FacebookButton: React.FC = (props) => (
```
kiwansim, line 48: is the leading underscore meant to denote that this is a private variable? it's a convention I haven't seen in other places of our code but perhaps this is best practice we should start doing

client/src/containers/NotRegisteredPage.tsx
```diff
@@ -32,6 +33,10 @@ export const SocialShareForNotRegisteredPage = () => (
   </div>
 );
 
+const JFCLLinkInternal: React.FC<any> = (props) => {
```
kiwansim, line 36: since we seem to be redeclaring `JFCLLinkInternal` in several files, thoughts on moving it to a `helper.ts` and just importing it to these pages?

kiwansim:
I know we are leaving a few buttons on the building Overview page as is (e.g. "View data over time" and "Take action on JustFix.org!") but wanted to check if the "Download" button in the Export Data modal was intentionally left with the previous styles as well Screenshot 2024-03-12 at 11 22 22 AM

client/src/components/SocialShare.tsx
steve52: For sure! Do I need to run anything to get it into the dictionary? Or do you want to handle that later?

client/src/containers/NotRegisteredPage.tsx
steve52: Yeah totally agree 👍 .

steve52: Actually, maybe it should just be in the /components folder? `helper.ts` is mostly utility functions and react hooks.

What do you think?

steve52:
Not sure about the "Download" button in the Export Data modal. I don't think it was in the requested set of updates.

@austensen @lesliejustfix Is this another button I should swap out?

client/src/containers/NotRegisteredPage.tsx
kiwansim: to clarify, I was thinking we can make a new `helper.ts` (or something equivalent) in the `/components` folder. are you suggesting we put this function in a standalone file (i.e. `JFCLLinkInternal.tsx`)? that makes sense to me too!

client/src/components/SocialShare.tsx
kiwansim: I think we're going to run the localization script for the whole feature branch all at once when we're ready for prod. but if you ever do need it -- the script is `yarn lingui extract`

steve52: Thanks, good to know!

client/src/containers/NotRegisteredPage.tsx
steve52: Ah, I see. 

Yeah, I was thinking of just putting it in a standalone file. There's already a couple of other small components in there, like `CloseButton.tsx`.

kiwansim: that sounds good to me! 🚀

kiwansim:
@lesliejustfix I recall the initial mocks for swapping out button styles on WOW leaving a few of the button styles as is, but I may have assumed incorrectly!

client/src/containers/NotRegisteredPage.tsx
steve52: Okay, moved this to it's own file!

steve52:
@kiwansim @lesliejustfix 'view over time' and 'take action' have been swapped out for our Link component.

The Download button is a bit trickier. It's using a library to generate the button right now and it's not as easy as it was with `react-router-dom` to swap out the underlying component :/ .

client/src/components/PropertiesSummary.tsx
```diff
@@ -51,19 +52,16 @@ const DataRequestButton: React.FC<{
   return (
     <I18nComponent>
       {({ i18n }) => (
-        <a
+        <Link
```
kiwansim, line 55: think this link needs `icon="external"`

kiwansim (approved):
smaller additional changes to come in a separate PR

PR #858, opened 2024-03-05
update unsubscribe page for unsubscribe-all via email

austensen:
We'll have two links in our emails - **Unsubscribe All** and **Manage Subscriptions**

Both will go to WOW `/unsubscribe` page with a param `u=<token>` and the "all" version will also have `all=true`.

If the token and `all=true` are present, it should automatically call `AuthClient.emailUnsubscribeAll(token)` and show a success messaging confirming you've unsubscribed from all building updates. 

If the token is present but not `all`, it should do what its' already doing: automatically fetch user subscriptions via `AuthClient.emailUserSubscriptions(token)` and display them all with remove buttons to call `AuthClient.emailUnsubscribeBuilding(token)`. And there will also be a button at the top to 'Unsubscribe From All'.

If there are no subscriptions but not because you had the `all=true` option, have message saying no subscriptions.

In either situation where there are no subscriptions, prompt user to subscribe to buildings or delete their account (same as on account settings page).

Edge case: What if you land on this page without the token param? Maybe redirect to login, as we do for all other pages like this when you aren't logged in. We'll wait to deal with that all at once in [sc-13754]

In a separate PR we'll add a button for the login-auth version "unsubscribe from all" to the Account Settings page. [sc-13811]

Waiting on design review. I just guessed about using the elements from account settings, and need to see how the unsubscribe-all button should be styled (just a placeholder for now)

image

image

image

[sc-13732]

### DIFF
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -186,15 +186,22 @@ const isEmailAlreadyUsed = async (email: string) => {
 /**
  * Sends an unauthenticated request to unsubscribe the user from the building
  */
-const emailBuildingUnsubscribe = async (bbl: string, token: string) => {
+const emailUnsubscribeBuilding = async (bbl: string, token: string) => {
   return await postAuthRequest(`${BASE_URL}auth/unsubscribe/${bbl}?u=${token}`);
+};
+
+/**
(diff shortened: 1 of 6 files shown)


kiwansim (approved):
looks good to me! ✨

PR #859, opened 2024-03-06
change privacy info modal to new-tab links

austensen:
Remove the info icon opening modal and instead just use links to privacy and terms of use page (opening in new tab)

And only show this when registering new account, not when logging in.

image

image

[sc-13835]

### DIFF
--- a/client/src/components/Login.tsx
+++ b/client/src/components/Login.tsx
@@ -139,16 +139,29 @@ const LoginWithoutI18n = (props: LoginProps) => {
   const renderFooter = () => {
     return (
       <div className="building-page-footer">
-        <div className="privacy-modal">
-          <Trans>Your information is secure</Trans>
-          <button
-            className="info-icon"
-            onClick={() => setShowInfoModal(true)}
-            aria-label={i18n._(t`Learn more about how we use your data`)}
-          >
-            <InfoIcon />
-          </button>
-        </div>
+        {isRegisterAccountStep && (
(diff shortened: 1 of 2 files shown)


client/src/components/Login.tsx
```diff
@@ -139,16 +137,29 @@ const LoginWithoutI18n = (props: LoginProps) => {
...
+              <a
+                target="_blank"
+                rel="noopener noreferrer"
+                href="https://www.justfix.org/en/terms-of-use/"
+              >
+                Terms of Service
+              </a>
+              .
+            </Trans>
+          </span>
```
kiwansim, line 161: this might be a question for Leslie, but I wonder if an external link icon would be helpful to add to the end of `privacy-links` span. not only to set the expectation that it will open a new tab, but also to differentiate it from the "Login" link below with the same link styling but has a different behavior. 

Screenshot 2024-03-06 at 6 38 45 PM

kiwansim (approved):
comment isn't a blocker! looks 💯

client/src/components/Login.tsx
austensen: Oh yeah leslie and steve were talking about usage of this icon recently and this is a great example of it - I don't remember what they landed on but lets check in about it today

lesliejustfix: We could make log in a standalone internal link, either below or to the right of "Already have an account?"

PR #864, opened 2024-03-15
update hpd complaints table for alerts api

austensen:
Update the HPD Complaints table we're using in the email alerts data api. Now using the new combined complaints and problems, and so we'll be counting problems (though still calling them complaints. This matches what we do on WOW already, how they are displayed on HPD Online, and  DAP Portal.

[sc-14005]


kiwansim (approved):
🛠️

PR #842, opened 2024-01-26
show modal & prevent subscription above 15 bldg limit

austensen:
Prevent users from signing up for more than 15 buildings, and show modal explaining this when they click "get updates" and direct users to edit their list of building in account settings and also link to typeform survey where they can request access to more buildings. 

image
image
image

[sc-13353]
[sc-13357]
[sc-13356]

PR #866, opened 2024-03-18
limit subscriptions to 15 (not 16)

austensen:
Limit users to 15 subscriptions (fix typo from #842 that was allowing 16)


kiwansim (approved):
1️⃣5️⃣

PR #867, opened 2024-03-20
Email verification landing page states, expiration, resend email without auth

kiwansim:
Relates to https://github.com/JustFixNYC/auth-provider/pull/30

[sc-14008]
[sc-13981]
[sc-13982]
[sc-13994]

### DIFF
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -236,9 +236,14 @@ const postAuthRequest = async (
 
   try {
     if (result.ok) {
+      console.log("result is ok");
+      return await result.json();
+    } else {
+      console.log("result is not ok");
       return await result.json();
     }
   } catch {
+    console.log("in error");
     return result;
   }
 };
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -131,19 +131,6 @@ const resendVerifyEmail = async (token?: string) => {
   }
 };
 
-// /**
-//  * Sends unauthenticated request for a new verification email
-//  * Used only in expired verification flow in logged out state
-//  */
-// const resendVerificationWithoutAuth = async (token: string) => {
-//   try {
-//     await postAuthRequest(`${BASE_URL}auth/token_resend_verify_email?u=${token}`);
-//     return true;
-//   } catch {
-//     return false;
-//   }
-// };
-
 /**
  * Sends an authenticated request to update the user email
  */
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -118,14 +118,31 @@ const verifyEmail = async () => {
 /**
  * Sends request to resend the account verification link to the user's email
  */
-const resendVerifyEmail = async () => {
+const resendVerifyEmail = async (token?: string) => {
   try {
-    await postAuthRequest(`${BASE_URL}auth/resend_verify_email`);
+    if (token) {
+      await postAuthRequest(`${BASE_URL}auth/resend_verification_with_token?u=${token}`);
+    } else {
+      await postAuthRequest(`${BASE_URL}auth/resend_verification`);
+    }
     return true;
   } catch {
     return false;
   }
 };
+
+// /**
+//  * Sends unauthenticated request for a new verification email
+//  * Used only in expired verification flow in logged out state
+//  */
+// const resendVerificationWithoutAuth = async (token: string) => {
+//   try {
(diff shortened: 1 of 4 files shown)


client/src/components/AuthClient.ts
```diff
@@ -118,9 +118,13 @@ const verifyEmail = async () => {
 /**
  * Sends request to resend the account verification link to the user's email
  */
-const resendVerifyEmail = async () => {
+const resendVerifyEmail = async (token?: string) => {
```
kiwansim, line 121: optional token would only be included from the "verification link has expired" flow

client/src/components/AuthClient.ts
```diff
@@ -237,6 +241,8 @@ const postAuthRequest = async (
   try {
     if (result.ok) {
       return await result.json();
+    } else {
```
kiwansim, line 244: don't know how this else clause got removed, but this is necessary for any non-200 result status code and text to be returned. 

checked that returning the non-200 result json works on other `AuthClient` functions that use `postAuthRequest.`

client/src/containers/VerifyEmailPage.tsx
```diff
@@ -2,55 +2,62 @@ import { useState, useEffect } from "react";
...
+  const delaySeconds = 5;
+  const baseUrl = window.location.origin;
+  const redirectUrl = `${baseUrl}/${i18n.language}`;
+
+  const updateCountdown = () => {
+    let timeLeft = delaySeconds;
+    const delayInterval = delaySeconds * 100;
+
+    setInterval(() => {
+      timeLeft && timeLeft--; // prevents counter from going below 0
```
kiwansim, line 53: mostly for slow speeds

client/src/containers/VerifyEmailPage.tsx
```diff
@@ -72,16 +79,28 @@ const VerifyEmailPage = withI18n()((props: withI18nProps) => {
         <Trans render="h4">Check your email inbox & spam</Trans>
         <br />
         <Trans>
-          Click the link we sent to verify your email address {!!user && user.email}. It may take a
-          few minutes to arrive. Once your email has been verified, you’ll be signed up for Data
-          Updates.
+          Click the link we sent to verify your email address. It may take a few minutes to arrive.
```
kiwansim, line 82: removing the user email here since user could be in expired verification link flow while not logged in

client/src/containers/VerifyEmailPage.tsx
```diff
@@ -90,11 +109,36 @@ const VerifyEmailPage = withI18n()((props: withI18nProps) => {
...
+      <a href={redirectUrl}>{redirectUrl}</a>
+    </div>
+  );
+
   return (
     <Page title={i18n._(t`Verify your email address`)}>
       <div className="TextPage Page">
         <div className="page-container">
-          {isVerified ? successPage() : isExpired && expiredLinkPage()}
+          {!loading &&
```
kiwansim, line 134: `loading` var is necessary to prevent flickering of a different state as the AuthClient async calls are getting resolved. 

the nested ternary operator is gross but i didn't have any other ideas. lmk if you do!

client/src/components/AuthClient.ts
austensen: oh yeah that's great to have for error handling

client/src/containers/VerifyEmailPage.tsx
```diff
@@ -59,7 +66,7 @@ const VerifyEmailPage = withI18n()((props: withI18nProps) => {
         <button
           className="button is-secondary"
           onClick={async () => {
-            setIsEmailResent(await AuthClient.resendVerifyEmail());
+            setIsEmailResent(await AuthClient.resendVerifyEmail(token));
           }}
         >
           <Trans>Resend verification email</Trans>
```
austensen, line 72:
```suggestion
          <Trans>Send new link</Trans>
```
Based on the [new designs](https://www.figma.com/file/Gv7RdLkCV21Z5FaY7RrRnH/Email-Alerts---Working-File-v2-%F0%9F%9B%A0%EF%B8%8F?type=design&node-id=39-1823&mode=design&t=iL55bjJ6gtVFQEaq-0) and conversations about the reset password flow, it sounds like Leslie wants all resend buttons to have this same text.

austensen (approved):
This looks great! And tested it out locally and al the different states worked correctly

client/src/containers/VerifyEmailPage.tsx
kiwansim: oo good catch, thank u

PR #868, opened 2024-03-21
add reset password landing pages, validate new password inputs

austensen:
This PR changes the forgot/reset password flows, and goes along with changes to the auth backend in https://github.com/JustFixNYC/auth-provider/pull/31

From login you can click "forgot password?", your email is autofilled. This now has input validation to ensure a valid email (though it doesn't check for an existing account).

image

Submitting your email here will send a reset link to your email, and show you a success page.

image

Then from your email you click the link back to wow with the reset token in url params. 

We now check that token immediately when you land on the page (and while waiting it shows the standard `Loading...` placeholder. If the token is valid and active still, it changes to reset password.
image

This now has full input validation for the password, same as for account creation. 

If that the reset is successful it shows this countdown then redirects you to login page. 

image

That's the "happy path" for this. 

If your email link is expired, you will get this message and a one-click button to resend the email, which uses the token to send the request using a new auth endpoint (https://github.com/JustFixNYC/auth-provider/pull/31)

image

After clicking to resend the link you get this success page:

image

If anything else goes wrong in this process - the link is broken in some other way, missing completely from url params, or if the password reset fails for some reason (maybe times out between the check and the submit) - then you get this page, and the button sends you back to the "forgot password" page where you can enter you email and start again. 

image

I tried to follow a similar process as used for the verify email flow in https://github.com/JustFixNYC/who-owns-what/pull/867. With the handling of status codes in the auth client, the useEffect checking the token, etc.

[sc-13900]
[sc-13986]
[sc-13901]

### DIFF
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -18,6 +18,18 @@ type VerifyEmailResponse = {
   error?: string;
 };
 
+export enum ResetStatusCode {
+  Success = 200,
+  Accepted = 202,
+  Invalid = 400,
+  Expired = 404,
+  Unknown = 500,
+}
+
+type PasswordResetResponse = Omit<VerifyEmailResponse, "statusCode"> & {
+  statusCode: ResetStatusCode;
+};
+
 let _user: JustfixUser | undefined;
 const user = () => _user;
 const fetchUser = async () => {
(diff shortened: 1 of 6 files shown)


kiwansim (approved):
looks good! ✨

austensen:
> looks good! ✨

Thanks - made it easy using your verify work as a reference!

PR #869, opened 2024-03-22
Content updates

austensen:
Various content changes, updating from the main content spreadsheet
[sc-13967]

Left out some content changes for verification and password reset pages since those also need some other changes
[sc-14144]
[sc-14145]


kiwansim (approved):
sorry for the delay! It all looks good 💯

PR #871, opened 2024-03-25
fix user object condition

austensen:
Found another case of conditioning on the `user` object that is now evaluating true when it's actually just the empty placeholder. This is another case of the issue fixed in #870 . It was causing the account settings page to appear but with all the user content missing. Now it correctly shows nothing, like before


kiwansim (commented):
when I register or login from the nav, the register/login call returns a 200 but the site shows a blank screen and the nav bar shows "Login" until I manually refresh. 

did you see this happen on your local instance? 

https://www.loom.com/share/f50fce9353fb453a92d1da7725355fc9

austensen:
After looking more closely I don't think this is actually causing the same kind of issue as the user object condition was on login. And once we implement the redirect based on on login status it would replace this condition anyway

PR #872, opened 2024-03-25
refactor unsubscribe page

austensen:
Refactor the unsubscribe/manage subscriptions page that you land on via email links.

When you first land on the page there is this standard wow placeholder while it fetches your subscriptions to be able to determine what to show. 
image

>Note: the plural words (are/is, building/buildings) are missing from these screenshots since we need to run all the internationalization which we are doing after merging in. 

Click "manage subscriptions" link from email, while having subscriptions:
image

Click the unsubscribe all button from "manage subscriptions" page, or coming from email when you have 0 subscriptions:
image

Click on "unsubscribe" link from email when you have subscriptions:
image

Click the unsubscribe all button from "unsubscribe" page, or coming from email when you have 0 subscriptions:
image

[sc-14085]


kiwansim (approved):
image

PR #873, opened 2024-03-26
add component library buttons, fix verified condition on building page

austensen:
I was starting to implement the component-library buttons, then Leslie flagged that the condition was wrong on the building pages. When you're not verified you should always see the prompt to verify and not allow users to subscribe until they are. previously it was allowing them to subscribe then the prompt to verify would show up after. So this fixes that plus a few button swaps that will get continued in another PR.

### DIFF
--- a/client/src/styles/Login.scss
+++ b/client/src/styles/Login.scss
@@ -17,7 +17,7 @@
     .submit-button-group {
       display: flex;
 
-      .button.is-primary {
+      .jfcl-button.jfcl-variant-primary {
         padding: 1rem 2rem !important;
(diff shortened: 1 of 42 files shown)


client/src/styles/Login.scss
```diff
@@ -95,20 +95,20 @@
     }
   }
 
-  .jf-alert {
-    width: auto;
-    font-size: 0.81rem;
+  // .jf-alert {
```
kiwansim, line 98: delete this commented out block

kiwansim (approved):
🥇

PR #878, opened 2024-03-27
Layout change and other style updates 

kiwansim:
- moves Building Updates card to right column
- update style of LOC Get Repairs callout 
- adds "View Data Over Time" link to the bottom of the Building Stats Table for better visual grouping 
- moves Google Map static image to bottom of the right column and hides in mobile view

[sc-13947]


client/src/components/Icons.tsx
```diff
@@ -204,3 +204,12 @@ export const ArrowIcon = (props: ArrowIconProps) => {
     </span>
   );
 };
+
+export const AddIcon = (props: SVGProps<SVGSVGElement>) => (
```
kiwansim, line 208: not happy to be adding this here but didn't wanna do it via component library at this moment

client/src/styles/Card.scss
```diff
@@ -0,0 +1,63 @@
+@import "_vars.scss";
```
kiwansim, line 1: my attempt to generalize the card component

client/src/styles/DetailView.scss
```diff
@@ -314,12 +286,6 @@
       }
     }
 
-    .card-body-links {
```
kiwansim, line 317: don't know what this was really doing, but needed to get rid of it for proper padding between cards in mobile

client/src/styles/DetailView.scss
```diff
@@ -213,23 +196,12 @@
 
     @include datatable();
 
-    .card-body-table-prompt {
-      @include for-phone-only() {
-        display: none;
-      }
-    }
-
-    .card-body-timeline-link,
```
kiwansim, line 222: `card-body-timeline-link` doesn't exist as a class anymore

client/src/components/GetRepairs.tsx
```diff
@@ -0,0 +1,56 @@
...
+              <div className="table-small-font">
+                <label className="card-label-container">
+                  <Trans>Need repairs in this building?</Trans>
+                </label>
+                <div className="table-content">
+                  <Trans render="div" className="card-description">
+                    Send a free, legally vetted letter to notify your landlord of repair issues
+                  </Trans>
+
+                  <JFCLLink
```
kiwansim, line 30: janky but does the job for now. let me know if you have a different idea for a pattern to follow?

client/src/components/BuildingStatsTable.tsx
```diff
@@ -208,8 +225,11 @@ const BuildingStatsTableWithoutI18n = (props: { addr: AddressRecord }) => (
         <EvictionFilings />
         <EvictionsExecuted />
       </div>
+      <div className="table-row timeline-link">
+        <TimelineLink url={props.timelineUrl} />
```
kiwansim, line 229: @austensen if you pull this branch, there are red underlines under the `url` variable about mismatched types. would love your eyes on this bc i think it's a simple thing i'm missing

austensen: It need `props: {}` in the args of the functional component - just pushed an update.

client/src/styles/Card.scss
austensen: love this, so much cleaner

client/src/components/Login.tsx
```diff
@@ -420,7 +421,12 @@ const LoginWithoutI18n = (props: LoginProps) => {
             labelText={i18n._(t`Email address`)}
           />
           <div className="submit-button-group">
-            <Button variant="primary" size="small" labelText={i18n._(t`Get updates`)} />
+            <Button
```
austensen, line 424: I forgot to ask you about this when I first added the component library buttons here, but for some reason only with the buttons within these overview cards the `cursor: pointer` is not working properly. Specifically for the `jfcl-button__label` element. Even though when inspecting it shows that it is inheriting `pointer` from the button, and nothing else should be affecting it. You have any idea what might be causing this? 
image 
image

I'll just make a ticket for this and we can address later https://app.shortcut.com/justfixnyc/story/14192

client/src/styles/EmailAlertSignup.scss
```diff
@@ -71,6 +12,10 @@
 
     .jfcl-button {
       align-self: center;
+
+      .jfcl-button__label {
+        align-items: center;
+      }
```
austensen, line 18: good catch - this should probably get added to the component library, right?

client/src/components/GetRepairs.tsx
austensen: actually I was just talking to Leslie about this and she decided this should just be a normal component-library external link https://app.shortcut.com/justfixnyc/story/14183/link-vs-button-questions

client/src/components/DetailView.tsx
```diff
@@ -414,25 +397,24 @@ class DetailViewWithoutI18n extends Component<Props, State> {
...
                   </div>
                   <div className="column col-lg-12 col-5">
-                    <div className="card-image hide-lg">{streetView}</div>
-                    <div className="card-body column-right">
+                    <EmailAlertSignup
+                      bbl={detailAddr.bbl}
+                      housenumber={detailAddr.housenumber}
+                      streetname={detailAddr.streetname}
+                      zip={detailAddr.zip || ""}
+                      boro={detailAddr.boro}
```
austensen, line 405: it really doesn't matter, but if we wanted I think both `<EmailAlertSignup>` and `<BuildingSubscribe>` could just take `props: { addr: AddressRecord }`

austensen (approved):
This looks amazing! Also appreciate all the cleaning up of repetitive/confusing styles etc.

client/src/styles/EmailAlertSignup.scss
austensen: added a story https://app.shortcut.com/justfixnyc/story/14195

client/src/components/DetailView.tsx
kiwansim: that's a great catch. will make that change!

client/src/components/Login.tsx
kiwansim: i think it's because of the way I overrode the datatable mixin in the first few rows of Card.scss. It was an easy fix so it will be in the next commit

kiwansim:
now also addresses [sc-14192]

PR #875, opened 2024-03-27
Use component library buttons in email alerts

austensen:
This PR swaps in component-library buttons for all the email alerts work, and creates `<SendNewLink>` that toggles between a "send new link" button -> "✔️ link sent" confirmation. There are also some changes to the reset password and verify pages for expired links for consistency and to update for content doc. 

[sc-13933]
[sc-14169]

PR #879, opened 2024-03-27
More component library button updates

austensen:
Was a bit too hasty in merging in #875 , so this fixes a couple things that either broke or were left out, plus some changes we decided on after. 

* fix some naming consistency for `isEmailResent` that I accidentally left out
* change "signup" toggle back to old style text button
* change user setting edit button to be component library text button
  * fix typo in cancel button type, "button" not "submit"
* change the unsubscribe-all on unsubscribe page to secondary variant

[sc-13933]
[sc-14183]

### DIFF
--- a/client/src/components/UserSettingField.tsx
+++ b/client/src/components/UserSettingField.tsx
@@ -120,7 +120,7 @@ const EmailSettingFieldWithoutI18n = (props: EmailSettingFieldProps) => {
   const { i18n, currentValue, onSubmit } = props;
   const userContext = useContext(UserContext);
   const { email: oldEmail, verified } = userContext.user as JustfixUser;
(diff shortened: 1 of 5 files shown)


client/src/components/UserSettingField.tsx
```diff
@@ -238,7 +238,7 @@ const UserSettingFieldWithoutI18n = (props: UserSettingFieldProps) => {
             <div className="user-setting-actions">
               <Button type="submit" variant="primary" size="small" labelText={i18n._(t`Save`)} />
               <Button
-                type="submit"
+                type="button"
```
kiwansim, line 241: what is the point of specifying `type="button"` on Buttons that aren't `submit` or `reset`? just wondering if i should also have been doing this

kiwansim (approved):

PR #881, opened 2024-03-28
fix input error spacing, add util classes

austensen:
This PR adds [margin and padding utility classes](https://github.com/JustFixNYC/justfix-website/blob/master/src/styles/_custom.scss) for design-system spacing increments that we create for org site. (later this should probably move to component library)
[sc-14221]

Spectre had some of these too, so I commented those out (until we can finally move away from all spectre stuff) and updated any references to them to use our new ones instead. 

Then I used the new spacing classes to fix the input error spacing for email/password/user-type.
[sc-13887]

And finally I moved the "forgot password?" link from the password component to login to match the updated placement in figma. 
[sc-13974]

image

image

image

UPDATE:

Also now all the input errors deal with long messages. svg's are wrapped in div to prevent shrinking within flexbox, and align icons to top. 

image
image

### DIFF
--- a/client/src/components/Login.tsx
+++ b/client/src/components/Login.tsx
@@ -3,7 +3,11 @@ import { Trans, t } from "@lingui/macro";
 import { I18n } from "@lingui/core";
 import { withI18n } from "@lingui/react";
 import classNames from "classnames";
+import { Button, Link as JFCLLink } from "@justfixnyc/component-library";
 
+import "styles/Login.css";
+import "styles/UserTypeInput.css";
+import "styles/_input.scss";
 import AuthClient from "./AuthClient";
 import { JustfixUser } from "state-machine";
 import { UserContext } from "./UserContext";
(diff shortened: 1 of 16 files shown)


client/src/components/Login.tsx
```diff
@@ -177,19 +180,32 @@ const LoginWithoutI18n = (props: LoginProps) => {
             </Trans>
           </span>
         )}
+        {isLoginStep && (
+          <LocaleLink
```
kiwansim, line 184: is there a reason why you used `LocaleLink` instead of `Link` from the component library here?

austensen: it does use it within LocaleLink with the `component={JFCLLink}` so it gets the styles, but I used LocaleLink so that it uses the proper `/en` `/es` on that link, which I believe we do everywhere in wow

austensen: i guess it would actually be better to just edit the LocaleLink to always use the jfcl component like this to simplify all our uses of it since this is kinda gross currently

kiwansim: oh and does this cover [sc-13974](
https://app.shortcut.com/justfixnyc/story/13974/move-forgot-password-link-out-of-the-input-field?vc_group_by=day&ct_workflow=all&cf_workflow=500000153)?

kiwansim: yeah I think editing LocaleLink would be cleaner! we can make a ticket for it though unless you wanna include it in this pr

austensen: Hmm actually I don't know if we're ready to use that in every location yet and it might take some time to review everything, I'll make a ticket for this to do later though

kiwansim (approved):

client/src/components/Login.tsx
austensen: oh yes, missed that ticket, thanks

PR #887, opened 2024-03-29
use 16px in text inputs to prevent zooming on mobile safari

austensen:
On mobile safari it was zooming in on text inputs when focused. This is because the font size was < 16px, so this bumps them all up to 16px (until we can swap with component library). Testing on the preview link it does not zoom now!

desktop:
image

mobile:
image

[sc-13908]

### DIFF
--- a/client/src/styles/EmailInput.scss
+++ b/client/src/styles/EmailInput.scss
@@ -30,6 +30,7 @@
 
     input {
       flex: 1;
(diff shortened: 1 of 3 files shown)


kiwansim (approved):
🔎 🔍 🔎 🔍 🔎 🔍 🔎 🔍 🔎 🔍

PR #889, opened 2024-04-01
update HPD Online link for bbl search results, useful links and alerts api

austensen:
When the new HPD Online first launched it wouldn't allow you to link to search results based on BBL like the old version. This is important because since WOW uses BBLs for our pages, HPD uses their own "HPD building ID" and a single BBL can have multiple buildings. When there is only one building we can know that and link directly to the page, but when there are multiple we need to link to these bbl search results and the user can select the building they are interested in. 

image

Now that HPD allows these links, we update our own "useful links" section. 

Also in this PR I add this link (single building or search results) as an optional "indicator" on the email alerts endpoint so that we can include those in the building update emails just like useful links. 

The corresponding PR that makes use of the new api changes: https://github.com/JustFixNYC/auth-provider/pull/34

[sc-12997]
[sc-13976]

### DIFF
--- a/client/src/components/UsefulLinks.tsx
+++ b/client/src/components/UsefulLinks.tsx
@@ -41,7 +41,7 @@ export const UsefulLinks: React.FC<UsefulLinksProps> = ({ addrForLinks, location
             href={
               !!hpdbuildingid && hpdbuildings === 1
                 ? `https://hpdonline.nyc.gov/hpdonline/building/${hpdbuildingid}/overview`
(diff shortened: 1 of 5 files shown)


austensen:
great point! added the comment, and later we can refactor this whole endpoint to use a pre-built table as part of the wow k8s job that has all the values we'll need for email alerts and the endpoint can just take a bbl (or maybe multiple bbls if we want to just run one api request per user for all their building updates)

PR #890, opened 2024-04-01
Nav bar updates

kiwansim:
* nav bar buttons no longer wrap
* website title is left justified with the JF logo. this reflects the design system navigation mocks. however, I left the rest of the nav bar styles unchanged, as discussed in the comment section of [sc-14207]

## desktop, tablet landscape
Screenshot 2024-04-01 at 4 59 02 PM

## smaller than tablet landscape
Screenshot 2024-04-01 at 4 59 20 PM
Screenshot 2024-04-01 at 4 59 30 PM


client/src/assets/img/logo-divider.svg
kiwansim, line 1: it feels kind of janky to have a svg for the divider itself, but any other method I tried with borders, margins, paddings, divs, etc. all didn't span the whole height of the nav bar or made the logo & product name scaled down.

client/src/styles/App.scss
```diff
@@ -144,15 +142,9 @@
   }
 
   h1.page-title {
-    margin: 0;
+    @include mobile-h2;
```
kiwansim, line 145: also, don't love that the most important h1 in our site hierarchy  is styled as a mobile-h2 but alas

kiwansim:
[sc-14051]

client/src/styles/App.scss
austensen: ugh yeah, can't wait to just update all the type styles for design system eventually

client/src/styles/App.scss
austensen, line 1: The "demo site" tag could use a bit of `top-margin`, and also it causes the nav to wrap below 353px width. This is obviously pretty minor, but since we'll use the demo site for beta testers I just wanted to mention. 
image

austensen (approved):
This looks great! So nice to finally have "JustFix" on the site like this.

client/src/styles/App.scss
kiwansim: oo good catch, I totally overlooked the demo site tag. i'll include the fix in this PR since as you said, we're gonna use the demo site for beta testers

kiwansim: Screenshot 2024-04-01 at 7 03 05 PM
Screenshot 2024-04-01 at 7 02 52 PM

austensen: Looks great!

PR #891, opened 2024-04-02
add verify step to nav signup flow, same as in modal

austensen:
Previous the signup flow from Nav would end after user type selection and redirect users to the account settings page. This was confusing during initial user testing since it never said anything about email verification until later it was show on building pages, also there is no need to do anything on account settings after sign up. This PR removes that redirect and shows the same "verify email" prompt as we show at the end of the on-page/in-modal sign up flow. So the final step looks like this.
[sc-14126]

(I also remove a few old references to a "verify reminder" step that doesn't exist anymore)

> Note that we'll be changing the style of this page, including adding a link to WOW homepage, as part of the "standalone page" standardized design implementation [sc-14218]

image
image

### DIFF
--- a/client/src/components/Login.tsx
+++ b/client/src/components/Login.tsx
@@ -28,7 +28,6 @@ enum Step {
   RegisterAccount,
   RegisterUserType,
   VerifyEmail,
-  VerifyEmailReminder,
 }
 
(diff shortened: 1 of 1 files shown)


kiwansim (approved):
💯

PR #893, opened 2024-04-04
Standalone page

austensen:
This adds a standardized "standalone" page design for login/signup, forgot-password, reset-password, verify email (plus a modified version for unsubscribe/manage subscriptions). 
[sc-14218]

This new page also gets a new minimal footer.
[sc-13964]

Mobile styles are also fixed for all these pages with the new layout.
[sc-14066]

All automatic/countdown redirects are removed from login/signup via nav, reset-password and verify. 
[sc-14144]

The login flow from nav now has an extra final step. If not verified you get the same prompt to do so as at the end of sign up, and it sends a new email. And if you are already verified then you get a a new "success" step with a link to the homepage. 

Focus now automatically shifts to the password input after submitting email in the login/signup flows. 
[sc-14271]

Also, I realized there was a bad copy/paste error on the reset-password page - it was sending a verify email instead of a reset one! So I fixed that and also realized there was a problem with the request because cookies weren't getting sent and those are needed for accessing session on the auth server.


client/src/styles/EmailInput.scss
```diff
@@ -10,9 +10,9 @@
     label {
       @include body-standard;
```
kiwansim, line 11: this can just be `@include wow-body-standard` now with lines 13-17 deleted!

client/src/styles/Password.scss
```diff
@@ -10,9 +10,9 @@
     label {
       @include body-standard;
```
kiwansim, line 11: same here re: change to`@include wow-body-standard`

client/src/styles/StandalonePage.scss
```diff
@@ -0,0 +1,124 @@
...
+        gap: 1rem;
+        text-align: center;
+        margin-top: 1.5rem;
+
+        &,
+        a,
+        .button.is-text {
+          font-size: 0.875rem;
+          font-weight: 500;
+          color: #464d5d; // TODO: this isn't in design system, confirm it's the right one to use and add
```
kiwansim, line 106: just left a [comment](https://www.figma.com/file/l3LL5TxSncwTO88xvUcQAL?type=design&node-id=1109-1287&mode=dev#759789831) on figma

kiwansim (approved):
the refactored pages are so sleek!! this looks amazing 🫡

client/src/containers/App.tsx
```diff
@@ -264,7 +264,17 @@ const Navbar = () => {
...
+    "forgot-password",
+    "login",
+    "verify-email",
+    "forgot-password",
+    "reset-password",
+    "unsubscribe",
+  ];
+  const hideNavbar = standalonePages.some((v) => pathname.includes(`account/${v}`));
+
+  return hideNavbar ? null : (
```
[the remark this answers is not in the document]
kiwansim: hmm i hear ya. i can't think of a better way to do this logically but there is a way to avoid returning a null, like so. tried it locally and it works!

```
return (
  <>
    {!hideNavbar && (insert JSX element here)}
  </>
);
```

client/src/components/StandalonePage.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const StandalonePage = withI18n()((props: StandalonePageProps) => {
+  const { title, className, children } = props;
+  const { home } = createWhoOwnsWhatRoutePaths();
+  return (
+    <Page title={title}>
+      <div className={classNames("StandalonePage Page", className)}>
+        <div className="page-container">
+          <LocaleLink to={home}>
```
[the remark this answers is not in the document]
austensen: This is so helpful, thanks for catching the issue and spelling out the fix so thoroughly!

PR #894, opened 2024-04-05
Fix flickering states in building updates box

kiwansim:
initially tried displaying a loading state in the parent component using `loginRegisterInProgress` as the child dynamically updates the progress. this seemed to work until I realized that the flickering behavior is coming from the lag of iterating through the `subscriptions` object, which defaults the first render to `showAddBuilding()`

this fix forces the first render to show a success state if the `<BuildingSubscribe>` was rendered on a login via the building page. Hopefully once we get rid of the modal flow and just direct all login/register to standalone pages, we can undo this hacky fix. 

[sc-13983]


austensen (approved):
Looks great! this is such a smart fix to a weird problem, thanks for figuring this out!

austensen (approved):

PR #902, opened 2024-04-09
update address page url for detailAddr

austensen:
replace the address page URL with the selected `detailAdddr`

Previously on WOW the address page url would always reflect the search address, and when you click on other properties in the portfolio from the map the detail page updates to that address but the URL remains the same. This caused confusion, since you could think you're copying the link for building you're currently viewing but its still the search address. Also it makes linking and using browser history navigation difficult or confusing. This will make other future changes easier, like getting back to your building after going through login/signup flow, or linking back to details from portfolio tba without needing to open a new tab or reloading the page.

[sc-14358]


kiwansim (approved):
🫡

PR #904, opened 2024-04-10
Simplified login/signup flow, removes modal on building page

austensen:
This PR removes the entire on-building-page login/signup flow and modal. Now there is a single standalone login/signup page and when you're on a building page and click the "add building" button if you aren't logged in it just sends you to that page. We now make use of react-router's location state to set the building page's AddressRecord as state so we can have access to that from within the standalone login/signup flow. This allows us to still subscribe onSuccess of the login/register call, include the address subHeaders, provide a link back to the address page from the ending VerifyEmail step of those flows, or to automatically redirect users back to the building page on successful login for already verified users. We also make use of the same location state approach to set `justSubscribed` when landing back on the address page so that we can ensure it shows subscribed state immediately and we can know to show the page-level alert notifying the user that they are logged in and subscribed. Because the location state persists after reloads, we transfer the data to regular component state within a useEffect then clear the location state on both login and building page. 
[sc-14359]
[sc-14376]

Also adds a loading state to all submit buttons in login/signup flow [sc-13914]
Also ensures that emails are not broken by wrapping in login/signup flow [sc-14348]

This also fixes the bug where after login/sign to subscribe to a building, the user context was out of sync and still showed the user as not logged in on the nav etc. It seems that this was due `userContext.subscribe()` overwriting the user values (email, verified) with the default undefined state because of the same kind of bug we had previous where an empty `JustfixUser` object is truthy, so you need to use `!!user?.email` in conditions instead (see [shortcut story](https://app.shortcut.com/justfixnyc/story/13983/remove-flash-of-the-get-updates-button-before-getting-the-success-message-saying-you-re-signed-up-for-a-building#activity-14381) for details).
[sc-14242]
[sc-13983]

https://www.loom.com/share/83b0ef2c023d42af9aaea7ab3bb288ce?sid=f38f9d6c-1519-4045-a935-79716abc0797

**update:** content changes:

login/signup via nav:
image

building page when not logged in
image


austensen:
[sc-14376]

kiwansim (approved):
looks awesome. 

one thing: do you think the "back to address" button here should be an internal link style with an arrow? 
Screenshot 2024-04-11 at 1 02 36 PM
2.

kiwansim:
Oh, and the current behavior of the "Get updates..." text doesn't match the conversation on this slack [thread](https://justfixnyc.slack.com/archives/CCWHGAR9C/p1712784661142619?thread_ts=1712780851.135899&cid=CCWHGAR9C). just want to double check

austensen:
> looks awesome.
> 
> one thing: do you think the "back to address" button here should be an internal link style with an arrow? Screenshot 2024-04-11 at 1 02 36 PM 2.

yeah good question - I initially did that but it looked a little funny having it say "back" and the arrow point right, and then I saw in the mocks it was just a plain link so did that. Happy to change it back though. Here's what it looks like.
image

Alternatively, what do you think of using this back icon that Leslie had in other mocks for within the signup flow. On font awesome we have this [left arrow in circle](https://fontawesome.com/v4/icon/arrow-circle-o-left)
image

PR #910, opened 2024-05-06
Use inspection date for hpd violations timeline and email alerts

austensen:
Previously we used Notice of Violation (NOV) Issue Date as the date for WOW timeline (filled in with inspection date when not available) and so used the same for email alerts. At the time this we set HPD Online used that as their primary date as well. Now with the new HPD Online they primarily use Inspection Date (they call "Date Recorded") and so this can cause some confusion in the email alerts since the NOV Issue date comes many days later, and it won't be obvious when comparing to HPD Online. Also since Inspection Date is the first date of all the columns, this is best to update users of new activity as quickly as possible. And since we want the timeline to match email alerts, we'll make the change in both places. When the inspection date isn't available (should almost never be possible) we'll fallback to NOV issue date.


kiwansim (approved):
📅

PR #913, opened 2024-05-08
Subscriptions plural

austensen:
For some unknown reason ligui's `<Plural/>` just will not work on this page - everything appears to tbe identical to the working version on the account settings page, but it just always appears with the whole `{plural 0, building, buildings}` style messages.po text on the site itself. 

For now we've just replaced `Plural` with a simple ternary giving separate Trans elements, and maybe we can revisit later on.

Working on PR preview:

image
image


austensen:
This change didn't help...

kiwansim (approved):

PR #914, opened 2024-05-08
update type styles to work with jfcl better

austensen:
Make some small adjustments to type styles so that the new component library elements work within the unique wow style.

* reduces JFCL button and link text size
* Increases body text elsewhere

[sc-14387]


kiwansim (approved):
the type size increase looks so nice on mobile -- really helps with readability

PR #917, opened 2024-05-13
add logout buton to account settings, update styles

austensen:
Adds a Logout button to the account settings page, and makes a few minor spacing/style changes:
- Adding a page rule under the "Account" label (to help provide some structure).
- Adding "You are logged in" message with a "Log out" button.
- Removing the "Log In" label that currently lives above the user email address.
- Adding a page rule above the "Building Updates" label (to more clearly separate the contact details section from the buildings section).

image
image

image


kiwansim (approved):
this looks good to me! relatedly I just asked Joel re: what the page title should be if we are renaming "Account" to "Email settings" on the nav bar, but that doesn't have to be a part of this PR

austensen:
> this looks good to me! relatedly I just asked Joel re: what the page title should be if we are renaming "Account" to "Email settings" on the nav bar, but that doesn't have to be a part of this PR

Oh good thinking. yeah maybe we can just do that when making the change on the nav bar

PR #918, opened 2024-05-13
use router redirect for account settings page 

austensen:
In #912 I added a redirect from Account Settings to Login when you're not already logged in. But I'm now seeing that there's a bug with how I was checking logged in state with the usercontext which made it appear not logged in when it was just still fetching the user. This made it work ok if you are logged in and navigte to the account page, but once there if you reload the page while it's fetching the user it thinks you're logged out and so immediately redirects. I've now reverted that commit of that previous PR, so this is a redo.

Additionally, when I first attempted this redirect I added it within the account setting page itself, but I think the better way to do this is using React Router's Redirect within the specification of the Route. So this PR switches to use that approach

I think I've solved this issue by using `!userContext.user` to indicate that it's still fetching the user because in this state the user object is just completely undefined. Then once it's been fetched there is a user object so `userContext.user` is truthy, even if you're logged out and all the values within the object are undefined (like email, verified, etc.). So then to test if you're logged in we use `!!userContext.user.email`. So now we can show a loading page while fetching before deciding to either show the account page (if logged in) or redirect to the login page. 

[sc-14498]

### DIFF
--- a/client/src/components/UserSettingField.tsx
+++ b/client/src/components/UserSettingField.tsx
@@ -138,7 +138,6 @@ const EmailSettingFieldWithoutI18n = (props: EmailSettingFieldProps) => {
   } = useInput(oldEmail);
 
   const eventUserParams = { user_id: user.id, user_type: user.type };
-  console.log({ user, eventUserParams });
 
   const handleSubmit = async () => {
     setExistingUserError(false);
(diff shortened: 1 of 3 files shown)


kiwansim (approved):
this is honestly so illuminating

PR #922, opened 2024-05-14
add redirect login -> account when already logged in

austensen:
This adds a redirect from the Login page to Account settings if you are already logged in. Also it shows a loading page while fetching the user object until we can decide whether to redirect or not. It uses a useEffect and extra state variable to track the redirect decision to avoid the problem of redirecting at the end of your login process before you can see the verification message. 

[sc-14573]

### DIFF
--- a/client/src/containers/LoginPage.tsx
+++ b/client/src/containers/LoginPage.tsx
@@ -3,11 +3,31 @@ import { t } from "@lingui/macro";
 
 import StandalonePage from "components/StandalonePage";
 import Login from "components/Login";
+import { useContext, useEffect, useState } from "react";
+import { UserContext } from "components/UserContext";
+import { LocaleRedirect } from "i18n";
+import { createWhoOwnsWhatRoutePaths } from "routes";
+import { FixedLoadingLabel } from "components/Loader";
 
 const LoginPage = withI18n()((props: withI18nProps) => {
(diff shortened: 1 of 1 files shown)


client/src/containers/LoginPage.tsx
```diff
@@ -3,11 +3,31 @@ import { t } from "@lingui/macro";
...
+import { LocaleRedirect } from "i18n";
+import { createWhoOwnsWhatRoutePaths } from "routes";
+import { FixedLoadingLabel } from "components/Loader";
 
 const LoginPage = withI18n()((props: withI18nProps) => {
   const { i18n } = props;
+  const { account } = createWhoOwnsWhatRoutePaths();
+  const userContext = useContext(UserContext);
+  const fetchingUser = !userContext?.user;
+  const isLoggedIn = !!userContext?.user?.email;
```
kiwansim, line 17: ✨so much clearer✨

we should consider refactoring other pages that have this similar user context logic so we aren't doing `!!userContext?.user?.email` everywhere, for readability's sake

kiwansim (approved):

client/src/containers/LoginPage.tsx
austensen: oh yeah good idea, I'll make a ticket so we don't forget.

PR #934, opened 2024-05-23
Update "email settings" language on standalone page 

kiwansim:
realized I missed this when I made the nav bar change!


kiwansim:
branch was stale and had extra commit messages. not a big deal for the code but I've been running into a bunch of git problems lately so didn't want to deal with this later if it caused issues for master/prod parity

PR #937, opened 2024-05-24
add branch name to GTM event params for A/B testing

kiwansim:
edit: the CTA change will be on `cta-ab-test` branch ONLY. the event param changes are what need to be merged into `master` so that our GA4 is logging the updated event param variables from the currently live split testing `master` branch. 

---

this PR adds branch name to the `eventModel` data layer variable in GTM / GA4 [sc-14639]

ideally the branch name should've been accessible at build time with `process.env.BRANCH` (see comments in [sc-14639]) but this kept returning `undefined` even after branch and deploy preview builds. I also tried accessing via `process.env.REACT_APP_BRANCH`, appending `REACT_APP_` as convention requires for apps created with `create_react_app`.  

since the out-of-the-box env variable was not accessible, I just manually set the branch variables via netlify. see configuration below. once we do away with split testing, `eventModel.branch` will always pass in `production`. 

Screenshot 2024-05-29 at 4 54 32 PM

### DIFF
--- a/client/src/components/EmailAlertSignup.tsx
+++ b/client/src/components/EmailAlertSignup.tsx
@@ -57,7 +57,7 @@ const BuildingSubscribeWithoutI18n = (props: BuildingSubscribeProps) => {
   const [showSubscriptionLimitModal, setShowSubscriptionLimitModal] = useState(false);
 
   const navigateToLogin = () => {
-    window.gtag("event", `register-login-via-building_${process.env.BRANCH}`);
+    window.gtag("event", `register-login-via-building_${process.env.REACT_BRANCH}`);
     const loginRoute = `/${i18n.language}${account.login}`;
     history.push({ pathname: loginRoute, state: { addr } });
   };
--- a/client/src/components/EmailAlertSignup.tsx
+++ b/client/src/components/EmailAlertSignup.tsx
@@ -57,7 +57,7 @@ const BuildingSubscribeWithoutI18n = (props: BuildingSubscribeProps) => {
   const [showSubscriptionLimitModal, setShowSubscriptionLimitModal] = useState(false);
 
   const navigateToLogin = () => {
-    window.gtag("event", "register-login-via-building");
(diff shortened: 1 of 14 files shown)


client/src/components/EmailAlertSignup.tsx
```diff
@@ -20,6 +20,7 @@ import { AddressRecord } from "./APIDataTypes";
 import SendNewLink from "./SendNewLink";
 
 const SUBSCRIPTION_LIMIT = 15;
+const BRANCH_NAME = process.env.REACT_APP_BRANCH;
```
kiwansim, line 23: this feels so redundant to keep declaring const at the top but to make a const global in Typescript didn't really seem like a sound option. lmk if you have any thoughts to make this look better!

client/src/components/EmailAlertSignup.tsx
```diff
@@ -186,8 +187,15 @@ const BuildingSubscribeWithoutI18n = (props: BuildingSubscribeProps) => {
...
-              const eventParams = { ...eventUserParams, limit: SUBSCRIPTION_LIMIT };
-              window.gtag("event", "subscription-limit-request", { ...eventParams });
+              const eventParams = {
+                ...eventUserParams,
+                limit: SUBSCRIPTION_LIMIT,
+                branch: BRANCH_NAME,
+              };
+              window.gtag("event", "subscription-limit-request", {
+                ...eventParams,
+                branch: BRANCH_NAME,
```
austensen, line 197: Already added above to `eventParams` so can delete this
```suggestion

```

client/src/components/EmailAlertSignup.tsx
austensen: hmm yeah I can't think of anything better right now. Seems fine for now, especially if we're going to try for a global solution later so it doesn't need to be passed with each event.

austensen (approved):

PR #938, opened 2024-05-28
Add basic structure for signature tables and api endpoints

austensen:
Everything for the API powering the signature dashboard. 

* Adds `/signature` (like portfoliograph and ocaevictions) process for updating UNHP provided CSV and creating/updating the tables, within the protected signature schema, used for all the dashboard endpoints
  * we start with a single CSV that Jacob (UNHP) provides us and we save on in an S3 bucket. Then we'll access that file and load it into the database. This is all done in `/signature/table.py`, following the same exact process as we do for `ocaevictions`. 
  * Then the rest of the tables we need are all created just through SQL scripts added to `/sql`, and these are specified in `who-owns-what.yml` and run as part of the new signature job on our kubernetes cluster (all defined in https://github.com/JustFixNYC/nycdb-k8s-loader/pull/154)
* Adds the new endpoints 
  * we have kept the views and queries very simple, mostly just selecting a single row from one of the new signature tables, so that all the logic is contained in the other sql files above that create those tables

All of the endpoints are documented using swagger in the signature-dashboard repo: https://signature-dashboard.netlify.app/api_docs

We've made a new branch for `signature` and this will merge into there before eventually merging into main, since there is already a lot of new code here and want to start reviewing things as we go even though there will be some additional changes needed before merging into main.

Some remaining issues/changes:
* add vacate orders indicators
* remove old endpoint for building search (address-bbl pairs)
* Figure out CORS issue and go back to `CORS_ALLOW_ALL_ORIGINS = False` in settings.py (I think this was only necessary for the swagger examples to run)
* Switch WOW email alerts endpoint to use `Bearer` for auth token (compatibility with swagger api docs)
* Decide if we need any validation on `collection` api param
* add basics tests for endpoints

[sc-14501]
[sc-14502]
[sc-14503]


sql/create_signature_building_charts.sql
```diff
@@ -0,0 +1,232 @@
...
+        coalesce(hpdcomplaints_nonemergency, 0) AS hpdcomplaints_nonemergency,
+        coalesce(hpdcomplaints_total, 0) AS hpdcomplaints_total,
+
+        -- combine regular dob violations and ecb violations into one grouping called "dobviolations"
+        coalesce(regulardobviolations_total, 0) AS dobviolations_regular,
+        coalesce(ecbviolations_total, 0) AS dobviolations_ecb,
+        coalesce(regulardobviolations_total, 0) + coalesce(ecbviolations_total, 0) AS dobviolations_total,
+
+        CASE 
+            WHEN pluto_latest.unitsres > 10 THEN coalesce(evictions_filed, 0)
```
kiwansim, line 208: i thought the rent stab unit threshold for publishing eviction data was 6 units?

kiwansim (approved):

wow/dbutil.py
```diff
@@ -33,5 +33,5 @@ def exec_sql(sql: str, params: Dict[str, Any] = {}) -> List[Dict[str, Any]]:
         return dictfetchall(cursor)
 
 
-def exec_db_query(sql_file: Path, params: Dict[str, Any]) -> List[Dict[str, Any]]:
+def exec_db_query(sql_file: Path, params: Dict[str, Any] = {}) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
austensen: For the map endpoint the query doesn't take any params, but this function required you to pass something, so I thought it made sense to just use the empty dict as the default so you don't need to pass an empty one in those cases

sql/create_signature_building_charts.sql
austensen: for OCA data we can report for only 11+ unit buildings. The rent stab 6+ units is the laws for which buildings are subject to RS (6+ units and built before 1974)

PR #946, opened 2024-07-09
Signature Dashboard backend

austensen:
This adds everything for the backend of the Signature Dashboard. 

Follows the example of `ocaevictions`, adding a new `signature` folder to install on the k8s-loader for a signature job on kubernetes. This fetches a csv from our AWS S3 that contains all the data provided by UNHP and uploads that to the DB. Then there are a number of SQL scripts that generate the tables that power the api. Specifically, `signature_buildings`, `signature_building_charts`, `signature_collections`, and `signature_collection_charts`. 

Then it also adds new API endpoints to be used in the dashboard. All of the important logic for creating the indicators is in the script creating the tables, so the api queries are very simple.


wow/apiutil.py
```diff
@@ -73,12 +72,20 @@ def authorize_for_alerts(request):
...
-    token = auth[1]
+    request_token = auth[1]
 
-    if not (settings.ALERTS_API_TOKEN == token):
+    if not (token == request_token):
         raise AuthorizationError("You do not have permission to access this resource")
 
 
+def authorize_for_alerts(request):
+    authorize_with_token(request, "token", settings.ALERTS_API_TOKEN)
```
kiwansim, line 82: so we are leaving this as auth tokens as opposed to using bearer tokens?

wow/views.py
```diff
@@ -297,6 +304,116 @@ def email_alerts_multi(request):
...
+
+
+@api
+def dataset_last_updated(request):
+    """
+    This API endpoint returns data on all properties in the signature portfolio
+    for the dedicated map page.
+    """
+    args = get_validated_form_data(DatasetLastUpdatedForm, request.GET)
+    print(args)
```
kiwansim, line 407: remove debug print (if unintentionally left in)

project/settings.py
```diff
@@ -98,9 +101,11 @@ def get_required_env(key: str) -> str:
     "https://wow-django-dev.herokuapp.com",
     "https://whoownswhat.justfix.org",
     "https://demo-whoownswhat.justfix.org",
+    "https://signature-dashboard.netlify.app",
```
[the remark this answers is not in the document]
austensen: oh right, thanks!

PR #969, opened 2024-09-11
Overview table redesign & other improvements

kiwansim:
This PR includes the following changes:
- overview table overhaul [sc-15434]
- removes tool tip pop-up on hover. instead uses info icons and modals for indicator descriptions
- remove info toast that persists when filtering by landlord on Portfolio tab [sc-15462]
- Google streetview image removed from desktop view, link to Google maps is put under "Useful Links" [sc-15483]

---- 

New overview table:
Screenshot 2024-09-16 at 9 41 12 AM

Current overview table:
Screenshot 2024-09-16 at 9 41 51 AM


client/src/components/BuildingStatsTable.tsx
```diff
@@ -10,250 +10,359 @@ import JFCLLinkInternal from "./JFCLLinkInternal";
...
-        </div>
-      )}
-    </I18n>
-  );
-};
+  const BBL = () => {
+    const addr = useContext(AddrContext).getBuildingStats();
+    const { boro, block, lot } = Helpers.splitBBL(addr.bbl);
+    const description =
+      "This is the official identifer for the building according to the Dept. of Finance tax records";
```
austensen, line 41: Do these descriptions for aria-labels need to use lingui for translations as well?

austensen (approved):
This all looks good! such an improvement

client/src/components/BuildingStatsTable.tsx
austensen: oh wow that's all great to know, thanks for looking into that! That makes sense to make those smaller changes now and work on reviewing everything else later.

PR #971, opened 2024-09-25
add GCE eligibility api endpoint

austensen:
Adds a new module like portfoliograph, ocaevictions, etc. which creates two new tables entirely from SQL queries. 

- `gce_screener` - a bbl-level table with all of the fields we need for the gce screener website to inform helper text for user survey and other building data that is combined to determine eligibility and populate guides for next steps to determine portfolio size.
- `gce_eligibility` - a bbl-level table using all of the same dat as above, but designed to use only public data (no user responses) to give our best guess for GCE eligibility to use for data requests and also powers the map of likely-GCE-covered properties that the housing data coalition is working on.

We use this new module for a new jobs on nycdb-k8s-loader: https://github.com/JustFixNYC/nycdb-k8s-loader/pull/172

This also adds a new API endpoint for GET requests for a bbl that returns the entire row from the `gce_screener` table that is used on https://github.com/JustFixNYC/gce-screener


austensen:
Thanks for taking the time to review in such detail, I know some of the related property stuff is a bit convoluted.

PR #972, opened 2024-09-26
indicator dropdown shows correct timeline visualized

kiwansim:
a small bug we never noticed -- it was showing the default visualization name in the dropdown rather than the active visualization. current buggy behavior is recorded in this loom [link](https://www.loom.com/share/f6cf355a4ce441b5a8b1b24fda0db8c8) and the fixed behavior is in this loom [link](https://www.loom.com/share/bdb62e7a3c65409e9a087fb88f5172bf).

unfortunately the timespan dropdown's inaccuracy for rent stab units (showing "quarter" or "month" as selected when the graph is showing the "year" view) is not fixed by this


austensen (approved):
nice, good catch!

PR #973, opened 2024-09-26
add latest date of lagged oca filings to alerts api

austensen:
Adds a variable to the email alert endpoints with lagged eviction filings for the date of the most recently filed lagged case so that we can report that month in the notification in the emails so users know they may not appear in the data table and will need to look further back in the data on DAP portal if they need the details

Companion PR on auth-provider: https://github.com/JustFixNYC/auth-provider/pull/51

[sc-15618]


wow/sql/alerts_lagged_eviction_filings.sql
```diff
@@ -1,9 +1,10 @@
...
+  COUNT(DISTINCT INDEXNUMBERID) AS LAGGED_EVICTION_FILINGS,
+  MAX(I.FILEDDATE) AS LAGGED_EVICTION_DATE
 FROM OCA_METADATA AS M
 INNER JOIN OCA_INDEX AS I USING(INDEXNUMBERID)
 INNER JOIN OCA_ADDRESSES_WITH_BBL AS A USING(INDEXNUMBERID)
 WHERE A.BBL = %(bbl)s
   AND I.CLASSIFICATION = ANY('{Holdover,Non-Payment}')
   AND I.PROPERTYTYPE = 'Residential'
-  AND M.INITIALDATE > %(prev_date)s and I.FILEDDATE < %(prev_date)s
+  AND M.INITIALDATE > %(prev_date)s and I.FILEDDATE < %(prev_date)s and I.FILEDDATE > %(oldest_filed_date)s
```
kiwansim, line 10: FYI my original approach for capping the eviction filings to be no older than 90 days was to do an inline calculation here (see code snippet below) to avoid adding an unnecessary API param. this didn't work so i ended up adding the query param after all. if you think there's a better approach, i'm all ears! 👂🏻
```
I.FILEDDATE > %(prev_date - INTERVAL '90 days')s
```

kiwansim (approved):

wow/sql/alerts_lagged_eviction_filings.sql
austensen: Amazing thanks so much for jumping in and doing so much to fix this! I think it's fine now with the extra param, but I think your original idea would work with a slight adjustment. The `%()s` is just for the python string interpolation, so  I think it needs to just be the `prev_date` variable, and then the rest of the SQL can stay outside. 
```I.FILEDDATE > %(prev_date)s  - INTERVAL '90 days'```

kiwansim: oooo i see. this would've been much simpler 🤦🏻‍♀️
good to know for if/when we want to get rid of this param

PR #974, opened 2024-10-03
overview stats table RS units drop to zero in red text

austensen:
There was a bug in the condition where it short-circuited on 0 before testing for the drop - need to only exclude nulls first
image

[sc-15644]


kiwansim (approved):
good catch! 🤝

PR #976, opened 2024-10-21
add note for delayed hpd reg data

austensen:
The HPD registration datasets on the Open Data portal haven't been updated since June 1, but are supposed to be automatically each month. We are talking with HPD about getting things updated, but in the meantime we wanted to put a note on the WOW building pages where we usually put the latest registration date explaining that we don't have the data and to look on HPD Online where they have the up-to-date info. Hopefully we'll have the data updated soon and then we can just revert this commit. See [slack thread](https://justfixnyc.slack.com/archives/CCWHGAR9C/p1729525389206699) for discussion with Joel. 

[sc-15724]


kiwansim (approved):
looks good! one note is to run lingui extract for the new translations!

PR #977, opened 2024-10-23
overview table style changes

kiwansim:
- unbolded indicator names
- info icon is bigger and a lighter color to match signature dashboard


austensen (approved):
Looks good!

PR #980, opened 2024-10-30
WIP: consolidate email alerts into single query/api endpoint

austensen:
This PR simplifies our email alerts data api to use just one endpoint with a constant set of request params and return value structure. Previously we thought we might want the flexibility to specify a subset of indicators and so created some individual indicator endpoints and some that take a list of indicators and compose a sql query from parts. This all just makes things harder to work with and we haven't used it at all. So a single query for a single endpoint will just make everything easier. I also changed the URL `alerts/building` so that we can keep all the endpoints with these non-wow projects better organized. 

Here is the related PR to update auth-provider for the api change: https://github.com/JustFixNYC/auth-provider/pull/55

[sc-15778]
[sc-15757]
[sc-15758]


kiwansim (approved):
looks so much cleaner 🥇

PR #985, opened 2025-01-17
Rs data fix

austensen:
From Praks:
>Noting that for [this building](https://whoownswhat.justfix.org/en/address/BROOKLYN/123/MELROSE%20STREET) while the overview tab says RS units is NA, the timeline tab shows all 468 units since 2020 (assuming because 421 a starts this year) 

Since it's a new building it's not in the old RS table, and we were doing a left join so all the later years were missing. 

Separate I also noticed a typo in the RS timeline query that was leaving out 2019

[sc-16095]

### DIFF
--- a/wow/sql/address_indicatorhistory.sql
+++ b/wow/sql/address_indicatorhistory.sql
@@ -107,6 +107,8 @@ WITH TIME_SERIES AS (
       union
       select '2018-01' as month, uc2018 as RENTSTABILIZEDUNITS_TOTAL from RENTSTAB_DATA
       union
+      select '2019-01' as month, uc2019 as RENTSTABILIZEDUNITS_TOTAL from RENTSTAB_DATA
+      union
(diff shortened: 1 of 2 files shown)


wow/sql/address_indicatorhistory.sql
```diff
@@ -107,6 +107,8 @@ WITH TIME_SERIES AS (
       union
       select '2018-01' as month, uc2018 as RENTSTABILIZEDUNITS_TOTAL from RENTSTAB_DATA
       union
+      select '2019-01' as month, uc2019 as RENTSTABILIZEDUNITS_TOTAL from RENTSTAB_DATA
```
kiwansim, line 110: wild that we have been missing this

kiwansim (approved):

wow/sql/address_indicatorhistory.sql
austensen: 😶 i know!

PR #1002, opened 2025-04-08
add bbl event param to wow address search GA4 tag

austensen:
A while back we needed to get data on all the addresses searched on WOW and we needed to geocode all the addresses which was a pain. This adds BBL as an event param to the wow address search GA4 tag so we can skip that step in future. 

[sc-16168]


kiwansim (approved):
so simple yet brilliant. this will be so nice when we have to pull analytics!

PR #1004, opened 2025-04-29
Fullscreen district selection map with widget overlay on desktop

austensen:
updates to area alerts signup process - add full screen map on desktop, new mobile design, updates route to `area-alerts`


austensen:
random build fail, retry worked, here's the link: https://deploy-preview-1004--wow-django-demo.netlify.app/en/district

kiwansim (approved):

PR #1009, opened 2025-05-14
updates for k8s wow job

austensen:
some changes to make sure everything is set up for running the wow job via k8s-loader for the new area alerts work.

We depend on the newly added `pluto_latest_districts_25a` nycdb dataset, which is a new kind of nycdb dataset that has no source data but is just a sql script to create a table from other nycdb datasets (`pluto_latest` and `boundaries_25a`). We're using the pinned version so that the boundaries don't change overtime without us knowing. 

The new setup in nycdb to handle this new dataset with dependencies and no source data is not yet finalized, so we're using a commit hash for an in-progress branch that works for our needs now but still needs some improvements to handle the dependency management and empty `files` and `schema` sections of the dataset yml definition. 

https://github.com/nycdb/nycdb/pull/386

Along with this change here we are also updating the k8s-loader to add the new dataset: 
https://github.com/JustFixNYC/nycdb-k8s-loader/pull/199
And this PR that updates to handle the new pre/post sql scripts for wow:
https://github.com/JustFixNYC/nycdb-k8s-loader/pull/201

Since we now have dependencies for nycdb spatial datasets we needed to update the docker image used in our CI tests so that it has access to `shp2pgsql` because for some reason circle ci's postgis image doesn't include that. So I added a new custom docker image that just builds on the one we were previously using and installs the packages required to import the shapefiles. This is now published to our dockerhub, and we shouldn't have to make any changes to this going forward - so I just did the publish manually.

I spun up a clone of our DB and ran the updated k8s-loader wow job on it and everything worked.


goodcause/table.py
```diff
@@ -15,5 +15,6 @@ def __init__(
 def populate_tables(wow_cur, config: GoodCauseConfig):
     print(f"Creating Good Cause Eviction tables")
     for f in config.sql_files:
+        print(f"- {f.stem}")
```
kiwansim, line 18: intentionally left in ?

kiwansim (approved):
looks great. major kudos to you for adding all the test files and figuring out the order of sql executions. this is a ton of stuff!

PR #1012, opened 2025-05-20
add census tracts to area alerts district types

austensen:
Adds census tracts to options for area alerts district types. Updates the front end displays and queries to generate the indicator and geojson data. 

After checking with Corey, we're going with the labels as "Census Tract BK 001230"

Companion PR: https://github.com/JustFixNYC/auth-provider/pull/80

[sc-16365]


kiwansim:
@austensen  Yeah I think it should say Census Tract to match the dropdown selections, but I think it's a good q to raise to the larger group at a design check in. It really doesn't feel intuitive to see "BK 006000" as a selection, but "Census Tract BK 006000" also feels weird

kiwansim (approved):

PR #1033, opened 2025-06-18
WIP: incorporate bldg alert data into wow_indicators

austensen:


sql/create_indicators_table.sql
```diff
@@ -1,48 +1,117 @@
 CREATE TABLE wow_indicators AS 
 
-	WITH hpd_viol AS (
+	WITH email_dates AS (
+		SELECT 
+			email_date_send::date,
+			(email_date_send - interval '1 day')::date AS email_date_end,
```
kiwansim, line 6: oo another opportunity for refactoring the alerts_district sql and auth provider code if the dates are being built into the indicators table ! this is really great

sql/create_indicators_table.sql
```diff
@@ -1,48 +1,117 @@
...
+			) AS hpd_comp__1mo,
+
 		    count(*) FILTER (
-		    	WHERE receiveddate >= (CURRENT_DATE - interval '6' month) AND type = any('{IMMEDIATE EMERGENCY,HAZARDOUS,EMERGENCY}')
+				WHERE receiveddate between email_date_start_1_week and email_date_end
+			) AS hpd_comp__week,
+
+		    count(*) FILTER (
+				WHERE receiveddate between email_date_bldg_month_start and email_date_bldg_month_end
+			) AS hpd_comp__bldg_month,
```
kiwansim, line 88: i think it might be good to put in a comment here that 
`hpd_comp__1mo` is a 30 day range from send date and `hpd_comp__bldg_month` is 1st of the month to date and this logic applies to a couple other indicators with the same naming convention

kiwansim (approved):
this looks awesome, thanks so much for taking on the refactor for building alerts!!

sql/create_indicators_table.sql
austensen: yeah I was thinking that too, but didn't want to touch anything in this PR

sql/create_indicators_table.sql
austensen: oh right yeah great point, I'll add a comment to the top about the naming convention for all these variations

sql/create_indicators_table.sql
kiwansim: yeah heard, i think it should be an entirely new PR too

PR #1037, opened 2025-07-01
Area alerts final VQA

kiwansim:
- Update map colors w/ JF purple
- Fix CTE order in `alerts_district.sql` that was breaking an edge case 
   - Example: zip code 10464
   - With the previous order, `top_5` was pulling buildings with no activity and ranking them as 4th  and 5th. The `first_shared_portfolio` CTE would then select either 4th or 5th building as one to calculate the portfolio buildings on. All was good until this data was sent to the auth-provider, which expects the portfolio buildings and at least one of the ranked buildings to share a portfolio_id, which was not happening. This was breaking an email send.

### DIFF
--- a/client/src/components/DistrictMap.tsx
+++ b/client/src/components/DistrictMap.tsx
@@ -102,7 +102,7 @@ export const DistrictMap: React.FC<DistrictMapProps> = ({
         type: "fill",
         source: "selected",
         paint: {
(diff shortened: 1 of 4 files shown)


austensen (approved):
Looks good! Ah now I understand what the issue was that I was confused about in the other PR, thanks

PR #1045, opened 2025-08-05
Add Example Email modal on Area Alerts page

kiwansim:
Did not want to mess with the currently existing modals across the board, so just overrode the styles on Area Alerts page only. 

The example email image will be replaced before this PR is merged.


client/src/containers/DistrictAlertsPage.tsx
```diff
@@ -240,6 +245,27 @@ const DistrictCreation = withI18n()((props: withI18nProps) => {
...
+      >
+        <div className="modal__content">
+          <h3>
+            <Trans>Sample Area Alert Email</Trans>
+          </h3>
+
+          <figure>
+            <img
+              src="/area-alert-example-email.png"
+              alt="Sample image of area alert email showing building data and risk indicators"
```
austensen, line 263: Seems like our CI test/lint has a problem with including the word "image" in alt text for 
```suggestion
              alt="Sample of area alert email showing building data and risk indicators"
```

austensen (approved):

client/src/containers/DistrictAlertsPage.tsx
kiwansim: thanks for catching & fixing this!

PR #1052, opened 2025-10-02
exclude null bbl records in wow & update sample portfolios

austensen:
With the last update of hpd registrations data there were some records with BBL missing and this broke the process for updating the Algolia index for search by landlord. I've added in just a couple filters on the sql to exclude records with missing bbl, and that solves things. 

Also, we learned that the pages we link to for our two sample portfolios (Steller and A&E) aren't working because they no longer have a registratinon record so there's no portfolio. So I've also looked up some other buildings in those portfolios to update the links.


kiwansim (approved):
looks good -- glad the solution was simple and the issue was just null bbls! thanks for looking into it!

PR #1066, opened 2026-06-22
Rollbar failed to fetch issue

kiwansim:
https://app.rollbar.com/a/justfixnyc-production/fix/item/who-owns-what/2703

We've seen a flooding of `Error: failed to fetch` errors on Who Owns What for a few months now. The occurrences that trigger this Rollbar error in particular is coming from the `/auth_check` endpoint with no HTTP status code. This is most likely happening because of adblockers or users' internet connection, which are ultimately expected network failures that shouldn't be reported to Rollbar. 

We had a leftover console statement in our `friendlyFetch()` that is now deleted and I also got rid of the call to Rollbar in the low level fetch helper. 

As a side effect to this fix, I think it would actually fix this other Failed to Fetch error as well which seems more related to CORS but i think has to do with the order of error reporting calls
https://app.rollbar.com/a/justfixnyc-production/fix/item/who-owns-what/3216

### DIFF
--- a/client/src/components/AuthClient.ts
+++ b/client/src/components/AuthClient.ts
@@ -34,7 +34,13 @@ type PasswordResetResponse = Omit<VerifyEmailResponse, "statusCode"> & {
 let _user: JustfixUser | undefined;
 const user = () => _user;
 const fetchUser = async () => {
-  const authCheck = await userAuthenticated();
+  let authCheck;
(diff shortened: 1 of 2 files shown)


austensen (approved):
Thanks so much for tracking down the source of this and putting up this fix!!

PR #1067, opened 2026-07-21
Building Alerts standalone page

kiwansim:
This adds a standalone page for subscribing to Building Alerts. From the landing page, users are able to directly search an address and create a JustFix account.

This is a first pass implementation using existing Login/Password infrastructure. We plan to replace this with OTP/Magic link in the future.

[Figma mocks](https://www.figma.com/design/WlBXW4MDOQ5bMOJSRAEgSG/2026-Q3-Iteration-Convo?node-id=0-1&p=f&m=draw)


client/src/components/Icons.tsx
```diff
@@ -13,6 +13,22 @@ export const DotIcon = (props: SVGProps<SVGSVGElement>) => (
   </svg>
 );
 
+export const LocationIcon = (props: SVGProps<SVGSVGElement>) => (
```
kiwansim, line 16: Added Icon manually for now because bumping the component library version creates scope creep. The custom font awesome kit requires `react-scripts` v5 but updating `react-scripts` may have build config implications

client/src/styles/BuildingAlertsPage.scss
```diff
@@ -0,0 +1,260 @@
...
+      font-weight: 500;
+      line-height: 135%;
+      margin-bottom: 0;
+
+      .jfcl-link {
+        font-family: inherit;
+        font-size: inherit;
+        font-weight: inherit;
+        letter-spacing: inherit;
+        color: inherit;
```
austensen, line 66:
```suggestion
        color: inherit;
        cursor: pointer;
```

austensen (approved):
This all looks so good! thanks for taking this on. I agree the password login is not so bad now that I'm going through the real flow, maybe otp isn't that much better

PR #1072, opened 2026-08-21
Add HPD Complaints text details to indicators table

kiwansim:
Looks like a bunch of changes were made but it's mostly formatting. If you hit 'Hide whitespace' in the settings when you review, it will only show the real changes!

### DIFF
--- a/sql/create_indicators_table.sql
+++ b/sql/create_indicators_table.sql
@@ -138,23 +138,15 @@ hpd_comp AS (
     FROM email_dates, hpd_complaints_and_problems
     GROUP BY bbl
 ),
-hpd_comp_by_type AS (
+hpd_comp_by_type_long AS (
+  SELECT *
+  FROM (
     SELECT
-        bbl,
-        json_agg(
-            json_build_object(
-                'minorcategory', minorcategory,
-                'problemcode', problemcode,
-                'count', complaint_count
-            )
-            ORDER BY complaint_count DESC
-        ) AS hpd_comp__bldg_month_by_type
-    FROM (
-        SELECT
             hcp.bbl,
             hcp.minorcategory,
             hcp.problemcode,
-            count(*) AS complaint_count
+            count(*) AS complaint_count,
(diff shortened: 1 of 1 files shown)


sql/create_indicators_table.sql
```diff
@@ -18,468 +18,587 @@
...
+        WHERE hcp.receiveddate BETWEEN ed.email_date_bldg_month_start
+                                   AND ed.email_date_bldg_month_end
+          AND hcp.type <> 'NON EMERGENCY'
+        GROUP BY
+            hcp.bbl,
+            hcp.minorcategory,
+            hcp.problemcode
+    ) AS complaint_counts
+    GROUP BY bbl
+),
```
austensen: In case we want to limit this to only the top X complaint types, I think we can just do this here rather than keeping all of them and filtering later on. 

```suggestion
hpd_comp_by_type_long AS (
  SELECT *
  FROM (
    SELECT
            hcp.bbl,
            hcp.minorcategory,
            hcp.problemcode,
            count(*) AS complaint_count,
            ROW_NUMBER() OVER (PARTITION BY bbl ORDER BY count(*) DESC) as rn
        FROM email_dates ed
        CROSS JOIN hpd_complaints_and_problems hcp
        WHERE hcp.receiveddate BETWEEN ed.email_date_bldg_month_start
                                   AND ed.email_date_bldg_month_end
          AND hcp.type <> 'NON EMERGENCY'
        GROUP BY
            hcp.bbl,
            hcp.minorcategory,
            hcp.problemcode
  ) AS x
  WHERE rn <= 3
),
hpd_comp_by_type AS (
SELECT
        bbl,
        json_agg(
            json_build_object(
                'minorcategory', minorcategory,
                'problemcode', problemcode,
                'count', complaint_count
            )
            ORDER BY complaint_count DESC
        ) AS hpd_comp__bldg_month_by_type
    FROM hpd_comp_by_type_long
    GROUP BY bbl
),
```

austensen (approved):
Looks great! thanks for figuring this all out

PR #1079, opened 2026-09-01
update rodent inspection query on signature for new schema

austensen:
The schema changed for the DOHMH Rodent Inspections table on Open Data and the `ApprovedDate` column was removed (see https://github.com/nycdb/nycdb/pull/408). So this updates the one SQL query, for Signature Dashboard, that uses the dataset/column using only `InspectionDate` which is the only date column remaining and was already the fallback when `ApprovedDate` was unavailable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 .DS_Store
 node_modules/
 venv/
-.venv/
 env/*
 /.env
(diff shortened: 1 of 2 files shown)


kiwansim (approved):
🐭🪤 looks good to me!

PR #1080, opened 2026-09-02
update sql to add alt_addrs column on wow_bldgs and include in api

austensen:
Updates the SQL to build the `wow_bldgs` table to add a new column `alt_addrs` that is a list of up to 4 alternative addresses from a join with PAD. 

This will be used to display these Also Kwon As addresses on WOW building pages. 

It's safe to add the column and include it in the api response even though the client changes haven't been made - it will just be ignored. 

A separate PR for frontend changes will follow, but this allows us to test that against the live updated db table/api

I've run the updated wow_bldgs creation on a temp table and everything looks right.


kiwansim (approved):

sql/create_bldgs_table.sql
```diff
@@ -154,6 +155,23 @@ left join rentstab on (registrations.bbl = rentstab.ucbbl)
...
+  from (
+    select distinct trim(pad.lhnd) || ' ' || trim(pad.stname) as addr
+    from pad_adr pad
+    where pad.bbl = registrations.bbl
+      and nullif(trim(pad.lhnd), '') is not null
+      and nullif(trim(pad.stname), '') is not null
+      and upper(trim(pad.lhnd) || ' ' || trim(pad.stname))
+          <> upper(registrations.housenumber || ' ' || registrations.streetname)
+    order by 1
+    limit 4
```
kiwansim, line 172: just noting that in the case there are 5+ alt addrs (impacts 1% of the universe of addresses), we are jus taking the first 4 addrs ordered numerically

austensen: yeah do you think that's ok, or should there be a different approach?

PR #1081, opened 2026-09-02
add alt addresses to building pages

austensen:
Uses the new `alt_addrs` column added to `wow_bldgs` in #1080 to display up to 4 "also known as" addresses on building pages when available.


kiwansim (approved):
so excited to have this in finally! i feel like this takes care of one of the more confusing problems that users come across on wow

PR #1085, opened 2026-09-05
update wow client proxy to include district in send-code api call

austensen:
Companion backend PR: https://github.com/JustFixNYC/auth-provider/pull/182


kiwansim (approved):
looks good to me!!

PR #1086, opened 2026-09-05
bug fix: OTP verification in new session 

kiwansim:
if a user (new or existing) clicks a magic link on a browser or device they do not already have an established app session state, the login doesn't go through. if there's a pending building subscription, that does actually get resolved in the backend. this has to do with a mismatch between two endpoints that are called at the same time. the auth_check endpoint returns the user object based on the app session state and verify-magic-link endpoint sets cookies and returns a valid user object

### DIFF
--- a/client/src/components/UserContext.test.ts
+++ b/client/src/components/UserContext.test.ts
@@ -1,0 +1,44 @@
+import { applyAuthCheckUser } from "./UserContext";
+
+const loggedInUser = {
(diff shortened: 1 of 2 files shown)


austensen (approved):
This looks good to me. However I'm having a hard time replicating the bug - I think because when you land on the verify page you can't really see your logged in state until you navigate to a page with with nav to show login/logged-in.

PR #1088, opened 2026-09-05
fix various VQA issues from android testing

austensen:
b5c7511049f582256c12456cf7c0147b4bf374b6 - the "follow building" button alignment was off and overflowing container. this is caused by a known android bug in how it calculates `-webkit-fill-available`. 
image
518edd33727e8c4aea48ebc0ccadcc97eb4abfba - the flex `gap` on follow button conatiner also doesn't work on these nested flex containers on android, so replace with top-margin
6ed76a094101140582784fc2ce53c282087fa5bc - margin around account pages container doesn't render properly on android (same nested flex issue as above)
image
b1f7b4d3f793e8b7b0fcf9a190c118eb61e9f59d - a similr android spacing issue was happening on the newStyle modals (area+building alerts preview email and building alerts unavailable address). This fixes that and also refactors for shared styles across those uses.


kiwansim (approved):
thanks so much for the thorough VQA on android, I can't believe some of the glaring things we've missed for so long. 

I wasn't able to verify that the fixes on the account settings page look as expected on iPhone bc the preview link isn't working with the updated OTP infrastructure. Happy to check that after this is merged but all other fixes look good

PR #1091, opened 2026-09-06
properly handle geosearch 503 errors with error message

austensen:
When Geosearch is down (503) this wasn't being handled correctly on the building-alerts page and it was mistaken for a selected address not being available in wow, so the unavailable-address modal would immediately launch after typing characters into the search bar. Now they are handled separately and a new error message is shown below the bar explaining that address search is temporarily unavailable. This also applies the same thing on the home page.

to reproduce the error, use devtools, network tab, type in search, right click a request and block the domain, type again


kiwansim (approved):
really good catch and the fix is perfect

PR #1092, opened 2026-09-06
use input-otp library

austensen:
We've been encountering all kinds of minor bugs with the otp across different devices/browsers and so I think, at least for now, it's safest to just go with this popular/well maintained library that should handle everything. This also gets us proper cell-level focus outlines and the cursor, which we didn't have before and made it feel broken.


client/src/components/OtpInput.tsx
```diff
@@ -1,16 +1,19 @@
 import classNames from "classnames";
-import type { ChangeEvent, ClipboardEvent, KeyboardEvent, Ref } from "react";
+import { OTPInput, REGEXP_ONLY_DIGITS } from "input-otp/dist/index.js";
```
kiwansim, line 2: flagging that we might want to move over to this library on rent-history as well!

kiwansim (approved):
looks good to me!

PR #1096, opened 2026-09-18
Building Alerts fast follows

JustFix-org:
This PR covers a potpourri of styles and functionality updates: 
- adds Google Map view to login page if coming from a building page or standalone building alerts
- renumber registration flow steps (code entry is now step 3 of 3) 
- banner in Login flow is now scrollable w/ the page (relevant for mobile view now that map takes up a lot of the vertical space) 
- update Login flow from building alerts standalone page to use Degular 
  -  expected side effect confirmed w/ Corey: the success state after clicking magic link to login will be in Degular for all users. we think the dissonance of Degular appearing on this one page for users who are logging in from other WOW pages is better than keeping the success page using Inconsolata, which Building Alerts users will not have ever seen previously in their user journey. 
- styles pass on alternate address section on building page
- styles pass on the email address change section of Account Settings. changes mostly created by updating spacing and alignment on a reusable component on Login.tsx 

Tested on iOS mobile, Chrome, Samsung Galaxy 13


austensen (approved):
This all looks great. Thanks for taking on all these extra items

PR #1097, opened 2026-09-22
add validation for phone number

austensen:
## Validate US phone numbers at signup to match auth-provider rules

companion PR: https://github.com/JustFixNYC/auth-provider/pull/195

Adds frontend phone validation on the register/login flow so invalid US numbers (e.g. `0001234567`) are blocked before they reach auth-provider.

Phone remains optional — blank input is still valid. Validation rules mirror auth-provider `validate_phone_number`: exactly 10 digits, all numeric, first digit not `0` or `1`.

- Add `cleanPhoneDigits()` and `isValidUSPhoneNumber()` to `helpers.ts`
- Use shared validator in `PhoneNumberInput` inline error state
- Re-validate on phone step submit in `Login.tsx`
- Only attach `phoneNumber` to send-code options when digits pass validation
- Add unit tests for `isValidUSPhoneNumber`

## Test plan

- [ ] Run unit tests:
  ```bash
  CI=true REACT_APP_AMPLITUDE_API_KEY=test yarn test --testPathPattern=src/util/helpers.test.ts --watchAll=false
  ```
- [ ] Manual QA on staging/local register flow:
  - Leave phone blank → proceeds (optional)
  - Enter `0001234567` → inline error, cannot advance
  - Enter valid `5551234567` → proceeds and phone sent to backend on send-code

## Deploy notes

Can deploy independently after auth-provider (which fixes crons and backfills legacy data). This PR prevents **new** invalid phones from being submitted at signup.


client/package.json
```diff
@@ -133,5 +133,6 @@
     "extends": [
       "react-app"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22"
```
kiwansim: is this a new explicit declaration we have to do?

.gitignore
```diff
@@ -4,6 +4,7 @@ venv/
 .venv/
 env/*
 /.env
+.env.production
```
kiwansim: should we also add `/client/.env` here while we're at it? i notice that the changes there are always tracked on my local

kiwansim (approved):
the logic of phone number validation looks good! just a couple random q's about package manager and .env tracking

client/package.json
austensen: We don't technically need it, but then where ever it's being built it might use different systems and cause differences across systems. For some reasons whenever I run `yarn install` locally it automatically adds it, so I'm always discarding those changes. But from reading on it a little it seems best to have it just for that consistency.

.gitignore
austensen: yeah I think those `client/.env` are intentionally committed to act as default. And since nothing within the SPA client can actually be secret I guess the thinking was it was ok to just commit the values. Though we can change that if you think it better not to have them here. 
I added this because I was experimenting with splitting up my env files, since previously I'd just comment out local/demo/prod versions in a single file and that made it easy to forget and use the wrong one.

